PR #31, opened 2026-05-02
Add dual NVLINK Docker Compose setup for Qwen3.6-27B

JusefPol:
Added a dual NVLINK Docker Compose configuration for Qwen3.6-27B model with support for two RTX 3090 GPUs.

### DIFF
--- a/models/qwen3.6-27b/vllm/compose/docker-compose.dual-nvlink.yml
+++ b/models/qwen3.6-27b/vllm/compose/docker-compose.dual-nvlink.yml
@@ -1,0 +1,114 @@
+# ===========================================================================
+# Dual RTX 3090 with NVLINK — DEFAULT for 2× cards. Qwen3.6-27B + MTP n=3 + fp8 KV + vision.
+# NOTE: this setup is still EXPERIMENTAL and requires plenty of long term testing. initial tests show everything working fine with higher speeds
+# What this gives you (vs the single-card default):
+#   - TP=2 across both 3090s
+#   - max_num_seqs=2 → 2 concurrent agents at full 262K context (KV pool 2.36×)
+#   - max_model_len=262144 → model's natural max, no rope_scaling needed
+#   - vision + tools + MTP n=3 + recall — everything on, no compromises
+#   - Measured (re-bench 2026-04-28 on club-3090 substrate, dev205 + Genesis v7.51-stable):
+#     69.05 narr (CV 2.3%) / 88.58 code (CV 3.4%) TPS single-stream, AL 3.4, VRAM 23.6 GB/card
+#
+# What's intentionally NOT enabled:
+#   - TurboQuant KV — sidesteps vllm#40831 entirely. fp8_e5m2 is plenty for
+#     262K across two cards. For 4-stream concurrency at 262K, switch to
+#     docker-compose.dual-turbo.yml (which uses TurboQuant KV + Genesis P65).
+#
+# Dependencies:
+#   - 2× RTX 3090 (Ampere SM 8.6), PCIe-only (no NVLink — works fine)
+#   - vLLM PR #40361 (Marlin pad-sub-tile-n) — patched files volume-mounted
+#     from /opt/ai/vllm-src/. PR is open upstream; drop the mount when it
+#     lands. See ../patches/README.md for the git-clone command.
+#
+# All dual-card variants in this dir:
+#
+#   File                                Ctx    Streams   Narr/Code TPS   KV       Vision
+#   docker-compose.dual.yml (this)      262K   2         69 / 89          fp8      ✅
+#   docker-compose.dual-turbo.yml       262K   4         54 / 73          TQ3      ✅
+#   docker-compose.dual-dflash.yml      185K   1         82 / 125         FP16     ✅
+#   docker-compose.dual-dflash-noviz... 200K   1         78 / 127         FP16     ❌
+#
+# Run:
+#   cd <repo>/models/qwen3.6-27b/vllm/compose
+#   docker compose -f docker-compose.dual.yml up -d
+# ===========================================================================
+services:
+  vllm-qwen36-27b-dual:
+    # Tracking latest nightly intentionally — this stack uses fp8 KV (not
+    # TurboQuant), so it doesn't accumulate the same anchor-drift exposure
+    # the single-card project does. Pin if you hit a regression; otherwise
+    # ride the wave.
+    image: vllm/vllm-openai:nightly-7a1eb8ac2ec4ea69338c51dc7afd4b15010abfa8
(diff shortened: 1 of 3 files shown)


JusefPol:
I added the dual.yml modification only as there are no benefits on using the turbo one. I have kept the context window to max, future users can play around if they want like me 4 instead of two concurrent at lower context size.

JusefPol:
Closed by mistake. Review for merge on main

noonghunna:
Hey [@JusefPol](https://github.com/JusefPol) — thanks for taking the time to do this. Adding an NVLink-aware variant alongside the PCIe default is exactly the kind of contribution that broadens the stack to the part of the community we don't have hardware for. The diff is small, additive, doesn't touch existing dual paths — that's the right shape for this. Some items to address before we land it:

## Must-fix before merge

### 1. Container name + port collision with `dual.yml`

```yaml
container_name: vllm-qwen36-27b-dual            # same as dual.yml
ports:
  - "${PORT:-8010}:8000"                        # same as dual.yml
```

Two side-by-side dual variants need distinct identities so users can switch between them with `scripts/switch.sh` without `docker compose down` racing. Suggested:

```yaml
container_name: vllm-qwen36-27b-dual-nvlink
ports:
  - "${PORT:-8013}:8000"
```

(`8013` is currently free — `8010 dual`, `8011 dual-turbo`, `8012 dual-dflash` are taken.)

### 2. Header text says "DEFAULT for 2× cards" — needs to flip to "opt-in"

Line 1 of the new file:

```
# Dual RTX 3090 with NVLINK — DEFAULT for 2× cards. Qwen3.6-27B + MTP n=3 + fp8 KV + vision.
```

The default for 2× cards stays `dual.yml` (PCIe-only is what most consumer rigs have). Could you reword to something like:

```
# Dual RTX 3090 + NVLink bridge — opt-in variant for users with NVLink installed.
# Same model + features as dual.yml; differs only in the NCCL P2P transport.
```

### 3. Stale "PCIe-only" comment in the NVLink-enabled env section

Line 63ish:

```yaml
# PCIe-only stack — disable NCCL features that assume NVLink.
- NCCL_CUMEM_ENABLE=0
# P2P Enabled.
#- NCCL_P2P_DISABLE=1
# Specify Nvlink as the P2P level
- NCCL_P2P_LEVEL=NVL
```

The "PCIe-only stack" comment is left over from `dual.yml`. Should be flipped — something like:

```yaml
# NVLink path — keep CUMEM disabled (3090 NVLink works fine without it),
# enable P2P, and force NCCL to only use NVLink as the P2P transport.
- NCCL_CUMEM_ENABLE=0
- NCCL_P2P_LEVEL=NVL
```

(Drop the `#- NCCL_P2P_DISABLE=1` line entirely — clearer than leaving it commented.)

### 4. Stale "All dual-card variants" table in header

The cloned table still says:

```
docker-compose.dual.yml (this)      262K   2         69 / 89          fp8      ✅
```

But this file IS `dual-nvlink.yml`. Could you update to add a new row for the NVLink variant and drop "(this)" from `dual.yml`'s line? Something like:

```
docker-compose.dual.yml             262K   2         69 / 89          fp8      ✅
docker-compose.dual-nvlink.yml (this) 262K  2        TBD              fp8      ✅
docker-compose.dual-turbo.yml       262K   4         54 / 73          TQ3      ✅
docker-compose.dual-dflash.yml      185K   1         82 / 125         FP16     ✅
docker-compose.dual-dflash-noviz... 200K   1         78 / 127         FP16     ❌
```

## Nice-to-have (could land in this PR or a follow-up)

### 5. Real bench numbers from your NVLink path

The header carries `dual.yml`'s 69/89 TPS — those are from PCIe. You mentioned in your comment "initial tests show everything working fine with higher speeds" — could you share actual numbers? `bash scripts/bench.sh` runs 3 warmups + 5 measured of the canonical narr+code prompts; we'd love to publish a real NVLink row. Even an n=3 informal bench is useful.

### 6. `verify-full.sh` output

If you can paste the output of `bash scripts/verify-full.sh` against the booted NVLink container, that confirms the change doesn't regress tools / streaming / MTP acceptance. We ask this for every new compose so the community has a verified baseline.

### 7. NVLink active-confirmation step

A line in the header showing how to verify NVLink is active before running this would help future users:

```bash
nvidia-smi nvlink --status -i 0
# Expect: "Link 0: 14.062 GB/s" (or similar) — if it says "<inactive>", you don't have a working bridge
```

Otherwise users with the bridge physically present but firmware-disabled will boot fine and not realize they're not getting the NVLink path.

### 8. Mention in `docs/DUAL_CARD.md`

A row in the TL;DR table at `docs/DUAL_CARD.md` lines 9-15 would help NVLink-equipped users find this file. Easy to add — happy to do this myself in a follow-up commit if you'd prefer to keep this PR focused on the compose file.

## A point worth confirming

Your file uses `--dtype float16` (matches `dual.yml`'s current line, which also has float16). Every other dual compose on this stack also uses float16 since it matches AutoRound's calibration dtype. Just flagging in case you wanted bfloat16 — current is fine, no change needed.

---

The 4 must-fix items are small (~10 lines of edits across the file). The nice-to-have items can land in this PR or a follow-up — your call. I'll review and merge once 1-4 are addressed; if you want, post the bench + verify-full output here too and I'll fold them into the header + an entry in `models/qwen3.6-27b/CHANGELOG.md` when we land.

Thanks again for the contribution.

noonghunna:
Took this over to land it without asking for more pushes from you, @JusefPol — appreciate the contribution. Two follow-up commits on top of yours:

**Fix commit** (`6ee534c`):
- Renamed container + bumped PORT default to `8014` so it can coexist with `dual.yml` on the same rig (the original used `vllm-qwen36-27b-dual` and `8010`, both conflicts with the canonical PCIe default)
- Header reframed as community-experimental opt-in (not the default), with a note that I don't have NVLink hardware to canonical-bench
- Variant table refreshed with a new "NVLink" column so the file's role is unambiguous
- Marlin patch mounts migrated to the vendored in-repo path (`../patches/vllm-marlin-pad/...`) — `d8b341f` dropped the `/opt/ai/vllm-src/` host dep that your branch was based on

**Registration commit** (`42f8de8`):
- Wired `vllm/dual-nvlink` into `switch.sh` + `launch.sh` so users can invoke via `bash scripts/switch.sh vllm/dual-nvlink` or `bash scripts/launch.sh --variant vllm/dual-nvlink`
- Intentionally **not** added to the launch wizard's auto-shown menu — would silently boot the wrong config on PCIe-only rigs that hit the wizard

If anyone running NVLink-bridged 2× 3090s reads this and benches the new variant, please drop numbers in [discussion #19](https://github.com/noonghunna/club-3090/discussions/19) — paste-ready report via `bash scripts/report.sh`.

Merging now. Thanks again, @JusefPol.

PR #64, opened 2026-05-05
docs(upstream): track llama.cpp MTP PR #22673 + non-adoption rationale

noonghunna:
## Summary

am17an's MTP support PR for llama.cpp ([ggml-org/llama.cpp#22673](https://github.com/ggml-org/llama.cpp/pull/22673)) is unmerged but functional. Benched locally on 1× 3090 (2026-05-05) at **+34% narrative TPS** over same-config baseline (22.83 → 30.69 with \`--spec-draft-n-max 3\`, ~57% accept rate).

This PR documents the path in \`docs/UPSTREAM.md\` as 🟡 Open with explicit **"NOT a club-3090 recommendation yet"** framing so future contributors don't ship this as default or opt-in without revisiting the four blockers.

## Why we're not adopting yet

| Blocker | Impact |
|---|---|
| PR unmerged | Every cross-rig user has to compile am17an's fork, OR club-3090 maintains a custom image |
| q8_0 KV ceiling at ~64-80K | Current \`llamacpp/default\` ships **262K + q4_0 KV + vision** — trading that for +34% TPS is bad for the cliff-immune long-context audience |
| Forces \`n_parallel=1\` | Kills \`llamacpp/concurrent.yml\` (4-slot variant) outright |
| RDson MTP GGUF doesn't bundle mmproj | Vision regression unless wired separately |

Audience analysis: max single-stream TPS users already have **vllm dual-turbo at 170 TPS** (5× faster than this MTP path). llama.cpp single-card users picked it specifically for 262K + cliff-immune + vision + concurrent. MTP regresses 3 of those 4 for +34% — bad trade.

## Bench data (for the curious)

Single 3090, ctx 32K, q8_0 K+V, \`-ngl 99 -fa on -np 1\`, 3 warm + 5 measured:

| Config | Narrative TPS | Code TPS | Accept |
|---|---|---|---|
| Baseline \`--spec-type none\` | 22.83 | n/a | n/a |
| MTP \`--spec-draft-n-max 3\` ⭐ | **30.69 (+34%)** | n/a | ~57% |
| MTP \`--spec-draft-n-max 5\` | 23.55 (regression) | 31.90 | 58-62% warm |

n=5 narrative regresses because per-step verification cost outpaces marginal gain when 4th/5th draft positions accept ≪50%. Code at n=5 hit 31.9 TPS — comparable to narrative-n=3, suggesting code's higher predictability could push n=3 code higher (untested).

## Re-test triggers (adopt only after ALL)

- (a) PR #22673 merges upstream (or stock \`:server-cuda\` image starts shipping it)
- (b) q4_0 KV + MTP test recovers ≥128K context without accept-rate collapse
- (c) Cross-rig validation on ≥2 other 3090s (efschu's 109 TPS was on a 5090 — different regime)
- (d) NIAH + coherence pass at chosen quant + KV combo

Detailed reasoning + full bench narrative is documented out-of-tree in this stack's \`learnings/qwen3.6-35b-a3b.md\` (\`llama.cpp MTP — PR #22673 path\` subsection).

## Test plan

- [x] One-line addition to existing llama.cpp section table — no schema/format changes
- [x] Markdown table renders correctly
- [x] All linked URLs resolve (PR #22673, learnings reference)


ubergarm:
I'm testing it now on my 3090 TI FE tuned with LACT. it is a *draft* PR and does seem to benefit decode (token generation) especially for structured type text

```
prompt eval time =     529.78 ms /   526 tokens (    1.01 ms per token,   992.86 tokens per second)
       eval time =    3255.89 ms /   280 tokens (   11.63 ms per token,    86.00 tokens per second)
      total time =    3785.68 ms /   806 tokens
draft acceptance rate = 0.90667 (  204 accepted /   225 generated)
statistics mtp: #calls(b,g,a) = 32 2853 2363, #gen drafts = 2853, #acc drafts = 2363, #gen tokens = 8559, #acc tokens = 5619, dur(b,g,a) = 0.027, 14940.889, 5.351 ms
```

i'm using the same quant you mention https://huggingface.co/RDson/Qwen3.6-27B-MTP-Q4_K_M-GGUF

i also released an ik_llama.cpp compatible quant with some tips here for MTP: https://huggingface.co/ubergarm/Qwen3.6-27B-GGUF/discussions/2#69fa0f7d8ab0c1b3e49d8e58 (*EDIT* link corrected)

you can grab any mmproj, doesn't matter about the quant provider so much for that. but at least on ik mmproj and spec decoding are not working together afaict.

the real story to me is checking out regression for prefill (prompt processing) and addressing that, though when using `aiperf` testing vllm against llama's the caching on llama seemed better than vllm so gave better effective performance for some workloads.

anyway, lots of options for 3090 folks for sure!

here is the command i'm testing it with now just for benching while doing other stuff.

```bash
model=/mnt/ai/models/RDson/Qwen3.6-27B-MTP-Q4_K_M.gguf

  #--dry-run \
CUDA_VISIBLE_DEVICES="0" \
./build/bin/llama-server \
  --model "$model" \
  --alias Qwen3.6-27B \
  -c 32768 \
  -ctk f16 -ctv f16 \
  --spec-type mtp --spec-draft-n-max 3 \
  -ngl 99 \
  --threads 1 \
  --host 127.0.0.1 \
  --port 8080 \
  --parallel 1 \
  --jinja \
  --no-mmap \
  --ctx-checkpoints 32 \
  -cram 32768

$ git branch | grep '*'
* mtp-clean
$ ./build/bin/llama-server --version
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 24080 MiB):
  Device 0: NVIDIA GeForce RTX 3090 Ti, compute capability 8.6, VMM: yes, VRAM: 24080 MiB
version: 305 (17df5830e)
built with GNU 15.2.1 for Linux x86_64
```

noonghunna:
@ubergarm — re-benched on our 1× 3090 with your `-ctk f16 -ctv f16` plus q4_0 for completeness. **Surprising result: KV format isn't the explainer.**

## Full sweep, our 1× 3090, ctx 32K, n=3, RDson MTP GGUF, 3 warm + 5 measured each

| KV | Narr TPS (essay) | Narr accept | Code TPS (LRU) | Code accept |
|---|---|---|---|---|
| q8_0 (original bench) | 30.69 | ~57% | 31.90 (n=5) | 58-62% |
| **fp16** (your config) | **30.50** | ~55-57% | **37.71** | 77-81% |
| q4_0 (matches `llamacpp/default`) | 30.04 | ~54% | 36.64 | 72-80% |

**Narrative TPS is essentially flat across all three KV formats** (30.04 ↔ 30.50 ↔ 30.69, well inside run-to-run CV). Accept rate also flat at 54-57% on the canonical narrative essay regardless of KV. So switching `-ctk q8_0` → `-ctk f16` on our hardware didn't move the needle for narrative.

The code prompt picks up ~+20% at fp16 / q4 over q8 from accept-rate climbing 60% → 78% (prompt-distribution effect dominates), but **even at fp16 + structured prompt we're at 37.71 TPS, not 86**. The 2.3× residual gap can't be KV format on our rig.

## Where the gap likely lives

Suspects in order of likely effect size:

1. **3090 Ti FE vs 3090** — your card has higher boost (+5-8%) and ~8-10% more memory bandwidth (1008 vs 936 GB/s). Decode is bandwidth-bound, so call it ~10-13%.
2. **LACT tuning** — voltage/clock optimization, plausibly +5-10% on memory-bound decode.
3. **Build commit** — we built `038d787` (top of `mtp-clean` as of last week). You're on `17df5830e` (build 305). Same branch but commits diverge; a kernel-path change between them isn't impossible.
4. **Your prompt distribution** — your 90.7% accept implies the request was even more structured-friendly than my LRU-cache prompt (which got me 78%). On really high-accept distributions, the per-step gain compounds: at n=3 with 90% accept, expected accepted-per-step ≈ 1 + 3·0.9 = 3.7 tokens vs ~3.2 at 78% — a +16% TPS multiplier on top of the same kernel.
5. **Memory-tuning flags** — `-cram 32768 --ctx-checkpoints 32 --no-mmap`. We didn't engage these; not sure if they're material to TPS or QoL features. Asking below.

Plausibly stacked: 1+2 ≈ +20%, 3 unknown, 4 ≈ +15%, leaves ~30% unexplained — which would land us at the upper bound of measurement variance + LACT + tuning. Doesn't fully reconcile to 86, but doesn't require anything mysterious either. **A 3090-Ti owner running our exact `llamacpp/default` (same llama.cpp commit + q4_0 KV + Q3_K_XL) for an apples-to-apples baseline would close the loop.**

## Questions for you

1. **Are `-cram 32768 --ctx-checkpoints 32 --no-mmap` material to your TPS, or are those there for KV-checkpoint resume / mmap-pressure reasons unrelated to throughput?** Asking before we copy them in blind.
2. **Your "prefill regression" observation** — that's the bigger story IMO. Got numbers? My hand-wave is "MTP head fires per token regardless of accept, so prefill takes a hit," but I haven't characterized it. Comparison vs vanilla (`--spec-type none`) on the same prompt set would isolate it cleanly.
3. **The ik_llama.cpp link in your comment is the same RDson URL as the first** — typo? If you have a separate ik-quant repo with MTP tips, I'd grab it.

## Doesn't flip the "not adopting yet" call

Even at 86 TPS the four club-3090-side blockers stay (PR unmerged, ctx ceiling at fp16/q8 + 32K, `--parallel 1` only, no mmproj in RDson GGUF). Daylight to vLLM dual-turbo (170 TPS aggregate / 81 single-stream) narrows but doesn't close. Watching #22673 for upstream merge.

That said — your data significantly raises the ceiling on what the path is capable of in expert hands, which is useful framing for anyone who lands on the UPSTREAM tracker entry and wonders "is 30 TPS the best you'll get?" Adding a footnote linking your data point.

Thanks for engaging on this with structured numbers — exactly the kind of cross-rig back-and-forth that makes the matrix more honest.

ubergarm:
1. it is important to disable `--ctx-checkpoints 0 -cram 0` caching if you want to benchmark the underlying implementation and not the cached responses. I just did some more benchmarks here with both client `aiperf` and llama-server settings used https://github.com/ikawrakow/ik_llama.cpp/pull/1741#issuecomment-4382174687 . i don't expect --no-mmap to matter for full GPU offload.

2. haven't run with/without, but on ik it was almost 20% and not sure on mainline. the linked pr above though ik improves the prefill regression with mtp. keep in mind i'm not sure the best set of arguments between ik and mainline e.g. on ik it is `-mtp --draft-max 4 --draft-p-min 0.75` works, but mainline doesn't have exact same arguments as shown in the linked report.

3. Oops, sorry yeah my copy paste buffer is wonky since updating, here is the ik_llama.cpp discussion and quant reference: https://huggingface.co/ubergarm/Qwen3.6-27B-GGUF/discussions/2#69fa0f7d8ab0c1b3e49d8e58 

> Daylight to vLLM dual-turbo (170 TPS aggregate / 81 single-stream) narrows but doesn't close.

I assume you mean running vllm with this: https://github.com/noonghunna/club-3090/blob/master/models/qwen3.6-27b/vllm/compose/docker-compose.minimal.yml#L55-L86 

*EDIT* oh this one? https://github.com/noonghunna/club-3090/blob/master/models/qwen3.6-27b/vllm/compose/docker-compose.dual-turbo.yml  . that is for 2x GPUs, I'm only testing one here. So probably not the one to compare against.

1. `qwen3.6-27b-autoround-int4` is likely lower KLD/PPL than mainline and especially ik GGUFs of similar size
2. kv-cache `fp8_e5m2` while it is supported native on sm86 arch, not as good as f16 or q8_0 or possibly even `-khad -ctk q6_0` given both mainline and ik support hadamard transforms aka "rotations" similar to turboquant stuff
3. huh i don't see flashinfer mentioned and no spec config. i'm not sure where the vllm command is at for this "dual-turbo" business

noonghunna:
@ubergarm — three good corrections, all taken:

**1. `--ctx-checkpoints 0 -cram 0` for true bench.** Hadn't traced what those flags actually did — assumed they were prompt-cache QoL features, not request-level caching. Will add to our bench notes that anyone wanting clean numbers should disable them on llama.cpp servers. Pulling your [ik bench data](https://github.com/ikawrakow/ik_llama.cpp/pull/1741#issuecomment-4382174687) into our learnings file as the proper Ampere/3090-Ti baseline for ik_llama.cpp.

**2. Apples-to-oranges comparison — fair catch.** The "170 TPS dual-turbo / 81 single-stream" framing was sloppy: dual-turbo is 2× 3090 TP=2 + custom local sidecars (Genesis patches + vendored Marlin pad fork), not a single-card config. Your 86 TPS is single-card 3090 Ti. The honest comparison for your rig would be **our single-card vLLM `minimal.yml` at ~32 TPS narrative** (no Genesis, no spec-decode, fp8 KV) — your 86 wall TPS via mainline+MTP+yourquant on a single card **outperforms our single-card vLLM by ~2.7×**. That's a stronger statement and the right one. Will edit our [discussion #67](https://github.com/noonghunna/club-3090/discussions/67) framing to cite single-card-vs-single-card explicitly.

**3. ik_llama.cpp link** — got the right one now, [ubergarm/Qwen3.6-27B-GGUF discussion #2](https://huggingface.co/ubergarm/Qwen3.6-27B-GGUF/discussions/2#69fa0f7d8ab0c1b3e49d8e58). Saving it.

Two further things I'd want to dig into when there's a window:

> kv-cache `fp8_e5m2` ... not as good as f16 or q8_0 or possibly even `-khad -ctk q6_0` given both mainline and ik support hadamard transforms aka "rotations" similar to turboquant stuff

This is a pointer I hadn't explored. We've measured fp8_e5m2 ↔ TQ3 ↔ q8_0 ↔ FP16 across Qwen3.6 (table in [`learnings/qwen3.6-35b-a3b.md`](https://github.com/noonghunna/3090) — out-of-tree), but `-khad` hadamard rotations + q6_0 is a path we never tried. If hadamard pre-rotation is approximately quality-equivalent to TurboQuant's per-tensor transform on a smaller bit budget, that would be a meaningful KV format option for the constrained-VRAM audience. Filing as a re-test trigger.

> dflash code TPS of 142 is impressive but the FP16 KV pressure makes it less suitable for long accumulated contexts

This part is mostly the engineering decision behind why we have multiple `dual-*` composes — DFlash is the "short-prompt code champion" for users whose workload is fresh-prompt iteration, while `dual-turbo` (TQ3) is the long-accumulated-context option. Different audiences.

> the linked PR above ik improves the prefill regression with mtp

Good signal that prefill regression is actively addressable on the ik_llama.cpp side. We'll watch it for upstream propagation; mainline llama.cpp prefill (PR #22673) is currently slower than no-spec-decode on our hand-runs but I haven't characterized it carefully. Worth a focused prefill bench with `--ctx-checkpoints 0 -cram 0 --no-mmap` per your notes.

Genuinely useful exchange — the kind of cross-runtime back-and-forth that improves both stacks. Will close the loop by updating `docs/UPSTREAM.md` to track #22673 with your data linked.

ubergarm:
Thanks! Yeah I'm fiddling around trying to get vLLM to run locally now too. My most interesting finding is that for a workload that might benefit from heavy caching of prompts, both llama's really shine there.

* cache run
  * `--enable-prefix-caching` in vLLM is *much slower* than either llama's `--ctx-checkpoints 32 -cram 32768`

* no cache
  * disabled vLLM's with `--no-enable-prefix-caching` and compared to llama's `--ctx-checkpoints 0 -cram 0` showing prefill is about the same across the board once you remove caching

```
┌────────────────────────────┬────────────────────┬─────────────────┬─────────┬──────────────────┬
│ Implementation             │ Prefill (no cache) │ Prefill (cache) │ Speedup │ Decode (no       │
│                            │                    │                 │         │ cache)           │
├────────────────────────────┼────────────────────┼─────────────────┼─────────┼──────────────────┼
│ mainline llama.cpp         │ 1315 tok/s         │ 19,762 tok/s    │ 15.0x   │ 69.8 tok/s       │
├────────────────────────────┼────────────────────┼─────────────────┼─────────┼──────────────────┼
│ ik llama.cpp (mtp_tweaks1) │ 1341 tok/s         │ 19,908 tok/s    │ 14.8x   │ 56.1 tok/s       │
├────────────────────────────┼────────────────────┼─────────────────┼─────────┼──────────────────┼
│ vLLM (int4-AutoRound)      │ 1368 tok/s         │ 4,750 tok/s     │ 3.5x    │ 88.4 tok/s       │
└────────────────────────────┴────────────────────┴─────────────────┴─────────┴──────────────────┴
```

The quants are different for sure, and my own IQ4_KS quant uses `q8_0` for all MTP tensors/layer unlike RDson's.

- Qwen3.6-27B-MTP-Q4_K_M (GGUF, used by both llama.cpp runs): RDson/Qwen3.6-27B-MTP-Q4_K_M-GGUF (https://huggingface.co/RDson/Qwen3.6-27B-MTP-Q4_K_M-GGUF) — MTP tensors quantized to Q4_K
- Qwen3.6-27B-int4-AutoRound (vLLM): Lorbus/Qwen3.6-27B-int4-AutoRound (https://huggingface.co/Lorbus/Qwen3.6-27B-int4-AutoRound) — MTP tensors kept at full bf16

Anyway, have fun and I'll keep poking at it. I'm not brave enough *yet* to try the wild looking "Genesis" patches and turboquant stuff given ik_llama.cpp with `-khad -ctk q6_0 -vhad -ctv q4_0` is as low as I'd like to go to squeeze out more context.

PR #84, opened 2026-05-06
composes: PYTORCH_CUDA_ALLOC_CONF env-override knob (WSL2 boot-crash workaround)

easel:
## Summary

A single-card RTX 3090 Ti rig on WSL2 hits `RuntimeError: CUDA driver error: device not ready` from `gptq_marlin_repack` immediately after weight load on the v7.72.2-uplift nightly pin (`vllm-openai:nightly-01d4d1ad`, driver 596.36). Bisect against the minimal compose (no Genesis, no spec-decode, no TQ3 KV) reproduced cleanly with just `--quantization auto_round`; `CUDA_LAUNCH_BLOCKING=1` did not move the failure site. TDR was ruled out (registry already extended + Windows rebooted).

`PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False` resolves the crash. This is the same workaround that already addresses JusefPol's NVLink boot-crash (PR #31), hardcoded in the `dual-nvlink*.yml` composes since then.

The exact failing call hasn't been isolated — `cuMemMap` (the virtual-memory API used internally by `expandable_segments:True`) is the suspected culprit since both known occurrences respond to the same env-var fix, but we haven't proven which specific call returns `cudaErrorNotReady`.

## Changes

- **14 composes** (single-card + PCIe dual-card): hardcoded `PYTORCH_CUDA_ALLOC_CONF=...` → `${PYTORCH_CUDA_ALLOC_CONF:-...}` env-override knob with current values preserved as defaults. Pattern mirrors the `MAX_MODEL_LEN` / `GPU_MEMORY_UTILIZATION` knobs from #79.
- **`dual-nvlink.yml` / `dual-nvlink-turbo.yml`**: untouched. Their existing JusefPol-driven default already has `expandable_segments` off — adding an override would be a footgun.
- **`.env.example`**: new entry under "vLLM tuning knobs" documenting when to flip it.
- **`docs/HARDWARE.md`**: new `### Fix — disable PyTorch expandable_segments` subsection alongside the TDR fix. Includes stack trace, what was ruled out, the two known occurrences, override recipe, and a single uncontrolled observation about weight-load time on the diagnostic rig (32s → 13s, flagged as suggestive not measured).
- **`docs/FAQ.md`**: WSL2 question now cross-links both fix subsections.
- **`CHANGELOG.md`** (top-level) and **`models/qwen3.6-27b/CHANGELOG.md`**: dated 2026-05-06 entries.

## Override usage

\`\`\`sh
# models/qwen3.6-27b/vllm/compose/.env
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False
\`\`\`

Then \`docker compose up -d\` as usual. No edits to tracked files needed.

## Test plan

- [x] Verified \`gptq_marlin_repack\` boot crash is reproducible on the diagnostic rig with \`expandable_segments:True\` (default + minimal composes, both fail identically)
- [x] Verified \`expandable_segments:False\` boots cleanly on the **default** compose (full Genesis + MTP K=3 + TQ3 KV + cudagraph capture, end-to-end through \`Application startup complete\`)
- [x] Verified \`\${PYTORCH_CUDA_ALLOC_CONF:-...}\` shell substitution behaves correctly inside the compose \`environment:\` block
- [ ] Cross-rig validation on a non-WSL2 environment that previously worked (regression check that the env-var pattern doesn't break bare-metal)
- [ ] Confirm \`dual-nvlink*.yml\` users aren't impacted (no edits there, but worth a sanity boot if anyone has the hardware)

## Notes

- This was diagnosed and is verified on **single-card WSL2**. The fix is well-scoped (env-var name change, no behavior change at default) so blast radius is low, but the cross-rig validation tickbox above remains open.
- Regarding the weight-load timing observation: the 32s → 13s delta is on the same rig with caches cleared, but cache warmth, FS state, and other factors weren't held constant. The HARDWARE.md text flags it as suggestive only; not a benchmark claim.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,6 +64,15 @@
 # tuning; see docs/USE_CASES.md for the curve.
 # MAX_MODEL_LEN=48000
 
+# PyTorch CUDA allocator config. Default keeps `expandable_segments:True`.
+# On some setups the engine crashes at boot inside `gptq_marlin_repack` with
+# `RuntimeError: CUDA driver error: device not ready`; setting this to
+# `expandable_segments:False` resolves it. Known occurrences: JusefPol on
+# NVLink (PR #31), this repo on single-card 3090 Ti / WSL2 / driver 596.36.
+# See docs/HARDWARE.md "disable PyTorch expandable_segments" for the full
+# failure signature.
+# PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False
+
 
 # -----------------------------------------------------------------------------
 # Genesis patch tree (vLLM only)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 Changes that span the entire stack — engine version pins, script behavior, repo structure. Per-model dated history lives in `models/<name>/CHANGELOG.md`.
 
+## 2026-05-06 — `PYTORCH_CUDA_ALLOC_CONF` exposed as override knob (boot-crash workaround)
+
+A single-card RTX 3090 Ti rig on WSL2 (driver 596.36) hit `RuntimeError: CUDA driver error: device not ready` from `gptq_marlin_repack` immediately after weight load on the v7.72.2-uplift nightly pin. Bisect (default → minimal-no-Genesis → minimal + `CUDA_LAUNCH_BLOCKING=1`) ruled out Genesis, spec-decode, TQ3 KV, async-residual error, and TDR (registry already extended + Windows rebooted). Setting `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False` resolves the crash. Same env-var workaround as JusefPol's NVLink boot-crash report (PR #31), already hardcoded in the `dual-nvlink*.yml` composes.
+
+**Changes:**
+- All 14 single-card and PCIe dual-card composes: `PYTORCH_CUDA_ALLOC_CONF=...` → `${PYTORCH_CUDA_ALLOC_CONF:-...}` env-override knob (defaults preserved). Pattern matches existing `MAX_MODEL_LEN` / `GPU_MEMORY_UTILIZATION` overrides from #79.
+- `dual-nvlink.yml` / `dual-nvlink-turbo.yml`: untouched (their existing JusefPol-driven default already has `expandable_segments` off).
+- `.env.example`: documented the override under "vLLM tuning knobs".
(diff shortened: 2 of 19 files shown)


noonghunna:
@easel — sincere thanks for this. Diagnosed the WSL2 + Marlin repack issue end-to-end, identified the right workaround (matches the JusefPol fix surface from #31), and shaped the patch in the established env-override knob pattern from #79 (\`MAX_MODEL_LEN\` / \`GPU_MEMORY_UTILIZATION\`). Clean PR — minimal blast radius, defaults preserved, dual-nvlink composes correctly left untouched.

Merged to master as [\`276ab89\`](https://github.com/noonghunna/club-3090/commit/276ab89).

## Cross-rig validation just landed

[@timxx](https://github.com/timxx) was the second WSL2 contributor to hit this in [issue #60](https://github.com/noonghunna/club-3090/issues/60) — single 3090 + driver 591.86 + CUDA 13.1 + WSL2 kernel 6.6.87.2-microsoft-standard. He'd already bisected to the same image SHA range and confirmed \`minimal.yml\` ALSO fails (which is the smoking gun that the bug is in the upstream Marlin path, not anywhere we control).

Pointed him at the merged fix. When his confirmation comes back, that closes your test-plan checkbox 4 (cross-rig regression check on a non-bare-metal-bare-metal WSL2 environment that previously worked) and gives us **two confirmed WSL2 boot recoveries**: yours (driver 596.36) + his (driver 591.86). Different driver versions + same fix working = generic enough to ship.

## What this unblocks

WSL2 is a non-trivial slice of cross-rig contributors — Windows-host devs who want to test single-card 3090 builds without dual-booting. Hardcoded \`expandable_segments:True\` in the composes was a silent footgun for them; your env-override pattern means they can fix it locally via \`.env\` without touching tracked files, and bare-metal contributors who don't need it see no behavior change. Net win for the matrix.

## Side note on documentation

Your additions to \`docs/HARDWARE.md\` (the new "Fix — disable PyTorch expandable_segments" subsection alongside the TDR fix) and the WSL2 cross-link in \`docs/FAQ.md\` are exactly the right shape — diagnostic-first, with rule-out hypotheses listed. Future WSL2 contributors will land there searching for "Marlin OOM" or "device not ready" and find both occurrences documented.

Thanks again — this is high-signal contributor work.

PR #114, opened 2026-05-10
bench-agentic.sh: multi-depth TTFT/decode-TPS-vs-context producer (depth-curve for timeout-sizing + optimizer)

easel:
**Retargeted 2026-05-28** (was "multi-turn prefill stress benchmark"). After review + an on-rig depth analysis, this PR is best understood as the **automated producer of the throughput-vs-context-depth curve** the stack currently lacks — not just a stress bench.

## Why this is the right tool (and why extending soak isn't)
`soak-test.sh --continuous` *captures* per-turn TTFT/decode/VRAM but: (a) its `turn-log.csv` has **no context-depth axis** (`completion_tokens` only — no `prompt_tokens`), and (b) its ramp tops out at **~22–25K** accumulated tokens — below the 35K+ cliffs where the curve matters. So "just summarize soak's logs" can't produce the curve without a producer change *and* a deeper ramp = effectively rebuilding this. `bench-agentic.sh` is purpose-built: it ramps deep (~53K, toward the cliffs) **and** logs per-turn depth. (Verified on-rig 2026-05-28.)

## What it produces
Per-turn `(prompt_tokens, ttft_ms, decode_tps)` as accumulated context grows — i.e. a **decode-TPS-vs-context-depth curve**. The only such data today is the hand-measured table in `docs/CLIFFS.md`; this automates it.

## Consumers
1. The runtime measurement-record `decode_tps_by_ctx` ladder (`scripts/lib/profiles/measurement_record.py`) — currently one rung (`canonical-short`); each depth point becomes a rung.
2. The planned **opt-in soak-derived timeout tier** in quality-test (sizes per-scenario timeouts from real per-depth speed instead of the single empty-context startup probe). See `docs/QUALITY_TEST.md` → "Per-scenario timeouts".
3. (Later) the v0.9.0 compose-optimizer's deferred `throughput`/`concurrency` objectives, which explicitly need a TPS-vs-context measured prior.

## Staged plan
**Now (revive):** the script is finished + robust — bring it current:
- Re-wire `report.sh --agentic` **by hand** against current master (the `@@ -704` hunk won't apply; `DO_BENCH` is ~line 866).
- Generalize the config-pins: drop the hardcoded `:8020` (use preflight autodetect), make the ~35K/~74K Cliff thresholds informational, and **relabel "prefix-cache effectiveness" → per-`arch_class` O(n) DeltaNet/SSM TTFT-growth** (honest for qwen3-next).
- Drop this body's old "preflight.sh fix" claim — it's not in the diff (`preflight_autodetect_endpoint` already exists, so the script runs without it).
- Document the fixture as an opaque context payload (its pre-v0.8.6 compose paths are benign).

**Deferred (to Lock-#6 schema-freeze):** the JSON-emit into the measurement-record ladder + the timeout consumer (opt-in flag + lookup). The producer's `decode_tps_by_ctx` typing is a Lock-#6 *candidate*, not frozen — building the emit now risks rework.

## Caveats
- A single agentic pass is **n=1 per depth** → a curve *shape* reproducer, **not** a calibration-grade ladder rung (needs multi-session aggregation per the 3-warm/3-measured protocol before a consumer trusts it).
- The dramatic curve is **vLLM-DeltaNet-specific** (TTFT ~25× higher than llama.cpp at 35K) — one cell of a per-`(engine, arch_class, config)` matrix, not a universal producer.

## Validation before merge
Live single-card vLLM Qwen3.6-27B AutoRound run; confirm the TTFT-growth reproduces the `CLIFFS.md` O(n) shape; capture VRAM per the bench protocol; `report.sh --full` renders the new section. Branch-first (substantive tooling).


noonghunna (approved):
LGTM — really nice work. The script is the right shape (preflight autodetect, line-buffered output, statistical reporting per-turn, prefix-cache effectiveness analysis with thresholds), and the headline finding (45.5× TTFT growth on vLLM single-card from 1.2K → 35.6K accumulated tokens, vs llama.cpp 25× faster at 35K) is exactly the cross-engine characterization we've been circling around the Cliff 2 work without measuring directly. The script becomes the canonical tool for that.

**Fixture vetted** — `<repo>` placeholder used consistently across all 15 turns. The two literal-path hits in turn 14 (`s|/home/${USER_NAME}|~|g` + `/mnt/models/huggingface`) are content from our own `scripts/report.sh` that your tool result captured — that's our public code, not your rig.

**Bonus:** the `preflight.sh` `127.0.0.1:PORT->` regex fix is a real bug — `BIND_HOST=127.0.0.1` setups are silently broken on master today.

**One minor question (non-blocking — could land as a follow-up):** the bench forces `tool_choice: "required"` and hard-errors with `RuntimeError("server returned no tool calls...")` if a parser doesn't honor it. The vLLM gemma4 and qwen3-coder parsers both support `required`, and stock llama.cpp `--jinja` does too — but it's worth a one-line note in the script doc saying "your tool-call-parser must support `tool_choice=required`". That way someone running this on a non-conforming parser gets the diagnostic from the doc, not from a stack trace.

Approving — will merge after a quick CI dry-run locally.

easel:
> **One minor question (non-blocking — could land as a follow-up):** the bench forces `tool_choice: "required"` and hard-errors with `RuntimeError("server returned no tool calls...")` if a parser doesn't honor it. The vLLM gemma4 and qwen3-coder parsers both support `required`, and stock llama.cpp `--jinja` does too — but it's worth a one-line note in the script doc saying "your tool-call-parser must support `tool_choice=required`". That way someone running this on a non-conforming parser gets the diagnostic from the doc, not from a stack trace.

The note is a good idea. 

As a side note, I do think the `tool_choice=required` is correct here. It surfaced a tool calling response error on turn 11 due to low quant in my testing that would have gone unseen without a full e2e benchmark that would have likely obscured the issue. In general we are all circling around the problem of “is it good enough for agentic workflows” and between these performance cliffs and the the silent breakage of tool calls just getting dropped on the ground thats the difference between “works for me” and “doesn’t work”.

noonghunna:
btw i switched thinking off by default on qwen compose files today. benchlocal-cli is out, thinking was a pain to work with while testing the bench. Feel free to give it a go.

easel:
> btw i switched thinking off by default on qwen compose files today. benchlocal-cli is out, thinking was a pain to work with while testing the bench. Feel free to give it a go.

Thinking is an interesting one. It seems like it's a key part of how the frontier models differentiate themselves, yet it's highly risky with these local models as it often spirals forever. I've also found that even the ability to control thinking reliably can be challenging.

easel:
As an aside, I'm progressing towards a set of comparable https://www.tbench.ai/ runs across frontier models, qwen3.6 via open router, omlx and club-3090. Right now I'm seeing a ~55% pass rate on open router, but much lower for the local models, typically due to timeouts.  It appears as if a 2.7x multiplier is about right for "vllm on 3090" vs. "open router".

Details will be landing at https://easel.github.io/fizeau/benchmarks/terminal-bench-2-1/providers/ -- it's a little ai sloppy atm but will tighten up as I get the underlying numbers sorted out.

I'm seeing that as the real test of viability -- the 56% measurement is already pretty eye-opening in that for all the noise about "local models are taking over the world" it's a pretty huge gap from the ~82% gpt-5.5 codex gets. I'm still working on getting good comparable for my harness vs. codex but right now it seems to be within a couple of percent.

noonghunna:
just posted some results here https://github.com/noonghunna/club-3090/discussions/117#discussioncomment-16872705

PR #115, opened 2026-05-10
llama.cpp: add MTP speculative-decode Docker build + compose

easel:
## Summary

- **`Dockerfile.mtp`** (`models/qwen3.6-27b/llama-cpp/build/`): new Docker build from `am17an/llama.cpp:mtp-clean` (PR [#22673](https://github.com/ggml-org/llama.cpp/pull/22673)). Three linker fixes required to produce a working CUDA binary on Ampere/Blackwell: CUDA VMM stub symlink + `-DGGML_CUDA_NO_VMM=ON`, staged `.so` copy (multi-stage build doesn't forward shared libs by default), and `libgomp1` runtime dep. Targets sm_89 (Ada/4090) and sm_120 (Blackwell/5090).
- **`compose/mtp/docker-compose.yml`** (`models/qwen3.6-27b/llama-cpp/compose/mtp/`): new compose for the `llama-cpp-mtp:latest` image using the havenoammo UD-Q3_K_XL GGUF (MTP heads are **embedded in the main GGUF** — no `--model-draft` needed, unlike external-draft configs). Env knobs: `CACHE_TYPE_K`/`CACHE_TYPE_V`, `SPEC_N_MAX` (default 3), `DISABLE_THINKING`, `CUDA_VISIBLE_DEVICES`. Follows master's `compose/single/` subdirectory convention.
- **`compose/single/docker-compose.yml`**: document `CACHE_TYPE_K`/`CACHE_TYPE_V` split (replaces single `KV_TYPE`; legacy fallback preserved), add `--spec-type ${SPEC_TYPE:-none}` flag, update profile comment to reference the new MTP compose.
- **`scripts/preflight.sh`**: add `llama-cpp-mtp-qwen36-27b` to the container-name pattern in `preflight_autodetect_endpoint`; add `|| true` to prevent `pipefail` exit when no container is running.
- **`scripts/bench.sh`**: enable `line_buffering=True` on stdout for live output during bench runs.
- **`BENCHMARKS.md`**: first Docker-based llama.cpp MTP data point — 5090 Laptop (24 GB, WSL2, Performance mode ~166W), 49.54/65.75 TPS narr/code, 262K ctx, q4_0 KV, TTFT 232/228 ms, CV 1.2%/2.0%.

## Build

```bash
cd models/qwen3.6-27b/llama-cpp/build
docker build -f Dockerfile.mtp -t llama-cpp-mtp:latest .
```

## Quick start

```bash
# Download MTP-capable GGUF (heads embedded in main file):
hf download havenoammo/Qwen3.6-27B-MTP-UD-GGUF \
  --include "Qwen3.6-27B-MTP-UD-Q3_K_XL.gguf" "27B_MTP.gguf" \
  --local-dir /mnt/models/gguf/qwen3.6-27b/havenoammo-mtp-q3kxl

# Run:
MODEL_DIR=/mnt/models/gguf docker compose \
  -f models/qwen3.6-27b/llama-cpp/compose/mtp/docker-compose.yml up -d
```

## Test plan

- [ ] `docker build -f Dockerfile.mtp -t llama-cpp-mtp:latest .` completes without linker errors
- [ ] `docker compose up -d` starts `llama-cpp-mtp-qwen36-27b` container; `curl http://localhost:8020/v1/models` lists the model
- [ ] `bash scripts/preflight.sh` auto-detects the MTP container and sets `URL`/`CONTAINER` correctly
- [ ] `bash scripts/bench.sh` produces narr/code TPS output consistent with the BENCHMARKS.md row (~50/66 TPS on a 3090 at 262K ctx)
- [ ] `SPEC_N_MAX=1 docker compose up` (smoke-test env override path)
- [ ] `DISABLE_THINKING=1 docker compose up` (smoke-test thinking suppression)


noonghunna (commented):
Thanks for the build — the three Dockerfile fixes (CUDA VMM stub symlink + `-DGGML_CUDA_NO_VMM=ON`, staged `.so` copy out of the multi-stage build, `libgomp1` runtime dep) are the exact footguns that only surface after you actually try to build, and they save the next person a half-day. Leaving as comment rather than request-changes since you've correctly kept this in draft — but flagging three things to land before un-drafting.

### 1. The 9K agentic cliff in your own BENCHMARKS row is the headline concern

> ⚠ Agentic stress shows TPS collapse at turn 7 (~9K accumulated ctx, 10.5 TPS vs 71–76 TPS at turns 1–6) — likely am17an fork behaviour with spec-decode at longer contexts

That's a far earlier cliff than the prior llama.cpp findings on master ([@lamentofhighborne's 131K + 91K NIAH on havenoammo, 164K + 91K NIAH on froggeric](https://github.com/noonghunna/club-3090/blob/master/BENCHMARKS.md)). And it's on the *exact* workload your other PR (#114) makes us care about. Before merging:

- `SPEC_N_MAX=1` A/B at turn 7 to confirm it's a spec-decode behavior (rejection thrashing at moderate ctx?) vs a recurrent-state issue
- `SPEC_N_MAX=0` (no MTP) on the same Docker image — if the cliff persists, it's the am17an fork's base llama.cpp diverging from master, not MTP specifically
- Capture `llama-server` log lines around the cliff (`grep -E "(speculative|reject|context)"`) for the diagnosis

The repo's prior position on MTP is "benched, not adopted" pending three triggers (q4_0 KV ✅, cross-rig data ✅, upstream PR merged ❌) — see `learnings/qwen3.6-35b-a3b.md` "Speculative decoding". The q4_0 + cross-rig boxes are now ticked, but the agentic cliff is a *new* concern that needs characterization before this becomes a default-shippable compose.

### 2. `compose/single/docker-compose.yml` adding `--spec-type ${SPEC_TYPE:-none}` may break stock image users

Stock `ghcr.io/ggml-org/llama.cpp:server-cuda` only has `--draft` / `--draft-max` for ngram speculation — the `--spec-type` flag is am17an-fork-only as far as I can tell. Passing `--spec-type none` to the stock image will fail at startup if the flag isn't recognized.

Two ways to fix:
- (preferred) Keep `compose/single/` untouched. Gate the `--spec-type` flag behind the `compose/mtp/` compose only, where the image is guaranteed to support it.
- Or, conditionally inject the flag via the entrypoint shell only when the image's `--help` lists it.

The `CACHE_TYPE_K`/`CACHE_TYPE_V` split with `${KV_TYPE:-q4_0}` legacy fallback is well-done and can stay.

### 3. Re-test the BENCHMARKS row at the same TGP as the 3090 baseline rows

The new row notes "**Performance mode TGP (~166W)** vs 3090-row baselines at 350W cap". That's a 2× power delta on what's effectively a head-to-head MTP-Docker comparison. Either:
- Re-bench at higher TGP (or note the wattage-normalized TPS-per-watt alongside the raw TPS), or
- Add a HARDWARE.md cross-reference next to the row so readers don't accidentally read "5090 Laptop = 3090 desktop" parity.

---

Once the agentic cliff is diagnosed (or scoped — e.g. "MTP is correct for short prompts; for sessions > 8K, prefer non-MTP variant") and the `compose/single/` regression risk is closed, happy to take another look. The Dockerfile work is genuinely useful regardless of where the MTP policy lands.

easel:
I’ve got claude working on diagnosing the cliff locally and will run some more benches as I have tokens to spend on it. My comment upstream got closed as off-topic so at this time I don’t expect much attention to the issue from them.

noonghunna:
Closing — superseded. This draft built MTP from `am17an/llama.cpp:mtp-clean` (PR #22673); the single-card MTP profile family shipped via #166 (merged 2026-05-20) — `llamacpp/mtp` + `llamacpp/mtp-vision`, the rolling mainline MTP image, thinking-off policy, and the registry wiring. The Dockerfile.mtp approach here is obsolete against the shipped path. Re-open only if a from-source MTP build path is wanted distinct from the rolling image.

PR #198, opened 2026-05-23
Fix verify-full to accept reasoning_content

laurimyllari:
## Summary

`verify-full.sh` always reports empty reasoning for llama.cpp based engines:
```
[6/8] Thinking / reasoning mode ...
  ✗ reasoning field empty (thinking mode didn't engage)
    → May indicate Genesis Patch 12 didn't land or chat_template_kwargs not honored. content='4.'
```
This is because the reasoning element is `reasoning_content` instead of `reasoning`.

Tested with vllm/default, llamacpp/mtp and ik-llama/iq4ks-mtp.

## Type of change

- [ ] New compose variant (`models/<model>/<engine>/compose/docker-compose.<name>.yml`)
- [ ] New patch / sidecar (`models/<model>/<engine>/patches/`)
- [ ] New script or tool (`scripts/`, `tools/`)
- [ ] New model (`models/<new-model>/`)
- [ ] Doc-only / typo
- [x] Other (describe): one-liner to accept either `reasoning` or `reasoning_content` in response

### DIFF
--- a/scripts/verify-full.sh
+++ b/scripts/verify-full.sh
@@ -341,7 +341,7 @@ check_thinking() {
 import sys, json
 d = json.load(sys.stdin)
 msg = d['choices'][0]['message']
-reasoning = msg.get('reasoning') or ''
+reasoning = msg.get('reasoning') or msg.get('reasoning_content') or ''
 content = msg.get('content') or ''
 finish = d['choices'][0].get('finish_reason')
 print(f'{len(reasoning)}|{len(content)}|{finish}|{(reasoning[:60] or \"(empty)\").replace(chr(10), \" \")}|{(content[:60] or \"(empty)\").replace(chr(10), \" \")}')


noonghunna:
Merged — thanks @laurimyllari, clean catch. ✅

Same cross-engine field-name split we've hit before: **vLLM emits `message.reasoning`, llama.cpp / ik_llama emit `message.reasoning_content`** (the streaming side has the matching `delta.reasoning` vs `delta.reasoning_content` dichotomy — already handled in the soak helper, and the bench token-counter). Your `reasoning or reasoning_content or ''` keeps the vLLM path byte-identical and unblocks the false-fail on llama.cpp-family engines.

Checked the rest of the tree before merging: `verify-full.sh` was the **last** `reasoning`-only extraction — `verify-stress.sh` and the soak helper already accept both — so this fully closes the gap. Landed on master.

Issue #197, opened 2026-05-23
[bug] CUDA error: out of memory with `llamacpp/default` and `llamacpp/mtp`


syangsao:
### What happened?

Seeing CUDA errors while using hermes agent against `llamacpp/default` and `llamacpp/mtp` which restart the docker image

```
12.19.916.474 I slot launch_slot_: id  0 | task 1820 | processing task, is_child = 0
12.19.916.528 W slot update_slots: id  0 | task 1820 | n_past = 127526, slot.prompt.tokens.size() = 127526, seq_id = 0, pos_min = 127525, n_swa = 0
12.19.916.529 I slot update_slots: id  0 | task 1820 | Checking checkpoint with [127415, 127415] against 127525...
12.20.020.693 W slot update_slots: id  0 | task 1820 | restored context checkpoint (pos_min = 127415, pos_max = 127415, n_tokens = 127416, n_past = 127416, size = 292.040 MiB)
12.28.236.257 I slot print_timing: id  0 | task 1820 | prompt processing, n_tokens =   4096, progress = 0.87, t =   8.32 s / 492.32 tokens per second
/app/ggml/src/ggml-cuda/ggml-cuda.cu:102: CUDA error
12.29.238.372 E CUDA error: out of memory
12.29.238.374 E   current device: 0, in function alloc at /app/ggml/src/ggml-cuda/ggml-cuda.cu:527
12.29.238.375 E   cuMemCreate(&handle, reserve_size, &prop, 0)
libggml-base.so.0(+0x1ac36)[0x7490189e2c36]
libggml-base.so.0(ggml_print_backtrace+0x21a)[0x7490189e30ba]
libggml-base.so.0(ggml_abort+0x15b)[0x7490189e329b]
/app/libggml-cuda.so(_Z15ggml_cuda_errorPKcS0_S0_iS0_+0xb5)[0x749006eab475]
/app/libggml-cuda.so(_ZN18ggml_cuda_pool_vmm5allocEmPm+0x21e)[0x749006ec74ae]
/app/libggml-cuda.so(_Z12launch_fattnILi256ELi8ELi8EEvR25ggml_backend_cuda_contextP11ggml_tensorPFvPKcS5_S5_S5_S5_PKiPfP6float2ffffjfi5uint3iiiiiiiiiiiliiliiiiilEimibbbi+0x1862)[0x7490072d5ac2]
[...]

PR #200, opened 2026-05-23
fix(verify-stress): add CTX_SIZE-scaled ceiling ladder (#199)

### LINKED ISSUE #199: verify-stress: NIAH ladder is fixed (~91K), doesn't scale to CTX_SIZE — high-ctx composes ship untested ceilings (#197)
## Problem

`verify-stress.sh`'s long-context NIAH ladder uses **fixed** rungs (~10K / 30K / 60K / 90K — header: *"needle ladder up to ~115K tokens"*) that **do not scale to the compose's `CTX_SIZE`**. So any compose with `CTX_SIZE > ~115K` is **never exercised above ~115K** — its top window is untested.

This ships false ceilings. A compose can advertise (and "verify-stress 7/7 + soak PASS") at a context it can't actually fill:

- The `llamacpp/mtp` compose ships `CTX_SIZE=262144` and passed verify-stress — but it only ever filled ~90K (35% of the window).
- **#197**: a user's hermes agent accumulated **127K tokens** → `launch_fattn` → `cuMemCreate: out of memory`. The 262K KV pool reserves the card to ~480 MiB free at boot; the flash-attention scratch at 127K fill doesn't fit. The real usable ceiling is ~100–127K, **nowhere near the advertised 262K** — and our deepest test stopped at ~90K, just short of the wall.

Root cause: we treat "boots + recalls a 90K needle" as "262K verified." The **fill ceiling was never measured** for any high-ctx compose.

## Fix — add a `CTX_SIZE`-scaled ceiling rung

1. **Scale the top NIAH rung to the live window.** Read `n_ctx` from the endpoint (`/props`), place the needle at ~50% depth in a haystack sized to **~0.90–0.95 × n_ctx**, recall it. Auto-scales (131K→~124K, 262K→~249K). A 262K compose would OOM here and **fail** — caught pre-ship.
2. **Gate on VRAM *margin*, not just pass/fail.** Capture peak free-VRAM during the ceiling rung; **fail/warn if margin < ~1–1.5 GB**, not only on a hard CUDA OOM. (Otherwise we re-bless configs that "pass" at 480 MiB free — fragile by definition.)
3. **Keep the existing fixed rungs** — they probe the DeltaNet "Cliff 2" prefill regression (an *architectural* cliff at ~50–60K, independent of `CTX_SIZE`). The new rung probes *window fillability* (a VRAM/attention-scratch ceiling that scales with `CTX_SIZE`). Complementary.

**Honesty caveat to bake into the margin:** a single-prompt NIAH peaks the same `fattn` scratch the agent hits, so it's a strong proxy — but sustained agent load *also* carries prompt-cache + context-checkpoint overhead (the #197 report showed a 292 MiB checkpoint restore). So single-shot is slightly *optimistic*; the ≥1–1.5 GB margin absorbs that delta.

## Plan
1. Branch + add the ceiling rung (scaled NIAH + VRAM-margin gate) to `verify-stress.sh`.
2. Run it across the single-card composes (`llamacpp/mtp` at 262K / 200K / 131K, etc.) → the **real ceilings + margins fall out** (the test *is* the measurement; no throwaway probing).
3. **Follow-on once it lands:** reset `CTX_SIZE` defaults + the SINGLE_CARD.md agent guidance to the measured-fully-fillable values (262K is a false ceiling; agent users likely belong at ~131K). Update the compose `Max ctx` header claims.

Motivated by #197.

noonghunna:
## Problem

`verify-stress.sh`'s NIAH ladder used **fixed rungs** (~10K / 30K / 60K / 90K) that **do not scale to the compose's `CTX_SIZE`**. Any compose with `CTX_SIZE > ~115K` was never exercised above ~90K — its top window was untested.

This shipped false ceilings. The `llamacpp/mtp` compose at `CTX_SIZE=262144` passed verify-stress but only ever filled ~90K (35% of the window). **#197**: a user's hermes agent accumulated 127K tokens → `launch_fattn` → `cuMemCreate: out of memory`. The 262K KV pool reserves the card to ~480 MiB free at boot; the flash-attention scratch at 127K fill doesn't fit. The real usable ceiling is ~100–127K, **nowhere near the advertised 262K** — and our deepest test stopped at ~90K.

## Fix — staggered ceiling ladder (probe 8)

Probe 8 is now a **ladder from ~95K → ~92% of n_ctx** in ~30K increments. Each rung:

1. Sends a NIAH request at that depth
2. Captures VRAM free before and after
3. Reports actual prompt tokens from `usage.prompt_tokens`
4. Stops at the first failure (OOM, recall miss, timeout, HTTP 400)

### Example output for a 262K compose

```
[8/8] Context ceiling ladder (staggered NIAH from ~95000 → ~0.92 × n_ctx) ...
    n_ctx=262144  ladder: 95000 → 125000 → 155000 → 185000 → 215000 → 241172 (6 rungs)
    VRAM free (ladder start): 23200 MB
    ✓ rung 1/6: target=95K  actual=94K tok (36%)  recalled 'crimson otter 42'  VRAM_free=18200MB
    ✓ rung 2/6: target=125K actual=124K tok (47%) recalled 'amber falcon 17'   VRAM_free=15800MB
    ✓ rung 3/6: target=155K actual=153K tok (58%) recalled 'violet iguana 33'  VRAM_free=12100MB
    ✓ rung 4/6: target=185K actual=183K tok (70%) recalled 'emerald narwhal 51' VRAM_free=8400MB
    ✗ rung 5/6: target=215K HTTP 500 (OOM at ~82% of n_ctx=262144)  VRAM_free=480MB

  ✗ ceiling ladder: wall at ~215000 tok — passed up to 183000 tok (70% of n_ctx=262144), then failed
    → Real fillable ceiling is ~183000 tok, not the advertised 262144.
```

## Configuration

| Env var | Default | Description |
|---|---|---|
| `CEILING_START_TOKENS` | 95000 | First rung target (just above probe 7's ~90K) |
| `CEILING_STEP_TOKENS` | 30000 | Increment between rungs |
| `CEILING_FRACTION` | 0.92 | Top rung as fraction of n_ctx |
| `VRAM_MARGIN_MB` | 1024 | Warn if free VRAM drops below this at ceiling |
| `SKIP_CEILING` | 0 | Set to 1 to skip the entire ladder |

## Plumbing

- `get_n_ctx()` — detects context window from `/props` (llama.cpp) → `/v1/models` max_model_len (vLLM/SGLang) → docker inspect fallback
- `get_vram_free_mb()` — sums `nvidia-smi memory.free` across all visible GPUs
- Ladder stops on first failure — that depth IS the wall (no point hammering a crashed engine)
- HTTP 400 (clean engine rejection) stops the ladder gracefully with skip
- Probe numbering updated from [N/7] to [N/8] throughout

## Tests

`scripts/tests/test-verify-stress-ceiling.sh`: **26 tests** covering:
- `get_n_ctx()` with llama.cpp /props, vLLM /v1/models fallback, total failure, 512K compose
- `get_vram_free_mb()` dual-GPU sum, single-GPU
- Ladder rung computation for 262K, 131K, 200K, 512K, 32K (skip), custom step/fraction
- `SKIP_CEILING=1` and `SKIP_LONGCTX=1` skip the ladder
- All 8 probe headers [N/8] present in output

```
$ bash scripts/tests/test-verify-stress-ceiling.sh
Results: 26 passed, 0 failed
test-verify-stress-ceiling: ok
```

## Why stagger instead of a single ceiling shot?

A single shot at 92% tells you either pass or "it broke somewhere between 90K and 241K." That is a huge gap. Staggering at ~30K increments:

1. **Finds the exact wall** — you know the last depth that worked and the first that did not
2. **Shows the VRAM slope** — you can extrapolate for different workloads
3. **Gracefully stops** — no point hammering an OOM'd engine with deeper rungs
4. **Configurable precision** — bump `CEILING_STEP_TOKENS=15000` for finer resolution

Closes #199.

### DIFF
--- a/models/qwopus3.6-27b-v2/llama-cpp/compose/single/mtp.yml
+++ b/models/qwopus3.6-27b-v2/llama-cpp/compose/single/mtp.yml
@@ -1,0 +1,99 @@
+# ===========================================================================
+# Profile (at-a-glance):
+#   Model:     Qwopus3.6-27B-v2-MTP (Jackrong) — Qwen3.6-27B fine-tune, distilled
+#              on Claude-Opus-4.6/4.7 reasoning traces ("TraceInversion"), with a
+#              custom-processed MTP head. Q4_K_M GGUF. Text-only (no mmproj).
+#   Engine:    llama.cpp (ghcr.io/ggml-org/llama.cpp:server-cuda, pinned b9246)
+#   Topology:  Single 3090 (TP=1)
+#   Drafter:   built-in MTP head, n=2 (--spec-type draft-mtp)
+#   KV:        q4_0 K + q4_0 V (densest mainline, Ampere-fast)
+#   Vision:    NO (repo ships no mmproj)
+#   Template:  native (GGUF-embedded) + thinking-off by default — see REASONING note
+#   Max ctx:   131072 default (@ -ub 1024); base Qwen3.6 reaches 262K @ -ub 512
+#   Status:    🧪 EVAL — unvalidated on this rig. Same DeltaNet arch + cliff
+#              behavior as base Qwen3.6-27B (cliff-immune ggml memory model).
+#
+#   ⚠ FIRST-BOOT CHECK (the whole point of this model is the MTP head):
+#     Jackrong used a CUSTOM Qwen-MTP split/merge methodology. Confirm our b9246
+#     build actually ENGAGES it — `docker logs <c> | grep -iE 'draft|mtp|spec'`
+#     should show the MTP/spec-decode path init + a non-zero accept rate under
+#     load. If MTP does NOT engage, this is just a slower Qwen3.6 fine-tune with
+#     none of the claimed ~1.66x speedup — fall back to plain decode + flag it.
+#
+#   Card claim (GB10, NOT Ampere — won't transfer in absolute terms): 10.46 vs
+#   6.29 t/s base (~1.66x via MTP), -27.7% completion tokens (more compact CoT),
+#   30/30 on their bench. Verify the *ratio* + token-compactness on our rig, not
+#   the absolute t/s.
+#
+# Quick start:
+#   1. Get the GGUF (SHA-verified):
+#        WEIGHTS=... (manual) hf download Jackrong/Qwopus3.6-27B-v2-MTP-GGUF \
+#          Qwopus3.6-27B-v2-MTP-Q4_K_M.gguf \
+#          --local-dir $MODEL_DIR/qwopus3.6-27b-v2-gguf/jackrong-mtp-q4km
+#   2. From this directory (single card → stop any other :8020 container first):
+#        MODEL_DIR=/your/models/dir docker compose -f mtp.yml up -d
+#   3. curl http://localhost:8020/v1/models  → should list the model.
+#
+# Override defaults via .env or shell:
+#   MODEL_DIR              host dir to mount as /models  (default: ../../../../../models-cache)
+#   GGUF_FILE             path under /models  (default: qwopus3.6-27b-v2-gguf/jackrong-mtp-q4km/Qwopus3.6-27B-v2-MTP-Q4_K_M.gguf)
+#   CTX_SIZE             KV pool size   (default: 131072 @ -ub 1024; for 262K set CTX_SIZE=262144 UBATCH_SIZE=512)
+#   BATCH_SIZE           llama.cpp -b   (default: 4096)
+#   UBATCH_SIZE          llama.cpp -ub  (default: 1024; lower to 512 to fund 262K)
+#   NP                   parallel slots (default: 1 — see ⚠ below)
+#   KV_TYPE              K and V quant  (default: q4_0)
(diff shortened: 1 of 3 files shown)


noonghunna:
Fixed the mock-vs-reality gap. Three changes in the latest push:

**1. `/props` parse — checks nested first, falls back to top-level:**
```python
nested = d.get('default_generation_settings', {})
v = nested.get('n_ctx') if isinstance(nested, dict) else None
if v is None:
    v = d.get('n_ctx', 0)
```

**2. Docker fallback — scrapes llama.cpp `-c`/`--ctx-size` too:**
```python
m = re.search(r'--max-model-len[=\s]+(\d+)', args)  # vLLM/SGLang
if not m:
    m = re.search(r'(?:^|\s)(?:-c|--ctx-size)[=\s]+(\d+)', args)  # llama.cpp
```

**3. Unit-test mock uses the real nested shape:**
```json
{"default_generation_settings":{"n_ctx":262144,"temperature":0.8}}
```
Also added tests for: top-level n_ctx (fork/old versions), nested-wins-over-top-level, 512K compose. 28 tests total, all pass.

Ready for live re-test against the booted 262K compose.

PR #201, opened 2026-05-23
fix(compose): single-card MTP CTX_SIZE default 262144 → 200000 (llama.cpp + ik_llama)

noonghunna:
## What

Lowers the **single-card MTP** context default from `262144` to `200000` on both engines:

| File | Knob | Before | After |
|---|---|---|---|
| `models/qwen3.6-27b/llama-cpp/compose/single/mtp.yml` | `-c` | 262144 | **200000** |
| `models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp.yml` | `--ctx-size` | 262144 | **200000** |

(`llamacpp/default` is an alias for the llama.cpp file, so this covers both `llamacpp/default` and `llamacpp/mtp`.)

## Why — "boots ≠ fills"

On a single 24 GB 3090, `CTX_SIZE=262144` is a **false ceiling**: it boots, pre-reserves its KV pool, and passes `verify-stress` (91K needle) + soak — but a prompt only **fills to ~125K** before OOMing. The wall is **not** the GDN cliff (Cliff 2) and **not** the reserved KV pool — it's the **flash-attention transient scratch at high fill** (`launch_fattn` / `cuMemCreate`), which scales with *populated* context, not what was reserved at boot.

Measured via the #200 ceiling-ladder probe (Q4_K_M, q4_0 KV, MTP n=2, `-ub 512`, single 3090):

| `CTX_SIZE` | Fills to | % declared | Free VRAM at top | Verdict |
|---|---|---|---|---|
| 262144 | ~125K | 47% | walls at 155K rung | false ceiling |
| **200000** | **183K** | **91%** | **1177 MB** | ✅ clears the 1024 MB margin gate |
| 224000 (math) | fills | — | ~400 MB | knife-edge, below margin |

So the old default advertised more context than it could fill — exactly the OOM @syangsao hit in #197.

## Scope notes

- **llama.cpp** change is the directly-measured config. ✅
- **ik_llama** change is now **laddered on-rig (2026-05-23)** — 200K is ik's *measured* max-safe ceiling. The ~0.5 GB weight lean does **not** buy more ctx: ik forces `n_ubatch=n_batch` (=4096) vs mainline `-ub 512`, so its larger graph buffer eats the lean and at equal ctx ik has ~the same free VRAM as mainline. Ladder (all rungs recall; only the 1024 MB margin gate fails): **224K → fills 205K, 477 MB at ceiling ✗; 210K → fills 193K, 863 MB ✗; 200K → ≈1131 MB ✓**. So the lean is margin-at-equal-ctx, not more-ctx — exactly what discussions/184 concluded. Also fixes a pre-existing header contradiction (`262144`/`131072`/`131072` claimed as "default" in three places). [commit `1d93343`]
- **Untouched:** the vision composes (`mtp-vision.yml` / `iq4ks-mtp-vision.yml`, 160K — mmproj competes for headroom, separately validated) and `iq4ks-two-stage.yml` (131K). Bumping vision up to 200K would risk OOM with images loaded.

## Follow-ups

- [x] Run the ceiling ladder on ik_llama — **done 2026-05-23: confirms 200K** (210K/224K both miss the 1024 MB margin: 863 / 477 MB). The lean is margin-not-ctx.
- [ ] Update @syangsao on #197 with the measured 200K number once merged.

Forensic detail: `docs/CLIFFS.md` 2026-05-23 note.

### DIFF
--- a/models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp.yml
+++ b/models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp.yml
@@ -11,14 +11,18 @@
 #   Template:  native (GGUF-embedded) — A/B'd froggeric v19 here, native won
 #              the 8-pack 103 vs 99 (toolcall TIED 9/9); see header note below
 #   Vision:    NO (text-only this profile; mmproj is a separate axis → -vision.yml)
-#   Default ctx: 262144 — full native ctx (n_ctx_train) on ONE 3090, q4_0 KV.
-#              Bumped from 131072 → native max 2026-05-21, keeping -ub 1024 (NOT
-#              the -ub 512 the 21.5 GB datapoint below used). For lower VRAM /
+#   Default ctx: 200000 — cross-engine-aligned with llama-cpp mtp.yml's measured
+#              max-safe fill (2026-05-23), q4_0 KV, -ub 1024. Was 262144 (full
+#              n_ctx_train) but that is a *boot* ceiling, not a verified *fill*
+#              ceiling on ik — see the "boots ≠ fills" note below. For lower VRAM /
 #              cross-rig-safe boot, set CTX_SIZE=131072.
 #   MAX ctx:   262144 (model native ceiling; n_ctx_train = n_ctx_orig_yarn = 262144).
-#              VRAM: default -ub 1024 @ 262144 → 22.5 GB / 24 (verified 2026-05-21,
-#              KV pre-allocated so this is steady-state; ~2.1 GB headroom — fine on a
-#              dedicated card, tight if anything else shares GPU 0). -ub 512 @ 262144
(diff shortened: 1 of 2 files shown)


noonghunna:
Closed the ik follow-up: **ceiling-laddered ik on-rig 2026-05-23**. **200K confirmed** as ik's measured-safe ceiling — 210K (863 MB free at ceiling) and 224K (477 MB) both miss the 1024 MB agent-safety margin. The ~0.5 GB weight lean doesn't buy more ctx because ik forces `n_ubatch=n_batch=4096` (vs mainline `-ub 512`), so the larger graph buffer eats the lean — at equal ctx ik has ~the same free VRAM as mainline. Compose header updated (`1d93343`) + PR body updated. Both legs now measured-safe at 200K; ready to merge.

PR #206, opened 2026-05-23
fix(#169): distinct default container_name per llama.cpp/ik single variant

### LINKED ISSUE #169: llama.cpp single composes share default container_name — discourages concurrent variants
## Problem

All three llama.cpp single-card composes ship with the same default `container_name`:

```yaml
# models/qwen3.6-27b/llama-cpp/compose/single/docker-compose.yml  (llamacpp/default)
# models/qwen3.6-27b/llama-cpp/compose/single/mtp.yml             (llamacpp/mtp)
# models/qwen3.6-27b/llama-cpp/compose/single/mtp-vision.yml      (llamacpp/mtp-vision)
    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b}"
```

This makes them **mutually exclusive at the default-name level** — bringing up a second one while a first is running fails with a docker name conflict. The `ESTATE_CONTAINER` env override exists as an escape valve, but it isn't documented anywhere user-facing.

Surfaced 2026-05-20 while triaging a `gpu-mode status` reporting bug post-v0.8.3 (separate fix landed in commit `db9c5e1`). Two variants were running simultaneously on the rig:
- `:8020` = `llama-cpp-qwen36-27b` (default name)
- `:8030` = `llama-cpp-qwen36-27b-vision` (set via `ESTATE_CONTAINER=...-vision` override)

The override worked, but discoverability is poor — a user trying to run `llamacpp/default` + `llamacpp/mtp-vision` simultaneously will hit "name in use" with no docs guide.

## Related

The same pattern exists across Gemma 4 composes — they share ports `:8032` (3 composes: dflash, int8, dflash-int8) and `:8033` (2 composes: awq, bf16), all with the same `ESTATE_CONTAINER`/`ESTATE_PORT` override pattern. Fixing this is best done as a stack-wide convention, not just for the llama.cpp single family.

### Option A — Distinct default container names per variant (preferred)

Update each compose to default to a name that encodes the variant:

```diff
# llamacpp/default → unchanged
-    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b}"
+    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b}"          # default

# llamacpp/mtp →
-    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b}"
+    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b-mtp}"

# llamacpp/mtp-vision →
-    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b}"
+    container_name: "${ESTATE_CONTAINER:-llama-cpp-qwen36-27b-mtp-vision}"
```

Plus equivalent for Gemma where ports collide.

**Pros:** Two variants Just Work simultaneously without env override. Matches the natural compose-name -> container-name mapping users expect.

**Cons:** Existing benchlocal-cli sandboxes / scripts that hardcode the container name `llama-cpp-qwen36-27b` for `docker logs` etc. would need an update (audit needed).

### Option B — Document the `ESTATE_CONTAINER` / `ESTATE_PORT` env overrides

Add a section to `docs/SINGLE_CARD.md` and/or `docs/EXAMPLES.md` documenting:

```bash
# Run two llama.cpp variants simultaneously (e.g., default on :8020 + mtp-vision on :8030):
bash scripts/switch.sh llamacpp/default
ESTATE_CONTAINER=llama-cpp-qwen36-27b-mtp-vision ESTATE_PORT=8030 \
    bash scripts/switch.sh llamacpp/mtp-vision
```

**Pros:** Zero code change; just docs.

**Cons:** Users still have to remember the override pattern; discoverability still requires reading docs.

## Recommendation

**A is the right fix.** B alone is a documentation workaround for a UX gap that should be closed at the source. The container-name audit (option A's caveat) is the one piece of real work — once that's clean, B becomes unnecessary.

## Acceptance

- [ ] Each llama.cpp single variant ships a distinct default `container_name`
- [ ] Same audit for Gemma 4 variants sharing `:8032` and `:8033`
- [ ] Update gpu-mode :8030 detection logic — currently uses `llama-cpp-*` glob, would still work but the comment in `scripts/gpu-mode.sh` should be updated to reflect distinct names
- [ ] Verify benchlocal-cli, rebench-full, soak-test, and report.sh don't hardcode the old name (audit list)
- [ ] At least one example of "run two variants simultaneously" in `docs/SINGLE_CARD.md`

noonghunna:
Closes #169.

All llama.cpp + ik single composes shipped the **same** default `container_name`, so bringing up a second variant failed with a docker name conflict (the `ESTATE_CONTAINER` override existed but was undocumented).

### Change
Distinct default per non-base variant; the base workhorses keep their names to avoid breaking hardcoded refs / benchlocal `docker logs <base>`:

| Variant | Default container_name |
|---|---|
| `llamacpp/mtp` (= `llamacpp/default`) | `llama-cpp-qwen36-27b` *(unchanged)* |
| `llamacpp/mtp-vision` | `llama-cpp-qwen36-27b-vision` |
| `ik/iq4ks-mtp` | `ik-llama-qwen36-27b` *(unchanged)* |
| `ik/iq4ks-mtp-vision` | `ik-llama-qwen36-27b-vision` |
| `ik/iq4ks-two-stage` | `ik-llama-qwen36-27b-two-stage` |

- `launch.sh`: track the two renames in `LAUNCH_DEFAULT_CONTAINER`, **and add the missing `ik-llama/iq4ks-two-stage` entry** (latent bug — it fell back to the wrong default container).
- `gpu-mode.sh`: comment-only update.
- `docs/SINGLE_CARD.md`: "run two variants at once" snippet (ports still default to `8020`, so the second one uses `ESTATE_PORT`, per the issue's example).

### Audit (the issue's caveat — all clear)
The autodetect globs in `preflight/soak/switch/health/update` are **prefix-anchored** (`^llama-cpp-qwen36-27b`, no `$`) and `gpu-mode` uses `== llama-cpp-*`, so suffixed names still match — verified, no logic change needed.

### Out of scope / notes
- **Gemma 4** composes already carry distinct container_names; their shared *ports* (`:8032` ×3, `:8033` ×2) are `ESTATE_PORT`-managed, same pattern as Qwen — left as-is.
- **benchlocal-cli** (separate repo) targets the *base* names, which are unchanged → unaffected.

### Validation
- `docker compose config` resolves each distinct name ✓
- every new name still matches the autodetect glob ✓
- `bash -n` (launch.sh, gpu-mode.sh) + `test-launch-compat.sh` pass ✓

### DIFF
--- a/docs/SINGLE_CARD.md
+++ b/docs/SINGLE_CARD.md
@@ -244,6 +244,18 @@ GPU_MEMORY_UTILIZATION=0.80 \
 
 Generally prefer **dropping `MAX_MODEL_LEN` first** (clean KV budget reduction, predictable behavior) over `GPU_MEMORY_UTILIZATION` (interacts with profiling overhead in non-obvious ways). Drop both if your envelope is really tight.
 
+### Running two llama.cpp / ik variants at once
+
+Each variant ships a **distinct default `container_name`** (`llamacpp/mtp` → `llama-cpp-qwen36-27b`, `llamacpp/mtp-vision` → `llama-cpp-qwen36-27b-vision`; ik's `iq4ks-mtp` / `-vision` / `-two-stage` likewise), so they no longer collide on the docker name. They **do** still share the default host port `8020`, so give the second one a port with `ESTATE_PORT` (and, if you want a custom name, `ESTATE_CONTAINER`):
+
+```bash
+# text workhorse on :8020 + vision variant on :8030, side by side
+bash scripts/switch.sh llamacpp/mtp
+ESTATE_PORT=8030 bash scripts/switch.sh llamacpp/mtp-vision
+```
+
+(Whether both fit depends on your VRAM envelope — two full-context single-card servers won't co-reside on one 24 GB card; drop `CTX_SIZE` on one, or use this across two cards.)
+
 ---
 
 ## Quick start
--- a/models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp-vision.yml
+++ b/models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp-vision.yml
@@ -94,7 +94,7 @@
 services:
   ik-llama-qwen36-27b-iq4ks-mtp-vision:
     image: ${IK_LLAMA_IMAGE:-ghcr.io/ikawrakow/ik-llama-cpp:cu13-server}
-    container_name: "${ESTATE_CONTAINER:-ik-llama-qwen36-27b}"
+    container_name: "${ESTATE_CONTAINER:-ik-llama-qwen36-27b-vision}"
     restart: unless-stopped
     ports:
       - "${ESTATE_PORT:-${PORT:-8020}}:8080"
--- a/models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp-vision.yml
+++ b/models/qwen3.6-27b/ik-llama/compose/single/iq4ks-mtp-vision.yml
@@ -14,7 +14,7 @@
 #              margin. REQUIRES -b 1024 (see below): a full 2048²/4M-pixel image
 #              encodes at ~23.65 GB. mmproj + encode buffers eat the VRAM the
 #              text profile spends on KV, so the vision ceiling (~180K) sits well
-#              below the text profile's 262K. Push higher only if you don't need
+#              below the text profile's 200K. Push higher only if you don't need
 #              prefill headroom — see VRAM RESIZING below.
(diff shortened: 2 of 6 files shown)


noonghunna:
Folded in a small follow-on (commit just pushed): five stale `262K → 200K` cross-refs in these same single-card compose headers (e.g. *"mtp.yml → … 262K"*, *"vs the text profile's 262K"*) that the earlier 262K→200K doc sweep missed because it excluded `.yml` files. The "262144 native max" + "boots ≠ fills" explanations are left intact. Pure header comments — no functional change; all three composes still `docker compose config`-parse clean.

PR #265, opened 2026-05-30
feat(registry): slug health/availability flag

noonghunna:
## What

Adds a lifecycle **health flag** to every compose slug in the registry, so `switch.sh --list`, launch, and switch are no longer blind to a compose's lifecycle stage. Implements design §12.5 (PR-A) — the status flag only; the model-default resolver / `<model>/default` / `--set-default` are a separate PR-B.

- **`compose_registry.py`**: `_entry()` gains a keyword-only `status` (default `production`) + `status_note`, validated against the canonical enum: `production · caveats · experimental · preview · upstream-gated · deprecated`. New `compose_header_status()` helper maps a compose's profile-schema `Status:` emoji to that enum.
- **Header sweep**: normalized *every* compose `Status:` header to a canonical enum value (per the locked §13 decision "sweep all"). Non-conforming strings fixed — e.g. the Genesis compose's `✅ Working (with Genesis)` → `⏸️ Upstream-gated`, `⛔ TOMBSTONED` → `🗑️ Deprecated`, `🔵 v0.7.3 ONBOARDING/PRIMARY` → `🧪 Experimental`, `🔵 PREVIEW` → `👁️ Preview`, `⭐ Code-optimized` / `✅ Shipped` / `✅ Working` → `✅ Production`.
- **`registry-emit.sh`**: emits `status` + `status_note` as the last two fields of the `VARIANT` row; both loaders (switch + launch) and the two parity tests read the extended field list.
- **`switch.sh --list`**: status marker — `production` unmarked · `caveats` → `(caveats)` · the (NA) set → `(NA: <word>)`. The model·topology grouping from #264 is preserved.
- **Launch/switch gate**: `production` launches silently; `caveats` launches with a one-line notice; the (NA) set (`experimental/preview/upstream-gated/deprecated`) warns + requires `--force` (the existing flag). `launch.sh` surfaces the flag before delegating to `switch.sh`.
- **Drift-guard test** (`scripts/tests/test-compose-status-drift.sh`): asserts every registry `status` ∈ enum, every compose `Status:` header maps to the enum, and registry status is consistent with the compose header — so they can't silently diverge again.

## Why

Status previously lived only in compose-header comments, so the launchers couldn't see it — and the headers drifted. The Genesis dual compose declared "✅ Working (with Genesis)" while its pin is parked/drifted, so a user could unknowingly boot a config that won't run clean. This makes lifecycle a first-class, drift-guarded field; it's a prerequisite for the resolver (which must skip non-functional defaults) and a safety fix on its own.

## Re-flagged slugs (20 of 45 → (NA) or caveats)

| Status | Slugs |
|---|---|
| `upstream-gated` | `vllm/dual-tq3-mtp-genesis`, `vllm/gemma-mtp-tp1` |
| `deprecated` | `vllm/dual-tq3-mtp` |
| `preview` | `vllm/dual-qwopus-bf16mtp`, `vllm/qwen-a3b-preview-single` |
| `caveats` | `vllm/long-text`, `vllm/long-text-no-mtp`, `vllm/long-vision`, `vllm/dual-carnice-bf16mtp` |
| `experimental` | `vllm/dual-bf16`, `vllm/dual-int8`, `llamacpp/bounded-thinking`, the 5 `ik-llama/prism-pro-dq-*`, `ik-llama/apex-mtp-compact`/`-long`/`-quality-dual`, the 4 `vllm/gemma-a4b*` |

The remaining 25 stay `production`.

## Verification (CI condition)

- Registry loads; **45 entries** (status adds no entries).
- `python3 tools/kv-calc.py --calibration` → `Overall: 17/17 (100%)`.
- `bash scripts/switch.sh --list` renders with markers, still grouped by model·topology, still 45 slugs.
- `bash -n scripts/switch.sh` / `launch.sh` / `registry-emit.sh` — OK.
- Full suite (`for t in scripts/tests/test-*.sh; do bash "$t"; done`): **35 pass / 1 fail**, the only failure being the pre-existing `test-submit-bench.sh` (red on clean master too — fixture-dependent). New `test-compose-status-drift.sh` and both parity tests pass.
- Gate behavior confirmed: NA slug blocked without `--force`, caveats launches with a notice, production launches silently, `--force` overrides.


noonghunna:
Added model/variant counts to `--list` (follow-on, same display layer):

- Header: `Models: 4 · variants: 45 (21 production · 4 caveats · 20 NA)`
- Per-model group header: `qwen3.6-27b  (29 variants)` etc.
- Widened the serving-artifact column so the `(NA: …)` / `(caveats)` markers align (the cosmetic shift noted earlier).

Re-verified: parity + drift-guard + full suite green (sole red = pre-existing `test-submit-bench`).

Issue #290, opened 2026-06-01
[bug] Regression for enablement of PCIE P2P after NVLink autodetection script addition


hlo-world:
### What happened?

In https://github.com/noonghunna/club-3090/issues/158 I posted a benchmark which I thought had PCIE P2P enabled. @noonghunna brought up that I had PCIE P2P enabled at a firmware level, but it was not enabled at the vLLM level.

Following the advice to reproduce https://github.com/noonghunna/club-3090/issues/95 (and subsequently I found https://github.com/noonghunna/club-3090/issues/91 to try to reproduce as well), I was **unsuccessful** in enabling PCIE P2P as suggested in the advice.

In the end, I have tried many combinations of env vars in different locations, as well as the cuda.py hack, and none of it yielded the desired result. Ultimately, I observed that:
- @aaronlockhartdev had completed his bench and submissions on May 7th
- I had began my bench and submissions on May 18th
- Between that time, a commit https://github.com/noonghunna/club-3090/commit/e00626a50e92fe44855c410b86ae3f33ace82f29 for NVLink autodetection was added on May 14th.

The commit may work as intended in its own right, however for the case if PCIE P2P were to be enabled by setting ```NCCL_P2P_LEVEL=PHB``` in .env, the script overrides it.

```
scripts/detect_nvlink.sh
[...]
# Apply environment overrides based on detection result
if [ "$_NVLINK_ENABLED" -eq 1 ]; then
  [...]
else
  export NCCL_P2P_DISABLE=1
> unset NCCL_P2P_LEVEL 2>/dev/null || true
  export PYTORCH_CUDA_ALLOC_CONF="${PYTORCH_CUDA_ALLOC_CONF:-expandable_segments:True,max_split_size_mb:512}"
  echo "[nvlink] NVLink DISABLED — NCCL_P2P_DISABLE=1, custom all-reduce OFF, expandable_segments ON"
fi
```

Side note, AFAIK there is currently no "env var" way to enable ```custom-all-reduce``` either.

I haven't submitted a PR (yet?) to this issue (nor am I pushing for repo owner to fix) because there's several ways to go about this, and don't know what the repo stakeholders would prefer:
1. low-hanging but messy, incomplete: ```detect-nvlink``` stores incoming ```NCCL_P2P_LEVEL``` and ```NCCL_P2P_DISABLE``` and restores it on the ```else```
[...]

PR #291, opened 2026-06-01
fix(nvlink): add NVLINK_MODE=pcie_p2p for PCIe P2P without NVLink (#290)

noonghunna:
Fixes the regression in #290 — thanks @hlo-world for the precise diagnosis (commit, exact lines, *and* fix options).

## Problem
`scripts/detect_nvlink.sh` (added in e00626a) models only two states — NVLink-present (`NCCL_P2P_LEVEL=NVL` + custom all-reduce on) and no-P2P (`NCCL_P2P_DISABLE=1`). On a no-NVLink rig, `auto`/`force_off` do `export NCCL_P2P_DISABLE=1; unset NCCL_P2P_LEVEL` — **clobbering a deliberate `NCCL_P2P_LEVEL=PHB`** set for patched-driver PCIe P2P (tinygrad/geohot). `force_on` sets `NVL` (NVLink-specific — excludes PCIe). And the entrypoint hardcodes `--disable-custom-all-reduce` in the non-NVLink branch, so there's no way to enable custom all-reduce for P2P.

## Fix
A third explicit-opt-in mode, `NVLINK_MODE=pcie_p2p`:
- `NCCL_P2P_LEVEL=PHB` (or your own `NCCL_P2P_LEVEL` if set),
- P2P **not** disabled,
- `_NVLINK_ENABLED=1` → the **existing** entrypoint enables custom all-reduce (drops `--disable-custom-all-reduce`).

Reuses the `_NVLINK_ENABLED` signal, so **no compose changes** — any dual compose picks it up via `NVLINK_MODE=pcie_p2p` in env.

## Safety / testing
Purely additive — `auto` / `force_on` / `force_off` produce identical env to before (verified):

| mode | `_NVLINK_ENABLED` | `NCCL_P2P_LEVEL` | `NCCL_P2P_DISABLE` |
|---|---|---|---|
| **pcie_p2p** (new) | 1 | PHB (or your value) | unset |
| force_on | 1 | NVL | unset |
| force_off / auto-no-NVLink | 0 | unset | 1 |

Full test suite **41/41 green**; `test-hwdetect.sh` passes.

## ⚠️ Needs your validation, @hlo-world
We **can't verify this on our hardware** — our reference rig is PCIe-only with no P2P capability (no patched driver), so we can't confirm P2P + custom all-reduce actually *engage* end-to-end. That blind spot is what let the original regression ship. Could you check out this branch and run a dual compose with `NVLINK_MODE=pcie_p2p` on your patched-driver rig?
- Does vLLM now recognize P2P over PCIe?
- Stable with custom all-reduce on (no hang/crash)?
- A quick TPS delta vs `force_off` would be 🔥.

If it works we merge; if a knob needs changing (default level, or custom-all-reduce should stay off for some setups), your call. Thanks again!

### DIFF
--- a/scripts/detect_nvlink.sh
+++ b/scripts/detect_nvlink.sh
@@ -1,9 +1,21 @@
 #!/bin/bash
-# NVLink auto-detection + override. Sources NVLINK_MODE from env (default: auto).
-# Exports: _NVLINK_ENABLED (0 or 1), sets NCCL/PYTORCH env vars accordingly.
+# NVLink / PCIe-P2P detection + override. Sources NVLINK_MODE from env (default: auto).
+# Exports: _NVLINK_ENABLED (0/1 — "fast P2P interconnect available → custom all-reduce ON")
+# and sets NCCL/PYTORCH env vars accordingly.
 # Handles 2-GPU setups (single NVLink bridge) and N-GPU setups (e.g. 2 bridges on 4 cards).
+#
+# NVLINK_MODE values:
+#   auto       — detect NVLink via nvidia-smi topo (default). No NVLink => P2P off.
+#   force_on   — assert NVLink present (NCCL_P2P_LEVEL=NVL).
+#   force_off  — no P2P at all (NCCL_P2P_DISABLE=1).
+#   pcie_p2p   — PCIe P2P WITHOUT NVLink (e.g. a patched consumer-GPU driver — the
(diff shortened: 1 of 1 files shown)


hlo-world:
Tested and confirmed that it addressed the issue https://github.com/noonghunna/club-3090/issues/290

PR #301, opened 2026-06-03
docs(benchmarks): full pipeline results — Carnice/PRISM/byteshape (2026-06-03)

Rhonstin:
## Summary

Full validation run via `scripts/compare-prism-vs-carnice.sh` — three single-card models through the complete pipeline (verify-full + verify-stress + bench + hermesagent-20 + --medium + soak-continuous) on 1× RTX 3090.

### Carnice 35B-A3B APEX-MTP-I-Quality (mainline llama.cpp, thinking ON)
- **hermesagent-20: 17/20 (85%)** — new rig record; only 3 fails (HA-04 Docker recall, HA-16 message routing, HA-17 parallel delegation)
- **--medium: 63/75 (84%)**: toolcall 87% · instructfollow 100% · structoutput 93% · dataextract 80% · reasonmath 60%
- **bench: 139/141 TPS** (CV 0.1%), PP 3967 tok/s
- verify-full 8/8 ✅, verify-stress 7/8 ⚠️ (VRAM margin 356 MB at 131K ctx ceiling)
- **soak FAIL**: 2× HTTP 500 — VRAM at 97% (23770/24576 MiB); ctx≤80K recommended for agent workloads

### PRISM-PRO-DQ 27B (mainline + `--spec-type draft-mtp`, thinking ON)
- **hermesagent-20: 13/20 (65%)** — mainline MTP gives +5 pp vs ik-llama run (60%)
- **--medium: 60/75 (80%)**: toolcall/instructfollow/structoutput all 93%, **dataextract 6/15 (40%) — critical weakness** on numeric field extraction
- bench: 49/49 TPS, PP 1078 tok/s
- verify-stress 8/8 ✅, soak PASS (p50 70.8 TPS)

### byteshape IQ4_XS 35B-A3B (ik-llama cu12 + MTP)
- **--medium: 71/75 (95%)** — highest --medium score on this rig; toolcall 100%, instructfollow 100%, reasonmath 100%
- bench: narr 126/129 TPS, code 148/155 TPS, PP 2277 tok/s
- soak PASS (p50 194.6 TPS, 0 errors)
- ⚠️ cu12 override required — upstream bumped to cu13 in latest intake, RTX 3090 has no Forward Compat

## Test plan
- [x] All three models ran through full pipeline
- [x] Results reproduce/extend PR #299 byteshape numbers
- [x] Carnice soak failure documented with root cause (VRAM %) and mitigation (ctx≤80K)
- [x] PRISM dataextract weakness documented


noonghunna:
Thanks for the thorough run, and for documenting the rough edges honestly (the Carnice soak failure at 97% VRAM and PRISM's dataextract weakness are exactly the kind of findings we want surfaced, not hidden).

We're going to **close this one** — not on the content, but on the shape. It bundles a lot of independent things into a single PR: a `cu13 → cu12` engine-image change on several shipped composes, a `repeat-penalty 1.0 → 1.1` sampling-default change, a batch of raw-`docker run` research benchmark rows for ~8 models, Genesis pin/UPSTREAM doc updates, and edits to core vLLM composes + scripts. We can't review, validate, or track those as one unit — and a problem in any one of them blocks all the rest.

Going forward, please **keep each PR tied to a single model or a single feature/concern** (now noted in CONTRIBUTING). Smaller, focused PRs land much faster here because each can clear its own validation gate on its own timeline.

How I'd slice this one — most valuable first:

1. **`cu13 → cu12` engine image** — submit this on its own. It's the highest-value change in here: if `cu13-server` really is degrading on RTX 3090 (no Forward Compat, driver supports CUDA 13.0 not 13.1), it affects every shipped ik-llama compose, and it may explain an MTP failure we hit independently this week. We'll validate it on our rig and land it quickly. Please include your driver version + the before/after symptom (CPU fallback? kernel error?) so we can reproduce.
2. **`repeat-penalty 1.0 → 1.1`** — its own PR with a short A/B, since it's a behavioral default change on shipped composes.
3. **Research benchmark rows** — happy to take these as a separate **docs-only** PR, clearly marked exploratory (your "quality-research only" labelling is the right call). Several of these models aren't production-ready yet, so the rows are useful as a survey but shouldn't imply induction.
4. **Genesis v7.72.5 / UPSTREAM** — please hold this; it touches a pin decision we've currently parked pending an upstream stable tag, so we'll handle that one ourselves.

Genuinely appreciate the effort — the cu12 finding especially. Reopen #1 as its own PR and we'll move fast on it.

noonghunna:
Quick correction on my point #1 above — I checked our rig after posting.

Our driver is `610.43.02`, which supports CUDA 13.1+ natively, so `cu13-server` runs correctly here (our composes are intentionally pinned to it). The cu13 degradation you hit looks specific to driver `580.126.09` (CUDA 13.0, no Forward Compat on consumer Ampere) — a real issue on *your* setup, but not a club-3090 default we should change.

So please **don't** spend time resubmitting the cu13→cu12 swap as a repo PR — we're staying on cu13. For your own rig, an `IK_LLAMA_IMAGE=ghcr.io/ikawrakow/ik-llama-cpp:cu12-server` override in your `.env` is the clean fix and needs nothing from us.

The other slices still stand if you want them: the `repeat-penalty` change as its own A/B PR, and the research benchmark rows as a docs-only PR. Thanks again — and sorry for the whipsaw on #1.

PR #305, opened 2026-06-03
fix(ik-llama): switch byteshape-iq4xs to cu12-server (cu13 silent CPU fallback on RTX 3090)

Rhonstin:
## Problem

`cu13-server` (CUDA 13.1.1) silently falls back to CPU on RTX 3090 (sm_86, consumer Ampere).

**Rig:** driver 580.126.09, CUDA runtime 13.0, RTX 3090 24 GB (sm_86)

**Symptom:** weights load to VRAM correctly, but all compute ops run on CPU. No kernel error, no warning — the only signal is decode throughput dropping to ~2 tok/s (CPU speed) instead of the expected ~115 tok/s. Easy to miss if you don't watch actual token rate.

**Root cause:** RTX 3090 consumer lacks CUDA Forward Compatibility support. `cu13-server` requires Forward Compat to run its CUDA 13.1.1 runtime on a driver reporting CUDA 13.0 — the consumer variant simply doesn't have that path, so it silently degrades.

## Fix

Switch `byteshape-iq4xs/mtp.yml` from `cu13-server` to `cu12-server` (CUDA 12.6.2). This is the last remaining `cu13-server` reference in a shipped ik-llama compose — all other ik-llama composes were already switched on 2026-05-28.

Also updates the `test-preflight-compose-deps.sh` test fixture to match.

## Validation

- `cu12-server` restores full GPU decode on this rig (~115 tok/s, matches BENCHMARKS row)
- `verify-full` 8/8 pass on byteshape IQ4_XS compose with cu12-server
- All other shipped ik-llama composes already run cu12-server with no issues

## Scope

2 files changed:
- `models/qwen3.6-35b-a3b/ik-llama/compose/single/byteshape-iq4xs/mtp.yml` — image + comment
- `scripts/tests/test-preflight-compose-deps.sh` — test fixture inline compose

### DIFF
--- a/models/qwen3.6-35b-a3b/ik-llama/compose/single/byteshape-iq4xs/mtp.yml
+++ b/models/qwen3.6-35b-a3b/ik-llama/compose/single/byteshape-iq4xs/mtp.yml
@@ -1,7 +1,7 @@
 # ===========================================================================
 # Profile (at-a-glance):
 #   Model:     Qwen3.6-35B-A3B (byteshape IQ4_XS 4.19bpw GGUF)
-#   Engine:    ik_llama.cpp (ghcr.io/ikawrakow/ik-llama-cpp:cu13-server)
+#   Engine:    ik_llama.cpp (ghcr.io/ikawrakow/ik-llama-cpp:cu12-server)
 #   Topology:  Single 3090 (TP=1)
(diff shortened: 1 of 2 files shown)


Rhonstin:
Closing — cu13 is correct for the repo default (driver 610.43.02+). The cu12 degradation is specific to driver 580.126.09 / RTX 3090 consumer (no Forward Compat). Will use IK_LLAMA_IMAGE override in local .env instead.

PR #292, opened 2026-06-01
[DRAFT] #254: reconcile vLLM engine/registry/composes to v0.22.0 (plan)

noonghunna:
**Draft / planning PR — not for merge yet.** Captures the plan to complete #254 (vLLM engine ↔ image reconciliation to v0.22.0) after today's image prune, so we can revisit + execute later.

Today we pruned local docker images to **only `vllm/vllm-openai:v0.22.0`** (deleted `v0.21.0`/`:latest` + the purged nightlies `bf610c2f`/`01d4d1ad`/`e47c98ef`). That fixed the *image inventory* but the registry/engine layer still references the deleted images — which is exactly #254, now non-optional.

**Full plan is in the committed doc:** [`docs/254-vllm-v0220-reconciliation-plan.md`](../blob/chore/254-vllm-v0220-reconciliation/docs/254-vllm-v0220-reconciliation-plan.md). TL;DR (4 layers, dependency order):

1. **Engine** — add `vllm-stable-v0220` (plain v0.22.0, **no overlays**; decision: don't reuse `vllm-gemma-stable`, it carries the gemma #40391 overlay qwen mustn't mount); deprecate the 4 dead nightly engines.
2. **Registry** — repoint `vllm/dual`, `vllm/minimal`, `vllm/qwen-35b-a3b-dual` → `vllm-stable-v0220`.
3. **Prod composes** — `${VLLM_IMAGE:-…v0.22.0}` → `${VLLM_IMAGE:?…}` (engine profile = sole version source; kills the silently-load-bearing literal).
4. **Scripts** — update engine↔image fixtures (`test-launch-compat`, parity, profiles-compat); `scripts/tests/*.sh` green.

⚠️ **Gated on per-compose stock-v0.22.0 boot+behavior validation** — `qwen-35b-a3b` fp8 is already validated; `vllm/dual` (27b fp8-mtp) + `minimal` need a live tool-call/MTP check. If any needs a Genesis patch it **stays deprecated**, not repointed.

Refs #254. Currently contains only the plan doc — revisit to implement.

Issue #254, opened 2026-05-29
Drive vLLM stable image solely from the engine profile (de-dup the v0.21.0 literal; fix qwen/dual engine↔image mismatch)


noonghunna:
## Summary

Stable vLLM composes hardcode the image version as a literal fallback (`image: ${VLLM_IMAGE:-vllm/vllm-openai:v0.21.0}`), duplicating the version the engine profile's `install.spec` already owns. `resolve_variant_pin` (`scripts/lib/profiles/launch_compat.py`) already derives `VLLM_IMAGE` from `install.spec` for tagged-release engines, so for launcher-driven boots the literal is dead code — but it's a second source of the version (single-source-of-truth smell), and for qwen `vllm/dual` it's masking an engine-attribution mismatch.

Surfaced while reviewing #253 (gemma vLLM prune), which follows this same pre-existing pattern — so #253 is consistent with the repo and this is a separate cleanup.

## Part 1 — gemma stable composes (clean)

The 3 gemma survivors use `engine: vllm-gemma-stable` (`install.spec: vllm/vllm-openai:v0.21.0`), which `resolve_variant_pin` exports as `VLLM_IMAGE`. The compose fallback `${VLLM_IMAGE:-vllm/vllm-openai:v0.21.0}` duplicates that literal, so `v0.21.0` now lives in 4 places (engine profile + 3 composes). On a future bump, launcher boots update via the engine profile, but a direct `docker compose -f …` (no launcher) silently uses the stale literal.

**Fix:** make the 3 gemma composes `image: ${VLLM_IMAGE:?VLLM_IMAGE not set — launch via scripts/launch.sh or scripts/switch.sh, or export it manually}`. The engine profile becomes the sole version source; direct-compose-without-env fails with a clear message — mirroring how the nightly composes already require `${VLLM_NIGHTLY_SHA}`.

## Part 2 — qwen `vllm/dual` (deeper; engine ↔ image mismatch)

`vllm/dual` registers `engine="vllm-nightly-clean"` (`install.spec: vllm/vllm-openai:nightly-bf610c2f…`, a **purged** SHA → exports `VLLM_NIGHTLY_SHA`, not `VLLM_IMAGE`). Its compose image line is `${VLLM_IMAGE:-vllm/vllm-openai:v0.21.0}`. Because `VLLM_IMAGE` is never exported for that engine, **qwen/dual actually runs the v0.21.0 literal, not the nightly-clean SHA**. Consequences:

[...]

PR #324, opened 2026-06-05
refactor(vllm): reconcile vLLM engines to v0.22.0 — two-engine split (#254 qwen+gemma leg)

noonghunna:
## What

Collapse the per-family v0.22.0 vLLM engine into a **single versionless `vllm-stable`**
(`spec: vllm/vllm-openai:v0.22.0`). The version lives **only** in `install.spec`, so the
next stable is a one-line spec bump — the engine *name* never changes (the "keep the tag,
change the number" contract).

Key realisation: **overlays are mounted per-compose, not per-engine** (the gemma int8
compose mounts the #40391 overlay via its own `volumes` + `install.sh`; the launcher mounts
nothing). So one engine can serve both `gemma4-swa-dense` and `qwen3-next-moe` — the former
per-family split (`vllm-gemma-stable`) was never a *mounting* requirement, only metadata.

## Changes
- **Retire `vllm-gemma-stable`**; repoint `vllm/gemma-{mtp-tp1,bf16-mtp,int8-mtp}` +
  `vllm/qwen-35b-a3b-dual` → `vllm-stable`.
- **Rename** the pip dense lower-bound `vllm-stable` → **`vllm-pip-baseline`** (frees the
  clean name for the docker engine); update its 17 `arch_patches.yml` pins + the fixtures.
- `arch_patches.yml`: pin renames + add `vllm-stable@v0.22.0` to the `qwen3-next-moe` row.
- Fixtures (`test-launch-compat`, `test-generate-compose`, `test-generate-from-profile`) +
  docs (`COMPOSE_GENERATOR.md`, `UPSTREAM.md`) updated.

## Why it matters (the #254 payoff)
`vllm/qwen-35b-a3b-dual` was routed through `vllm-nightly-clean`, whose pin
(`nightly-bf610c2f`) is **purged from Docker Hub** — the launcher injected a 404'd image, so
the slug was dead via `switch.sh`/`launch.sh` despite the compose *saying* v0.22.0. After this:

```
resolve-variant-pin vllm/qwen-35b-a3b-dual  →  VLLM_IMAGE=vllm/vllm-openai:v0.22.0  ✅
resolve-variant-pin vllm/gemma-int8-mtp     →  VLLM_IMAGE=vllm/vllm-openai:v0.22.0  ✅ (unchanged)
resolve-variant-pin vllm/dual (27b)         →  VLLM_NIGHTLY_SHA=bf610c2f…           (untouched — see scope)
```

Live-validated earlier on this rig: the dual boots + serves on stock v0.22.0 (verify-full 8/8,
~172 TPS decode, coherent).

## Scope
The **qwen3-next-moe + gemma leg only.** `vllm/dual` + `vllm/minimal` (27b `qwen3-next-hybrid`)
stay on `vllm-nightly-clean` pending their own stock-v0.22.0 boot + tool-call/MTP validation
(rest of #254 — don't repoint a compose not confirmed on the engine). `vllm-nightly-clean` is
therefore kept (still load-bearing for those slugs), not deprecated here.

Supersedes the separate-engine (`vllm-stable-v0220`) approach sketched in the #292 plan —
since overlays are compose-mounted, a single unified engine is cleaner and matches the
"one stable tag, bump the number" intent.

## Tests
`for t in scripts/tests/*.sh` → **41/41 green**.

Refs #254, #292.


noonghunna:
## Codex review — no blockers ✅

Bounced this approach off Codex before merge. Verdict: **no high/critical correctness blocker.** It confirmed the load-bearing assumption the whole unify rests on:

> "`vendored_overlays` on `EngineProfile` is metadata/drift-guard only, not the mounting mechanism … the actual load-bearing path is per-compose `volumes`/`command` + `arch_patches.yml` `loads:true`. That makes the one-engine / multi-family split sound."

All review questions cleared:
- **Unified engine vs split** — sound on the current implementation path.
- **`vendored_overlays` = metadata only** — confirmed (not a mount selector).
- **`feature_provenance` on a shared engine** — no false-pass/fail; not used as a gate input.
- **Stale `@v0.21.0` arch pins** — inert: `validate_engine_pin()` only matches the live `install.spec` against `loads:true`. Safe to keep as history.

**One residual** Codex raised: the deferred `${VLLM_IMAGE:?}` hardening leaves the compose literal as an unguarded drift path *outside* the launcher (a future "bump the engine, forget the literal" would silently serve a stale image to direct `docker compose` users).

Addressed here with a **CI drift-guard** (`scripts/tests/test-compose-image-drift.sh`) rather than the `:?` UX cost — it asserts every fixed `${VLLM_IMAGE:-<tag>}` default `==` the engine's `install.spec`, so that drift now reds the gate. It immediately earned its keep: it surfaces + exempts (with a printed note) `vllm/minimal` + `vllm/dual` — the unmigrated 27b leg still on the purged nightly, tracked as the rest of #254; they drop out of the exempt set as each migrates.

Gate: **42/42 green.**

PR #359, opened 2026-06-11
dgemma: list as 🧪 experimental (discoverable in switch.sh --list)

noonghunna:
Follow-up to #358. A new model whose *only* variant is `upstream-gated` vanished entirely from `switch.sh --list` (which hides deprecated + upstream-gated), even though 22 `experimental` + 1 `preview` variants already display there with `(NA: …)` markers. There's no principled reason a parked-but-real **new model** should be less discoverable than 22 experimental configs.

Per maintainer call, list DiffusionGemma as **🧪 experimental** instead:
- Now shows in `switch.sh --list` as `vllm/diffusiongemma-dual … (NA: experimental, 262K)`.
- Launch still requires `--force` (experimental is non-functional) — unchanged.
- No `switch.sh` logic change; just the status (compose header + registry kept in sync, `test-compose-status-drift` green).

The #45163 gating / sideload rationale is still fully documented (compose prose, overlay README, `docs/UPSTREAM.md`).

Gate: status-drift, profiles-compat, diagnose-profile, model-default-resolver, patch-attribution, switch-registry-parity all green.

### DIFF
--- a/models/diffusiongemma-26b-a4b/vllm/compose/dual/fp8/base.yml
+++ b/models/diffusiongemma-26b-a4b/vllm/compose/dual/fp8/base.yml
@@ -8,25 +8,12 @@
 #   Vision:    no (text diffusion)
 #   Max ctx:   262144 (NIAH retrieval validated to ~250K)
 #   Genesis:   N/A — Genesis is Qwen3-Next-specific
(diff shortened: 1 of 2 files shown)


noonghunna:
Superseded by the :gemma adoption PR (this branch includes the ⏸️→🧪 experimental flip as part of dropping the sideload for the official image).

PR #369, opened 2026-06-11
chore: make setup script work with WEIGHTS=FP8 and qwen3.6-27b

hlo-world:
## Summary

Wasn't able to ```export WEIGHTS=fp8 && bash scripts/setup.sh qwen3.6-27b``` for ```dual-max``` because setup did not check for fp8.

This adds that check and adds target to pull the hf model from Qwen hf repo.

Just a quick fix, if owner intentionally left this workflow unreachable because fp8 is experimental, then please close without merging. Thanks.

### DIFF
--- a/scripts/lib/profiles/models/qwen3.6-27b.yml
+++ b/scripts/lib/profiles/models/qwen3.6-27b.yml
@@ -45,6 +45,7 @@ weights:
     size_gb: 29
     format: fp8
     status: experimental
+    hf_repo: Qwen/Qwen3.6-27B-FP8
     engine: vllm
     kind: main
     verify_glob: "*.safetensors"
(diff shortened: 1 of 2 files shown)


noonghunna:
Merged — thanks @hlo-world! And to answer your question directly: no, fp8 wasn't intentionally left unreachable. The `fp8` weight key existed in the profile but never had an `hf_repo` wired, so `WEIGHTS=fp8 bash scripts/setup.sh qwen3.6-27b` fell through to the "not recognized" branch. Your patch is exactly right — it points at the official `Qwen/Qwen3.6-27B-FP8` release (the same weights the `vllm/qwen-27b-dual-max` / `-multi-max` tiers already use) and routes it through `setup.sh`'s standard whole-repo download + per-file SHA-verify path.

I pushed a small follow-up (`eae0e36`) for three cosmetic bits on top of your commit:
- `ft8` → `fp8` typo in the error message
- added `fp8` to the "unrecognized WEIGHTS" help string
- refreshed the entry's `manual_note` (it still said "no direct pull recipe wired"; now documents the `WEIGHTS=fp8` path, keeping `hf-download.sh` as the stall-resistant manual option for the 29 GB layer-split repo)

Guards (`test-model-weights-registry` / `test-setup-picker` / `test-profiles-compat`) stay green. Appreciate the clean, well-scoped fix 🙏

PR #373, opened 2026-06-11
Add DiffusionGemma-26B-A4B llama.cpp single-card serving (PR #24423)

ashalliants:
## What

Adds a **single-3090** serving path for **DiffusionGemma-26B-A4B** (Google's block-diffusion dLLM) via **llama.cpp / GGUF**, behind an OpenAI-compatible endpoint. This **complements** the team's vLLM/fp8 **dual-card** path already on master (`vllm/diffusiongemma-dual`) — different engine subdir (`llama-cpp/` vs `vllm/`), no file overlap. The model `README.md` now presents both routes side by side (pick by card count).

Marked **⏸️ Upstream-gated** — not promoted to the catalog/DEFAULTS.

## Why a separate route

DiffusionGemma is a true diffusion LM. On a **single** 3090, the fp8 weights the vLLM path needs (~25 GB + fp8 Marlin) don't fit, and FP8/NVFP4 need Hopper/Blackwell. The only single-card option is **GGUF-Q4 via llama.cpp draft [PR #24423](https://github.com/ggml-org/llama.cpp/pull/24423)** — which ships an interactive CLI + a logits microservice but **no HTTP server**.

## How

```
harness ──/v1──▶ FastAPI shim ──NDJSON──▶ patched llama-diffusion-cli (DG_NDJSON=1, resident, KV-cache on)
```

- `patches/diffusion-ndjson-server.patch` — small **env-gated** `DG_NDJSON` stdin/stdout mode added to `llama-diffusion-cli` (reuses the PR's `apply_template` + `run_turn`; no new model logic; thinking-off default for compact answers).
- `serve/diffusion_openai_server.py` — FastAPI shim: `/v1/chat/completions` + `/v1/models`, strips the `<|channel>thought…<channel|>` / `<turn|>` markers, single in-flight (np=1), **HTTP 413** on prompt-overflow.
- `Dockerfile` — **primary delivery**: builds the patched CLI **from the existing draft-PR branch** (pinned sha via build-arg) and bundles the shim into a custom image (no stock image carries `llama-diffusion-cli`). `serve.sh` is the no-Docker host fallback.
- `compose/single/q4-k-m/openai-shim.yml` — builds + runs that image.
- `docs/UPSTREAM.md` — adds the llama.cpp #24423 row (the vLLM #45163 row is upstream's, untouched).

## Measured (1× 3090, Q4_K_M, `-ngl 99`, host path, 2026-06-11)

| Metric | Value |
|---|---|
| Effective throughput (thinking-off, coding) | **~75 tok/s** (5–6 steps, ~1–2 s / 256-tok block) |
| Peak VRAM | **22.3–23.0 GB / 24** (tight) |
| Prompt ceiling @ `DG_CTX=3072` | ~2800 tok → HTTP 413 beyond |

**Speed verdict:** on a single 3090 diffusion is **not** a speed win vs our autoregressive MoEs — the FP8 speedups need Hopper (or the dual-card vLLM route, which is faster + goes to 262K). Value here is capability/experimentation on one card.

## Known limitations

- Draft-PR dependency (unmerged, AI-assisted/logit-verified); image pins the branch sha.
- VRAM knife's-edge — raising `DG_CTX` risks OOM.
- Trivial ultra-short prompts can return empty (entropy-bound sampler quirk); real coding prompts are fine.
- v1 cutouts: no native tool-call JSON (content only); "streaming" = one SSE chunk.

## Testing

Validated on 1× 3090 via the host path: builds clean (SM 86), NDJSON round-trip, `/v1/models` + `/v1/chat/completions` + streaming, multi-turn + system, clean output (no marker leak), correct code (quicksort/is_prime/fib/merge-sort/SQL), 413 guard, VRAM as above. The Docker image path mirrors the host path exactly but was **not** built in the authoring environment (no Docker daemon in that WSL distro) — noted in the compose caveats.

Design rationale: `docs/superpowers/specs/2026-06-11-diffusiongemma-serving-design.md`.


tomByrer:
Thanks for this; I was wondering how to best test this on my rig!

I wonder if a PR should _not_ be accepted day 0; might be a few days for better quants since we're in the tight-VRAM zone.

That said, I do love the idea of having a PR to start off testing & honing on day 0.

noonghunna:
Thanks @ashalliants — careful work: the NDJSON shim design, the honest "diffusion isn't a single-card speed win" verdict, and the thorough host-path validation are all appreciated.

I'm going to hold off on incorporating this for now — and to be straight, it's a maintenance/timing call, not a quality one. Two reasons:

1. **Supporting it well needs validation time I can't commit right now.** It's a multi-part custom delivery (FastAPI shim + a CLI patch + a Dockerfile that builds from a branch), behind an experimental capability that — by your own measurement — isn't a speed win on one card. The test+maintenance surface doesn't pencil out for us at the moment.
2. **It's a patch on an unmerged, moving draft.** #24423 is still draft and actively churning — it changed just today (it now ships a purpose-built stdin generation server, which arguably supersedes the `llama-diffusion-cli` patch this PR wraps). Carrying vendored patches against a moving upstream has real overhead for us — it's why our engine-pinning policy keeps patches minimal, and we keep relearning it. We'd rather track a stable upstream surface than maintain a shim against a branch that re-bases under us.

The cleaner path from our side: **when #24423 lands in mainline llama.cpp** — ideally with native HTTP serving, or at least a stable server target to shim — we'd happily revisit. At that point it's tracking upstream, not maintaining a patch on a draft.

For what it's worth, the dual-card vLLM route (`vllm/diffusiongemma-dual`) is already on master, so two-card users aren't blocked — this is specifically the single-card-over-HTTP gap, and the honest move is to let upstream close it rather than carry a shim in the meantime.

Genuinely good work, and the design doc is a solid reference — please keep them coming.

Issue #390, opened 2026-06-12
[bench] qwen3.6-35b-a3b on mainline llama.cpp, single 3090 — 131k ctx at 15.1GB peak


deucebucket:
### Rig + bench report — paste contents of `bash scripts/report.sh --full` (or `--bench` if you couldn't run the full chain)

# club-3090 rig report

Generated: 2026-06-12 16:18:18 UTC

_Redacted output (paths, host, user, tokens). Re-run with `--no-redact` for full data._

## System

- **OS:** Bazzite
- **Kernel:** 6.19.14-ogc5.1.fc44.x86_64
- **Environment:** bare metal
- **Locale:** en_US.UTF-8
- **Timezone:** CDT
- **Uptime:** up 4 days, 15 hours, 25 minutes

## CPU + RAM

- **CPU:** AMD Ryzen 7 5800XT 8-Core Processor (16 threads)
- **RAM:** 62Gi total, 49Gi available
- **Swap:** 15Gi

## GPU hardware

- **GPU 0:** NVIDIA GeForce RTX 3090 | 24576 MiB | driver 580.142 | VBIOS 94.02.26.48.16 | persistence=Disabled
  - **Power:** limit=420.00 W (default=420.00 W, max=450.00 W) | current_draw=144.32 W
  - **PCIe:** x16 lanes negotiated (GPU max x16, Gen up to 4) | bus 00000000:07:00.0
- **CUDA Runtime (per driver):** 13.0
- **ECC mode:** [N/A] (3090s don't have ECC; expect N/A)

### NVLink

_No NVLink detected (PCIe-only)_

### Topology

<details><summary>PCIe / GPU topology matrix</summary>

```
	[4mGPU0	CPU Affinity	NUMA Affinity	GPU NUMA ID[0m
GPU0	 X 	0-15	0		N/A

Legend:

  X    = Self
  SYS  = Connection traversing PCIe as well as the SMP interconnect between NUMA nodes (e.g., QPI/UPI)
  NODE = Connection traversing PCIe as well as the interconnect between PCIe Host Bridges within a NUMA node
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe bridges (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing at most a single PCIe bridge
  NV#  = Connection traversing a bonded set of # NVLinks
```

</details>

### PCIe / P2P detail (lspci)

[...]

PR #293, opened 2026-06-01
feat(35b-a3b): add byteshape IQ4_XS single-card compose + benchmarks

Rhonstin:
## Summary

- New compose `models/qwen3.6-35b-a3b/ik-llama/compose/single/byteshape-iq4xs/mtp.yml` for [byteshape/Qwen3.6-35B-A3B-MTP-GGUF](https://huggingface.co/byteshape/Qwen3.6-35B-A3B-MTP-GGUF) IQ4_XS 4.19bpw (~18 GB, MTP head embedded)
- Adds `tools/build-sandboxes.sh` — wrapper that clones benchlocal-cli and builds hermesagent-20 / bugfind-15 / cli-40 sandbox Docker images
- BENCHMARKS.md row with full 8-pack quality results

## Validation (1× RTX 3090, 2026-06-01)

- `verify-full` **8/8**
- `verify-stress` **8/8** — NIAH→240K/262K, 25K tool-prefill OK, no Cliff 2b (MoE KV self size 144 MiB at 262K ctx)
- **8-pack quality 111/150 (74%)**: toolcall 15/15 · instructfollow 15/15 · structoutput 14/15 · dataextract 13/15 · reasonmath 12/15 · bugfind 13/15 · hermesagent-20 12/20 · cli-40 17/40
- Highest overall 8-pack on this rig (+4 pp vs beellama DFlash 107/150=71%)
- VRAM: 21.8 GB, ctx: --fit auto-selected 262K (full n_ctx_train)
- bench.sh + soak-continuous pending → Status: `⚠️ Production w/ caveats`

## Notes

- hermesagent-20 60% (12/20) — beellama DFlash remains hermes-agent leader (14/20, 70%)
- dataextract significantly stronger than beellama: 13/15 (87%) vs 8/15 (53%)
- MoE KV is cheap: no Cliff 2b risk even at 262K (compare dense 27B which OOMs at ~25K tool-prefill on vLLM)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #393, opened 2026-06-13
feat(35b-a3b): add deucebucket-cerebellum single-card compose (mainline llama.cpp)

deucebucket:
## Summary

the compose invited in #390. mainline llama.cpp single-card lane for qwen3.6-35b-a3b, running the cerebellum v3 mixed-precision GGUF (11.96 GB, ablation-informed per-tensor precision, [deucebucket/Qwen3.6-35B-A3B-Cerebellum-GGUF](https://huggingface.co/deucebucket/Qwen3.6-35B-A3B-Cerebellum-GGUF)). fills the empty llama-cpp/ slot under this model: ~15.1 GB peak at 131K q8/q8 KV, ~9 GB free on a 24 GB card, the only 35b-a3b single-card config with a 16 GB story.

one new file, `models/qwen3.6-35b-a3b/llama-cpp/compose/single/deucebucket-cerebellum/base.yml`, mirrored structurally from `models/qwen3.6-27b/llama-cpp/compose/single/unsloth-q4km/mtp.yml`. both flag corrections from the issue are in: `-np 1` (the -np 4 ctx split into 4x32768 is what 400'd my bench-agentic at turn 12) and `--jinja --reasoning off --reasoning-format deepseek` plus top-p 0.95. image stays at the b9246 pin default per stack policy. exact GGUF filename is in the quick-start comment for SHA verification. status 🧪 experimental, header has the sibling table vs apex-fit-q8q5 and byteshape-iq4xs.

## measured (my rig, 2026-06-12, full report.sh --full chain on #390)

- 1x 3090 at stock 420 W limit, bare metal b9603, desktop compositor on the same gpu
- bench n=5: 147.9 narr / 146.2 code wall TPS, CV under 1%. caveat: the server was at -np 4 for those runs; this compose ships -np 1, so treat them as provisional until your validation produces the -np 1 numbers
- peak vram 15,115 MiB at 131K. 196K boots at 15,869 MiB and fill-probed to ~119K, but no NIAH ladder at depth, so the compose defaults to 131K and the header flags 196K as boot+fill only
- verify-full PASS, soak-continuous PASS (0 errors, 7 MiB growth, 100% retention)
- deterministic 5-pack 63/75: toolcall 13, instructfollow 15, structoutput 14, dataextract 9, reasonmath 12. dataextract is the soft spot (the family-wide JSON numeric-typing thing)

## left for your side, per #390

- port (XXXXX placeholder in the compose)
- registry slug wiring (no scripts/lib/profiles/ touched). heads up: `test-compose-registry-disk` (47-file count + disk-only allowlist) and `test-switch-registry-parity` fail until the registry entry lands; everything else in the suite is green vs master baseline (only the known test-submit-bench fixture failure)
- sandboxed packs (bugfind / hermesagent-20 / cli-40) to complete the /150 and set eventual status
- b9246 -> b9603 pin co-validation

no BENCHMARKS.md row and no models/qwen3.6-35b-a3b/CHANGELOG.md entry in this PR: my TPS numbers are -np 4 so a row would misstate the shipped config, and that model dir has no changelog yet (the byteshape intake #293 didn't add one either). happy to add either if you want them here.

related: #390

### DIFF
--- a/models/qwen3.6-35b-a3b/llama-cpp/compose/single/deucebucket-cerebellum/base.yml
+++ b/models/qwen3.6-35b-a3b/llama-cpp/compose/single/deucebucket-cerebellum/base.yml
@@ -1,0 +1,159 @@
+# ===========================================================================
+# Profile (at-a-glance):
+#   Model:     Qwen3.6-35B-A3B (deucebucket Cerebellum v3 mixed-precision GGUF,
+#              Q3_K_M-class, 11.96 GB — ablation-informed per-tensor precision map)
+#   Engine:    llama.cpp (mainline ggml-org server-cuda)
+#   Topology:  Single 3090 (TP=1)
+#   Drafter:   none — no MTP head in this GGUF; mainline, no spec-dec
+#   KV:        q8_0 K + q8_0 V
+#   Vision:    no (mmproj not mounted)
+#   Template:  native (GGUF-embedded) via --jinja
+#   Max ctx:   131072 default (~15.1 GB peak — ~9 GB free on a 24 GB card);
+#              196608 boots at ~15.9 GB peak and fill-probed to ~119K prompt
+#              tokens, but the NIAH depth ladder has NOT been run there yet —
+#              treat 196K as boot+fill, not a verified ceiling (boots ≠ fills).
+#   Genesis:   N/A — llama.cpp engine; Genesis is vLLM/Qwen3-Next-specific
+#   Status:    🧪 Experimental — author-rig validated (verify-full PASS,
+#              soak-continuous PASS, bench n=5 CV<1%); club-side validation,
+#              sandboxed packs, port + registry slug pending (#390)
+#   Engine-profile: llama-cpp-local
+#   Quality:   toolcall-15 13/15 (87%) · instructfollow-15 15/15 (100%) ·
+#              structoutput-15 14/15 (93%) · dataextract-15 9/15 (60%) ·
+#              reasonmath-15 12/15 (80%) — 63/75 on the 5 deterministic packs
+#              (benchlocal --medium, thinking=pack-defaults, author 1× 3090,
+#              2026-06-12). Sandboxed packs (bugfind / hermesagent-20 / cli-40)
+#              not run — no docker on the bench host; /150 to be completed
+#              club-side per #390.
+#   Best for:  lean-VRAM single card — a coherent 35B-A3B in ~15 GB on stock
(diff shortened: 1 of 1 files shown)


noonghunna:
@deucebucket first — this is genuinely one of the cleanest community composes we've gotten. You read the template closely: the `-np 1` correction with the slot-math comment, the honest "boots ≠ fills" framing on the 196 K ceiling, the per-tensor ablation quant writeup. Thank you for the care.

After thinking it through, I'm going to land this as a **credited benchmarks data point rather than a catalog compose** — and I want to be straight about why, because it's about our rig, not your work.

The compose's headline value is the **lean-VRAM / 16 GB fit** — a real differentiator. But we don't have a 16 GB card in the rig. A curated compose carries an implicit "we validated this and we'll help you troubleshoot it," and the failure modes that actually bite on 16 GB — OOM-at-fill (boots fine, dies filling 131 K), the exact boots ≠ fills boundary you flagged at 196 K, card-specific driver/clock quirks — are the ones we can't reproduce, let alone fix. Listing it as *the* 16 GB path would be a support promise we can't keep, which is worse for a 16 GB user than us being honest that we don't run that tier.

And on 24 GB (what we *can* stand behind), it's covered: the two ik-llama single-card paths give more context (196 K / full 262 K) + higher quality + MTP, with this trading those for narrative speed + headroom — a real niche, but not one that needs a third catalog slot to maintain.

So, to make sure the work lands and gets credit:
- The **README `llama.cpp ❌ → ✅`** correction your report drove is already in (mainline llama.cpp *does* serve this MoE — your run is the proof).
- A **benchmarks row** crediting your numbers (marked author-rig), so the data's on the matrix.
- Linking your [HF repo](https://huggingface.co/deucebucket/Qwen3.6-35B-A3B-Cerebellum-GGUF) + this PR so 16 GB owners — who this genuinely helps — can find it directly.

The ablation-informed per-tensor quant is a neat approach and the lean footprint is a real achievement — this isn't a quality call, it's a "what hardware can we honestly support" one. I'll close the PR with this, but the credit above is real and lasting. If you build things that land on 24 GB — or want to co-author the 16 GB validation story once a few owners report numbers — genuinely keep them coming. Thanks again.

deucebucket:
all fair, and the row plus links is appreciated — the 16gb support-promise logic makes sense from where you sit. i'll put a call for numbers on the HF repo so 16gb owners can start building that validation story.

ivanbaldo:
Maybe it can be tested with Google Collab with the free T4?
Although no Docker there but still could be done manually running the commands I guess...

Issue #320, opened 2026-06-05
[test] weights-registry guard should verify hf_repo resolves on HF, not just string-match


noonghunna:
## Motivation

In #316 the `qwen3.6-35b-a3b:autoround-int4` profile pointed at a placeholder repo id (`Qwen/Qwen3-MoE-A3B-Instruct-AutoRound-Int4-mixed`) that **never existed on HF**. The guard test `scripts/tests/test-model-weights-registry.sh` asserted the **same wrong string**, so it *codified* the typo rather than catching it.

It stayed hidden because on the maintainer rig the weights were already on disk (never re-fetched), so the 404 only surfaced when a fresh user ran `setup.sh` and `hf download` 404'd (#316).

## Problem

The guard does a pure string round-trip (`assert_entry <key> <repo> <subdir>`). A non-existent or renamed `hf_repo` passes the gate as long as the string matches the profile — exactly the failure mode that wastes a new user's first run.

## Proposal

Add an **opt-in, network-gated** check that each `hf_repo` in the registry resolves on HF (HTTP 200 via the HF API), so a bad/renamed repo reds the check.

- **Network-gated** behind an env flag (e.g. `CLUB3090_CHECK_HF_REPOS=1`) or a separate target — the offline string-assert test keeps running everywhere (CI may be offline / HF rate-limited).
- Consider running it **periodically / manually** rather than every commit, to avoid HF rate-limit flakiness.
- Report the repo id + HTTP status on failure.

This also catches the *upstream-rename* case (old name → 404) early, which is the same root cause as the re-quant in #316.

## Acceptance

- With the flag set, a bad or renamed `hf_repo` fails with `<repo id>` + HTTP status.
- With the flag unset, behavior is unchanged (offline-safe).

Refs #316. Complements the `revision`-pin enhancement (sibling issue).

noonghunna:
[...]

Issue #319, opened 2026-06-05
[enhancement] Support a revision/commit pin in the weights-fetch schema (weights.py)


noonghunna:
## Motivation

#316 surfaced that HF quant repos mutate **silently**. `Intel/Qwen3.6-35B-A3B-int4-mixed-AutoRound` was renamed (from `…-int4-AutoRound`) and **re-quantized** on the upstream side after we benchmarked it — so a fresh `setup.sh` pulls a *different revision* than the one our published numbers were measured against (9 of 11 safetensors shards differ in size vs the revision we validated; the old commit is no longer in the repo's history).

This is the same class of problem as upstream rolling-tag churn that we already pin around for engine images — but for **weights** we have no equivalent lever.

## Problem

Weights entries in `scripts/lib/profiles/models/<model>.yml` support `hf_repo` / `hf_repos`, but there is **no `revision` field**:

- `scripts/lib/profiles/weights.py` doesn't parse one (no `WEIGHT_REVISION` is emitted).
- The fetch path (`setup.sh` / preflight fetch-now) passes no `--revision` to `hf download`.

So we cannot pin the exact commit our BENCHMARKS row corresponds to — users silently get whatever `main` HEAD is at download time, which can be a re-quant we never validated.

## Proposal

- Add an **optional** `revision:` key (commit SHA or tag) per weights variant in the profile schema.
- Parse it in `weights.py` → emit `WEIGHT_REVISION`.
- Thread it into the download invocation: `hf download <repo> --revision <sha> …`.
- Keep it **optional** — unset = track HEAD (today's behavior). Set = reproducible against exact bytes.

## Acceptance

- A profile with `revision:` pins the fetch to that commit.
- A guard test asserts the field round-trips through `weights.py`.

Refs #316.

tomByrer:
Perhaps the better answer is just to copy/paste the weights into your/our own Hugging Face repo, so both the URL & model don't stealth-change? (if the license allows)

Folks do this all this all the time mirroring CivAI for image models & LORAs on HF.

noonghunna:
[...]

Issue #316, opened 2026-06-05
[bug] Error: Model 'Qwen/Qwen3-MoE-A3B-Instruct-AutoRound-Int4-mixed' not found.


bluehawksai:
### What happened?

bash scripts/setup.sh qwen3.6-35b-a3b
[model]   qwen3.6-35b-a3b:autoround-int4 -> Qwen/Qwen3-MoE-A3B-Instruct-AutoRound-Int4-mixed  -> qwen3.6-35b-a3b-autoround-int4
[preflight] checking environment...
[preflight] docker:  29.5.1 (compose v2 ok)
[preflight] gpu:     3× detected
[preflight]            GPU 0: NVIDIA GeForce RTX 3090 (UUID: GPU-80de4402-3351-9b49-1f19-fa130427fdeb)
[preflight]            GPU 1: NVIDIA GeForce RTX 3090 (UUID: GPU-567e82fc-2970-3a1d-f177-f155bb906cc1)
[preflight]            GPU 2: NVIDIA GeForce RTX 3090 (UUID: GPU-ec5e5e47-08ca-c75f-0651-adaef02ca80c)
[preflight] disk:    1276 GB free at /home/chandanv/models (need ~25 GB)
[preflight] WARNING: HF_TOKEN is not set in the environment.
[preflight]          Qwen3.6-27B is T&C-gated on HuggingFace; downloads will fail without a token.
[preflight]          Fix: visit https://huggingface.co/settings/tokens, create a read token,
[preflight]               accept the model T&C at https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct
[preflight]               (and any other Qwen3-Next variant you'll use),
[preflight]               then export HF_TOKEN=hf_... in your shell or .env file.
[preflight] ok.

Setup root:   /home/chandanv/club-3090
Model dir:    /home/chandanv/models
[genesis] qwen3.6-35b-a3b doesn't use Genesis — skipping clone.
[model]   Using 'hf download' (hf_transfer if available) ...
/home/chandanv/.local/lib/python3.10/site-packages/huggingface_hub/constants.py:277: FutureWarning: The `HF_HUB_ENABLE_HF_TRANSFER` environment variable is deprecated as 'hf_transfer' is not used anymore. Please use `HF_XET_HIGH_PERFORMANCE` instead to enable high performance transfer with Xet. Visit https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hfxethighperformance for more details.
[...]

PR #408, opened 2026-06-14
feat(weights): optional revision: pin in weights-fetch schema (#319)

deucebucket:
## summary

adds an optional `revision:` key per weights variant so a fetch can be pinned to an exact commit or tag, the same way we already pin engine images. unset keeps today's behavior (track HEAD), so nothing changes for any current entry.

the chain:
- `weights.py` emits `WEIGHT_REVISION` from the variant's optional `revision:` (empty when unset)
- `setup.sh` `_hf_download_repo` adds `--revision <sha>` only when one is set
- `setup.sh` `_verify_downloaded_files` pins the post-download sha-verify etag lookup to that same revision. without this, a pin to an older commit would etag-check against current HEAD and false-fail
- `preflight.sh` manual-fetch hint mirrors the flag so the copy-paste fallback grabs the same bytes
- `docs/ADDING_MODELS.md` documents the field
- new guard `scripts/tests/test-weights-revision.sh` asserts the round-trip (set -> `WEIGHT_REVISION`, unset -> empty)

## what problem this solves

#316: upstream quant repos re-quant silently. `Intel/Qwen3.6-35B-A3B-int4-mixed-AutoRound` was renamed and re-quantized after we benched it, so a fresh `setup.sh` pulls different bytes than our published numbers were measured against. engine images already have a pin lever, weights had none. this adds it.

## measured impact

no perf impact, it's a fetch-schema lever. the impact is reproducibility: a pinned variant fetches exact bytes, and the sha-verify checks against that same revision instead of HEAD.

## compared against

current behavior (no pin). confirmed at HEAD that `weights.py` emitted no `WEIGHT_REVISION` and the fetch path passed no `--revision`. on this branch, unset behaves identically (no flag, `/resolve/main/`); set adds `--revision` and switches verify to `/resolve/<sha>/`.

## trade-off

none for existing configs. the field is optional and no entry sets it in this PR.

## verification

this doesn't change what the server does, only what `setup.sh` downloads, and only when a `revision:` is present (none are yet), so it isn't a verify-full change. checked three ways:

- round-trip: the new guard passes; real entries emit empty (not absent); sha and tag values pass through verbatim
- no regression: full `scripts/tests/*.sh` sweep green except the pre-existing `test-submit-bench` fixture failure, which fails identically on master and touches none of these files
- command construction: stubbed `hf` / `curl`, ran the real `_hf_download_repo` and `_verify_downloaded_files`. unset -> `hf download <repo> --local-dir ...` + `/resolve/main/`; pinned -> `hf download <repo> --revision <sha> ...` + `/resolve/<sha>/`. `bash -n` clean on both scripts.

## left to your side

mechanism only. i deliberately didn't pin any real entry: which variant to pin, and to which validated sha, is a per-entry rig call that's yours to make. the #316 qwen moe's good revision is gone from the repo history anyway, so that one can't be pinned retroactively. once #316 re-validation lands, wiring `revision: <sha>` into an entry is a one-line follow-up.

happy to also pick up the #320 sibling (the hf-repo-resolve guard) so the two halves of #316 land together.

refs #319, #316

### DIFF
--- a/docs/ADDING_MODELS.md
+++ b/docs/ADDING_MODELS.md
@@ -195,6 +195,10 @@ weights:
     format: autoround                      # autoround | awq | gguf | …
     status: production                     # production | experimental | community-experimental
     hf_repo: <Org/Repo>
+    revision: <sha-or-tag>                 # OPTIONAL (#319): pin the fetch to an exact
+                                           #   commit/tag. Unset = track HEAD (default).
+                                           #   Set it to reproduce the bytes a BENCHMARKS
+                                           #   row was measured against (guards re-quants).
     files: ["*.safetensors"]               # or explicit GGUF filenames
     engine: vllm                           # vllm | ik-llama | llama-cpp | beellama (filesystem dir name)
     kind: main                             # main | draft | mmproj | gguf
--- a/scripts/lib/profiles/weights.py
+++ b/scripts/lib/profiles/weights.py
@@ -118,6 +118,7 @@ def _recipe(model_id: str, variant: str) -> dict[str, str]:
         "WEIGHT_ENGINE": str(meta.get("engine") or ""),
         "WEIGHT_KIND": str(meta.get("kind") or ""),
         "WEIGHT_REPO": str(meta.get("hf_repo") or ""),
+        "WEIGHT_REVISION": str(meta.get("revision") or ""),
         "WEIGHT_SUBDIR": str(meta.get("local_subdir") or meta.get("path") or ""),
         "WEIGHT_FILES": " ".join(str(f) for f in files),
         "WEIGHT_VERIFY_GLOB": str(meta.get("verify_glob") or "*.safetensors"),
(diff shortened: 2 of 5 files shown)


noonghunna:
Merged — thank you! 🙏 Reviewed the full chain and ran it: new guard passes, `bash -n` clean on both scripts, full `scripts/tests/*.sh` sweep green (no regression), and `WEIGHT_REVISION` round-trips (set→sha, unset→empty). Particularly appreciated that you pinned the **verify** etag lookup to the same revision (`/resolve/<sha>/`) — without that an older pin would SHA-check against HEAD and false-fail; easy to miss, and you got it right. Symmetric with the engine-image pin and zero blast radius (optional, no entry sets it yet). Natural first user: I'll pin a `revision:` on the new `hauhaucs-35ba3b-dual` entry to lock the bytes its BENCHMARKS row was measured against. Yes please on the redirect-canonicalization cleanup too.

Issue #133, opened 2026-05-14
investigate KV-cache reuse: A/B vLLM prefix-caching + evaluate LMCache integration


noonghunna:
## Context

Got pointed to [LMCache](https://lmcache.ai/) and its [KV cache calculator](https://lmcache.ai/kv_cache_calculator.html). LMCache adds persistent prefix caching, tiered storage (GPU → CPU → disk → object store), and cross-instance KV sharing on top of vLLM/SGLang. Headline claim is "7× TTFT reduction" for cache-hit scenarios with repeated prefixes.

Tracking two threads in one issue: (1) measuring whether built-in `--enable-prefix-caching` is already firing on our composes, and (2) evaluating LMCache integration. After community pushback (see thread below) the LMCache half is now a real candidate rather than parked — the math + the multi-model homelab use case both favor it more than my initial analysis suggested.

## Part 1 — A/B vLLM prefix-caching firing rate (actionable, gating)

For each canonical compose, verify whether prefix-caching is active + measure the TTFT delta on repeated-prefix workloads (IDE-agent flows with repeated system prompt + tool defs are the prime case).

### Investigation steps

1. **Confirm prefix-caching state**:
   - `docker logs <container> | grep -i 'prefix.cache\|enable_prefix_caching'`
   - vLLM v1 enables by default, but Genesis-loaded paths + nightlies worth confirming
2. **Measure cache hit rate** via vLLM internal logs (look for `Prefix cache hit rate` or equivalent metric line)
3. **Run repeated-prefix bench**:
   - Prompt 1: long system prompt (~2K tokens) + question A
   - Prompt 2: same system prompt + question B
   - Measure TTFT delta — cache hit should drop prompt 2 TTFT significantly
4. **Document per-compose**: `vllm/dual`, `vllm/dual-turbo`, `vllm/dual-dflash`, `vllm/tools-text`, `vllm/bounded-thinking` — particularly the agent-style configs

### Acceptance

- [ ] Confirmed prefix-caching status (on / off / under-firing) per canonical compose
- [ ] Measured TTFT delta on repeated-prefix workload for at least `vllm/dual` + `vllm/dual-turbo`
- [ ] Findings documented in `BENCHMARKS.md` or `LEARNINGS.md`
- [ ] If under-firing on any compose: filed as a separate bug

## Part 2 — LMCache integration (candidate, evaluate after Part 1)

Revised analysis after community feedback in the comments:

| Capability | Relevance to club-3090 | Notes |
[...]

PR #433, opened 2026-06-18
feat(lmcache): NVLink auto-detect + Compose v5 entrypoint fix

steamEngineer:
## Summary

- Port **NVLink auto-detection** from `vllm/qwen-27b-dual-max` (`dual/fp8/mtp.yml`) onto the incubating LMCache compose (`vllm/qwen-27b-dual-lmcache`)
- On NVLink rigs: enable NCCL P2P (`NCCL_P2P_LEVEL=NVL`) and **omit** `--disable-custom-all-reduce` so TP=2 all-reduce uses the GPU–GPU NVLink path instead of PCIe
- On PCIe rigs: unchanged behavior (`NCCL_P2P_DISABLE=1`, `--disable-custom-all-reduce`)

**Includes #429** (Compose v5 `$$` entrypoint escape — required for this compose to parse on Docker Compose v5.1+). If #429 lands first, this PR can be rebased to a single-commit NVLink-only diff.

## Why this was needed

The LMCache compose shipped as **PCIe-forced** even though every other dual vLLM compose in this repo auto-detects NVLink at boot via `scripts/detect_nvlink.sh` + `NVLINK_MODE` (see `docs/DUAL_CARD.md`).

Observed on an NVLink bridge rig running LMCache stress tests:

- High sustained **PCIe TX/RX** on both 3090s
- **Flat NVLink counters** during TP traffic
- Container env showed `NCCL_P2P_DISABLE=1` and entrypoint always passed `--disable-custom-all-reduce`

Root cause: `lmcache.yml` omitted the `detect_nvlink.sh` mount, `NVLINK_MODE` env, and conditional entrypoint that `dual/fp8/mtp.yml` uses. That was likely an oversight when authoring the third-party `lmcache/vllm-openai` image path (#133), not an LMCache requirement — the controlled A/B in #133 toggled only the connector, not NCCL topology.

**Note:** LMCache KV offload (GPU ↔ host DRAM) still uses PCIe per GPU; NVLink does not remove that. This change fixes **tensor-parallel GPU↔GPU** traffic only.

## Implementation notes

| Change | Purpose |
|--------|---------|
| Mount `scripts/detect_nvlink.sh` | Same boot-time topology probe as dual-max |
| `NVLINK_MODE=${NVLINK_MODE:-auto}` | `auto` / `force_on` / `force_off` override (see `docs/HARDWARE.md`) |
| `ipc: host` | Matches other dual composes (NCCL) |
| Conditional `vllm serve` | NVLink → custom all-reduce on; PCIe → `--disable-custom-all-reduce` |
| `unset PYTORCH_CUDA_ALLOC_CONF` on PCIe path | `detect_nvlink.sh` sets `expandable_segments:True` on PCIe, which **breaks LMCacheMPConnector** (documented boot constraint in compose header). NVLink path keeps `detect_nvlink`'s `max_split_size_mb:512` only — compatible. |

## Test plan

- [ ] `docker compose -f models/qwen3.6-27b/vllm-lmcache/compose/dual/fp8/lmcache.yml config` — exit 0 (Compose v5 `$$` escaping from #429)
- [ ] **NVLink rig:** `docker logs … \| grep '\[nvlink\]'` → `detected NVLink … P2P ENABLED`; `nvidia-smi nvlink --getthroughput` shows traffic under TP load
- [ ] **PCIe rig (or `NVLINK_MODE=force_off`):** logs show PCIe mode; boot + serve unchanged
- [ ] LMCache still boots with connector enabled (no `expandable_segments` validation error on PCIe)
- [ ] Optional: `bash scripts/switch.sh --force vllm/qwen-27b-dual-lmcache` end-to-end on NVLink hardware

### DIFF
--- a/models/qwen3.6-27b/vllm-lmcache/compose/dual/fp8/lmcache.yml
+++ b/models/qwen3.6-27b/vllm-lmcache/compose/dual/fp8/lmcache.yml
@@ -109,15 +109,17 @@ services:
         #   LMCACHE_L2=1                -> enable L2 at the in-repo default /lmcache-kv (fs adapter)
         #   LMCACHE_L2_ADAPTER='{...}'  -> full override (custom path/backend); wins over LMCACHE_L2
         # chunk-size MUST be a multiple of vLLM's unified block (1584 for int8-PTH KV).
-        LM_ARGS=(--chunk-size "${LMCACHE_CHUNK_SIZE:-1584}" --l1-size-gb "${LMCACHE_L1_GB:-30}" --eviction-policy LRU)
-        if [ -n "${LMCACHE_L2_ADAPTER:-}" ]; then
-          LM_ARGS+=(--l2-adapter "${LMCACHE_L2_ADAPTER}")
-        elif [ "${LMCACHE_L2:-0}" = "1" ]; then
+        # $$ escapes $ so Docker Compose passes bash syntax through (v5.1+ interpolates
+        # entrypoint strings; $${VAR[@]} without escape → "invalid interpolation format").
+        LM_ARGS=(--chunk-size "$${LMCACHE_CHUNK_SIZE:-1584}" --l1-size-gb "$${LMCACHE_L1_GB:-30}" --eviction-policy LRU)
+        if [ -n "$${LMCACHE_L2_ADAPTER:-}" ]; then
+          LM_ARGS+=(--l2-adapter "$${LMCACHE_L2_ADAPTER}")
+        elif [ "$${LMCACHE_L2:-0}" = "1" ]; then
           LM_ARGS+=(--l2-adapter '{"type":"fs","base_path":"/lmcache-kv"}')
         fi
-        lmcache server "${LM_ARGS[@]}" &
+        lmcache server "$${LM_ARGS[@]}" &
         sleep 6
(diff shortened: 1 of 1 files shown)


noonghunna:
Merging — thank you @steamEngineer 🙏

Two things this nailed:

- **NVLink auto-detect for `vllm-lmcache`** — this was the one dual compose still missing it, and the PCIe-specific touch (unsetting `PYTORCH_CUDA_ALLOC_CONF` so `expandable_segments` doesn't fight `LMCacheMPConnector`) is a sharp catch.
- **The Compose v5.1 `$$`-escape** — your diagnosis here was the seed for a much bigger fix. It prompted an audit of every dual/multi compose, which found the *same* un-escaped-entrypoint bug silently killing NVLink detection on **9 more** composes (the qwen3.6-27b dual default + both `multi4` + the gemma-4-31b duals) — all baked to `if [ "0" = "1" ]` (the PCIe branch) at parse time on Compose v5.1+, regardless of hardware. That's now fixed repo-wide in #436 (merged), which deliberately left `lmcache` to this PR so yours landed cleanly.

Appreciate both the fix and the cross-rig NVLink validation you've been doing. A follow-up guard test (asserting no shipped compose carries the un-escaped form) is going in next so this can't regress.

PR #447, opened 2026-06-19
Fix estate GPU visibility overrides

Whamp:
## Summary
- Generate a per-instance Compose override during estate `up` so each instance sets both `CUDA_VISIBLE_DEVICES` and `NVIDIA_VISIBLE_DEVICES` inside the container.
- Prevent parallel vLLM estate instances from both trying to use physical GPUs 0/1 when Docker exposes all reserved devices.
- Add regression coverage for the generated override and `docker compose -f base -f override up` wiring.

## Validation
- `python3 -m py_compile scripts/lib/profiles/estate_cli.py`
- `bash scripts/tests/test-diagnose-estate.sh`
- `bash scripts/tests/test-estate-json.sh`
- `bash scripts/tests/test-profiles-compat.sh`
- `bash scripts/tests/test-launch-compat.sh`
- Live 4×3090 estate boot: `qwen-dual` on GPUs 0,1 and `gemma-dual` on GPUs 2,3 both healthy in parallel.

### DIFF
--- a/scripts/lib/profiles/estate_cli.py
+++ b/scripts/lib/profiles/estate_cli.py
@@ -368,6 +368,46 @@ def compose_cmd() -> list[str]:
     return shlex.split(os.environ.get("COMPOSE_BIN", "docker compose"))
 
 
+def compose_service_names(compose_name: str) -> list[str]:
+    """Return service names from a registry compose file."""
+    path = compose_abs_path(compose_name)
+    data = yaml.safe_load(path.read_text(encoding="utf-8")) or {}
+    services = data.get("services")
+    if not isinstance(services, dict) or not services:
+        raise EstateCliError(f"compose file for `{compose_name}` has no services: {path}")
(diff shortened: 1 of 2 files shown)


noonghunna (approved):
Thanks @Whamp — clean fix for a real pain point on multi-GPU rigs.

Tested locally: all 4 estate/compat tests pass (diagnose-estate, estate-json, profiles-compat, launch-compat). The override generation + `docker compose -f base -f override up` wiring looks correct.

Could you share the live 4×3090 boot log showing both `qwen-dual` (GPUs 0,1) and `gemma-dual` (GPUs 2,3) healthy in parallel? That would be a nice artifact to reference in the PR body for future readers.

noonghunna (commented):
Merged. If you have the parallel boot log handy (showing both instances seeing their claimed GPUs), it'd be good documentation — but not a blocker. Thanks for the fix.

PR #449, opened 2026-06-20
Fix NVLink boot crash when compose pre-sets expandable_segments

steamEngineer:
## Summary

Boot-crash fix for NVLink / patched-PCIe-P2P dual rigs, now hardened beyond the first pass.

Dual composes inject `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,max_split_size_mb:512` for the **PCIe path (custom all-reduce OFF)**. On a fast-interconnect rig `scripts/detect_nvlink.sh` turns **custom all-reduce ON**, where `expandable_segments` backs allocations with a `cuMemMap` VA range that `cudaIpcGetMemHandle` rejects during graph-buffer registration:

```
Failed: Cuda error /workspace/csrc/custom_all_reduce.cuh:455 'invalid argument'
Worker proc VllmWorker-1 died unexpectedly, shutting down executor.
```

The old `${VAR:-default}` preserved the compose-injected value, so the crashing `expandable_segments:True` survived on the P2P path.

## What changed vs the first commit

- **Surgical strip, not clobber.** Strip ONLY the `expandable_segments` token on the P2P / custom-AR path and preserve every other knob the user set (`max_split_size_mb`, `garbage_collection_threshold`, ...); fall back to `max_split_size_mb:512` only when nothing remains. The PCIe path keeps `expandable_segments` untouched. (The prior `export PYTORCH_CUDA_ALLOC_CONF=max_split_size_mb:512` silently discarded user overrides.)
- **`docs/UPSTREAM.md` row added** for [vllm#42609](https://github.com/vllm-project/vllm/issues/42609), per repo convention (link the tracker row, don't inline-cite in code). Its "DP>1 AND TP>1" title **under-scopes** the bug: the same `cuh:455` crash fires TP-only (cf. [vllm#42583](https://github.com/vllm-project/vllm/issues/42583), [vllm#45604](https://github.com/vllm-project/vllm/issues/45604)).
- **Test rewritten** to assert the surgical contract (preserve neighbors, strip in any position, no dangling commas, fall-back) and marked executable.

## Validation (2x RTX 3090, NVLink NV4, TP=2)

Faithful before/after on the real dual config, custom all-reduce ON both times:

- **Before** (`expandable_segments:True` surviving): boot crashes at `custom_all_reduce.cuh:455 'invalid argument'`, worker dies. Confirms the crash reproduces at plain **TP=2 / DP=1** (not just the DP>1+TP>1 case in the upstream title).
- **After** (`expandable_segments` stripped -> `max_split_size_mb:512`): boots clean, custom all-reduce registers cuda graph addresses, `Application startup complete`, `/v1/models` ready, serves a completion (`"Hi there friend."`).
- `bash scripts/tests/test-detect-nvlink-alloc-conf.sh` PASS (real bash on the rig); `bash scripts/tests/test-compose-nvlink-escape.sh` PASS (55 composes, no regression).

Lives entirely in the existing NVLink-support seam (`NVLINK_MODE` / `pcie_p2p`); no change to the no-NVLink default path.


steamEngineer:
## Rig validation + exposure matrix (what this actually fixes)

Ran a full A/B on the dual-3090 **NVLink (NV4)** rig to nail down exactly which configs this PR rescues and why. Short version: the crash is narrower and more deterministic than "expandable_segments breaks some setups" — and this PR covers every exposed config without touching the safe ones.

### Confirmed mechanism

A dual compose hits `custom_all_reduce.cuh:455 'invalid argument'` **iff all three hold at boot**:

1. **custom all-reduce ON** — i.e. the NVLink/P2P entrypoint branch (`exec vllm serve "$@"`, no `--disable-custom-all-reduce`)
2. **CUDA graphs ON** — no `--enforce-eager`
3. **`PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`** reaching the engine

It fires specifically during CUDA-graph-buffer IPC registration (`get_graph_buffer_ipc_meta`), not general allocation — which is why eager-mode composes never see it. TP>1 alone is enough; DP>1 is **not** required (matches what I posted on [vllm#42609](https://github.com/vllm-project/vllm/issues/42609)).

### diffusiongemma was a red herring (it's eager, not the image)

I'd wondered if its separate `vllm/vllm-openai:gemma` image was what saved it. It isn't. With the exact crash env present (verified `expandable_segments:True` + `NCCL_P2P_LEVEL=NVL` live in `/proc/<pid>/environ`):

| diffusiongemma run | CUDA graphs | Result |
|---|---|---|
| as-is (`--enforce-eager`) | OFF | clean boot + served |
| same compose, `--enforce-eager` dropped | ON | **crash `cuh:455`** at `Capturing CUDA graphs` |

Same image, same alloc conf, same custom-AR — flipping only graph capture flips the crash. Eager is the shield.

### A/B: this PR's `detect_nvlink.sh` strip (buggy master vs fixed branch)

| Compose (TP=2, NVLink, custom-AR ON, graphs ON) | master (clobber `${VAR:-default}`) | this PR (strip token) |
|---|---|---|
| `qwen3.6-27b autoround-int4/fp8-mtp` (dual default) | crash `cuh:455` | clean + served, ~22 GB/card |
| `qwen3.6-27b fp8/mtp` | crash `cuh:455` | clean + served, ~21.5 GB/card |

The fixed branch logs `expandable_segments stripped (PYTORCH_CUDA_ALLOC_CONF=max_split_size_mb:512)` and graph capture then succeeds. (Note: the master message even mislabels it — prints "expandable_segments OFF" while `/proc` showed `expandable_segments:True` still live.)

### Latest release still needs this guard

| v0.23.0 (latest) + qwen default | Result |
|---|---|
| master script | crash `cuh:455` |
| this PR's script | clean + served |

The in-engine fix [vllm#43923](https://github.com/vllm-project/vllm/pull/43923) is still open, so it's in **no** release (v0.22.x or v0.23.0). This launch-time guard remains necessary until that lands and we pin past it.

### Exposure matrix — what's impacted vs already safe

**Exposed (this PR rescues these):** custom-AR ON on NVLink + graphs ON + expandable_segments
- `qwen3.6-27b` `autoround-int4/fp8-mtp` (dual default), `fp8/mtp`, `awq-bf16-int4/int8`
- `gemma-4-31b` `autoround-int4/bf16-mtp` (dual default), `qat-w4a16/int8`
- `gemma-4-26b-a4b` `awq/mtp`
- `qwen3.6-35b-a3b` `autoround-int4/fp8`

**Already safe (untouched by this PR):**
- Shielded by `--enforce-eager`: `diffusiongemma fp8/base`
- Shielded by hardcoded `--disable-custom-all-reduce`: the older nightly/flat composes (`qwen int8`, `tq3-nomtp`, `tq3-mtp-genesis`, several `gemma-4-31b` nightly, `gemma-4-12b bf16/mtp`)
- Shielded by omitting expandable_segments entirely: `vllm-lmcache/lmcache`

So the surgical strip is exactly the right scope: it removes the one token that breaks the graphs+custom-AR path, preserves every other knob, and leaves the already-safe composes unaffected.

_(Tested on 2x RTX 3090 / NVLink NV4, TP=2 DP=1, vLLM v0.22.0 + v0.23.0. Weights present on the rig limited the live A/B to the two qwen composes; the gemma/35b rows are exposed by inspection — identical custom-AR-ON + graphs-ON entrypoint.)_

noonghunna (approved):
LGTM. Surgical fix — strips only the crashing `expandable_segments` token on the P2P/custom-AR path while preserving other alloc knobs. Tests pass, live-validated. Thanks @steamEngineer .

noonghunna:
Thanks @steamEngineer — merged via manual merge commit (GitHub's mergeability checker was stuck on a false conflict, but the branch was already clean against master). Surgical fix, clean tests, solid UPSTREAM.md tracking. Live validation on your 2×3090 NVLink rig confirms the custom_all_reduce IPC handle issue is resolved.

PR #452, opened 2026-06-20
docs: fix vLLM upstream table rendering in UPSTREAM.md

steamEngineer:
## Summary
- The `## vLLM (vllm-project/vllm)` table in `docs/UPSTREAM.md` rendered as literal `|` text instead of a table on GitHub.
- Root cause: a stray blank line between the table delimiter row (`|---|---|---|---|`) and the first data row. GitHub Flavored Markdown requires data rows to be contiguous with the delimiter; the blank line terminates the table early, so every row below renders as raw text.
- Fix: delete the single stray blank line. No content changes.

## Notes
- Pre-existing render bug on `master`, independent of #449.
- Docs-only change; no compose/engine touched, so the verify-full/verify-stress gate does not apply.


noonghunna (approved):
LGTM. One-line GFM table fix. Thanks.

noonghunna:
Thanks @steamEngineer — merged.

PR #461, opened 2026-06-24
detect_nvlink: auto-enable PCIe P2P when nvidia-smi reports it OK

noonghunna:
## Summary
`auto` mode now auto-enables **PCIe P2P** (not just NVLink). After the NVLink check, it probes `nvidia-smi topo -p2p r` and, if P2P is **OK across all pairs** (a patched consumer-GPU driver — tinygrad/geohot/aikitoria — on a P2P-capable layout: shared root complex / switch), promotes to PCIe P2P with **custom all-reduce ON** (`NCCL_P2P_LEVEL=PHB`). Previously this needed a manual `NVLINK_MODE=pcie_p2p`.

Motivated by the aikitoria `open-gpu-kernel-modules` P2P patch — users who run it on a shared-root-complex layout now get custom all-reduce automatically; `pcie_p2p` remains as a **force** override.

## Safe by construction
- **No-op for stock drivers / separate-root-complex rigs** — they report `CNS`/non-OK → stays off. Verified on the maintainer's 2× 3090 (cards on separate EPYC root complexes): `auto` still resolves **P2P DISABLED**.
- Requires **all** off-diagonal pairs == `OK` (partial P2P → off — custom all-reduce wants full P2P).
- `force_on` / `force_off` / `pcie_p2p` unchanged.

## Tests
- New `test-detect-nvlink-autop2p.sh` — 6 mocked cases (enable on OK · stay off on CNS · NVLink wins · 4-GPU all-OK vs mixed vs none).
- Existing `test-detect-nvlink-alloc-conf.sh` unchanged. Full `scripts/tests` gate: **58/58**.

### DIFF
--- a/scripts/detect_nvlink.sh
+++ b/scripts/detect_nvlink.sh
@@ -5,17 +5,38 @@
 # Handles 2-GPU setups (single NVLink bridge) and N-GPU setups (e.g. 2 bridges on 4 cards).
 #
 # NVLINK_MODE values:
-#   auto       — detect NVLink via nvidia-smi topo (default). No NVLink => P2P off.
+#   auto       — detect a fast P2P interconnect via nvidia-smi (default): NVLink (topo -m)
+#                OR, failing that, PCIe P2P that `nvidia-smi topo -p2p r` reports as OK
+#                between all pairs — i.e. a patched consumer-GPU driver (tinygrad/geohot/
+#                aikitoria) on a P2P-capable layout (shared root complex / switch). Neither
+#                => P2P off. NOTE: stock GeForce drivers software-disable P2P (report CNS),
+#                and cards on separate root complexes can't P2P — both correctly stay off.
 #   force_on   — assert NVLink present (NCCL_P2P_LEVEL=NVL).
 #   force_off  — no P2P at all (NCCL_P2P_DISABLE=1).
-#   pcie_p2p   — PCIe P2P WITHOUT NVLink (e.g. a patched consumer-GPU driver — the
-#                tinygrad/geohot P2P patch). Sets NCCL_P2P_LEVEL=PHB (or your own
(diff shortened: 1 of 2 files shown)


noonghunna:
### Note — prerequisites for `auto` to actually enable P2P

To be clear about scope: this auto-detect **only *uses* P2P when it's already available — it does not create it.** Two things, both *outside* this script, must be in place first or `auto` correctly stays in PCIe-no-P2P mode:

1. **A patched driver.** Stock NVIDIA GeForce drivers *software-disable* P2P — `nvidia-smi topo -p2p r` reports `CNS`. Without a consumer-GPU P2P patch the probe never returns `OK`, so `auto` leaves P2P off. Patches:
   - **[aikitoria/open-gpu-kernel-modules](https://github.com/aikitoria/open-gpu-kernel-modules)** — patches the **610.43.02** driver, supports **RTX 3090** (PCIe BAR1) / 4090 / 5090; needs `iommu=pt`.
   - **[tinygrad/open-gpu-kernel-modules](https://github.com/tinygrad/open-gpu-kernel-modules)** — the original (4090) P2P patch it derives from.
2. **A P2P-capable PCIe layout — the GPUs on the same root complex / PCIe switch.** Cards on *separate* root complexes can't P2P regardless of the driver (the DMA can't route across them), so `topo -p2p r` stays non-`OK` and `auto` stays off.

When **both** are true, `topo -p2p r` reports `OK` across all pairs and `auto` flips on custom all-reduce automatically. When either is missing it stays off — e.g. the maintainer's rig (2× RTX 3090 on **separate EPYC root complexes**) reports `CNS` and correctly resolves *P2P DISABLED*.

`NVLINK_MODE=pcie_p2p` remains the **force override** for asserting P2P when you trust the patch but the probe doesn't report `OK`.

Issue #482, opened 2026-06-26
[bug] services/litellm/config.yaml pins a scene-specific served-model-id → tool-calls silently 404 after a gpu-mode scene switch


alexpolo1:
## Summary

`services/litellm/config.yaml` hardcodes the upstream `model:` (the vLLM **served** id) to a single scene's value. When `gpu-mode` brings up a *different* serving scene, the served id no longer matches and LiteLLM 404s that route — silently breaking tool-calling for every client routed through it. The public `model_name` stays the same, so clients see "model exists" but every call fails upstream.

## Repro (observed 2026-06-26)

- Shipped config: `model_name: qwen3.6-27b-autoround` → `model: openai/qwen3.6-27b-autoround`.
- Live scene was the LMCache profile (`vllm/qwen-27b-dual-lmcache`, container `vllm-qwen36-27b-lmcache` at `:8010`), whose vLLM `--served-model-name` is **`qwen3.6-27b-fp8`**.
- `GET /v1/models` on `:8010` returns only `qwen3.6-27b-fp8` → LiteLLM's upstream `qwen3.6-27b-autoround` 404s.
- Symptom: LiteLLM 30-min health check fails; Hermes/clients get no tool-calls (the request never reaches a live model).

## Root cause

The route's upstream `model:` is tied to **which scene is active**, but `gpu-mode` can switch scenes (autoround dual ↔ lmcache fp8 ↔ …) independently of this file. Nothing keeps them in sync, so the config silently drifts out of validity after a scene change.

## Fix applied locally (proposed)

Keep the public `model_name` for client back-compat; repoint the upstream `model:` to the served id of the live scene:

```diff
   - model_name: qwen3.6-27b-autoround
     litellm_params:
-      model: openai/qwen3.6-27b-autoround
+      model: openai/qwen3.6-27b-fp8
       api_base: http://host.docker.internal:8010/v1
```

## Open question for maintainers (the durable fix)

The diff above re-pins to fp8, but that just moves the coupling — switch back to the autoround scene and it 404s the other way. Options to actually fix the drift class:

- **(a)** Make every scene's vLLM `--served-model-name` include a **stable alias** (e.g. all scenes also serve `qwen3.6-27b`), so the LiteLLM `model:` can target the alias and survive any scene switch.
- **(b)** Have `gpu-mode <scene>` rewrite / template this `model:` line when it brings a scene up (single source = the active scene).
[...]

PR #484, opened 2026-06-26
Fix LiteLLM served-model-id drift breaking tool-calls

alexpolo1:
Repoints the LiteLLM route's upstream `model:` from `qwen3.6-27b-autoround` to the live scene's served id `qwen3.6-27b-fp8`, keeping the public `model_name` for client back-compat.

**Why:** with the LMCache scene live (`vllm-qwen36-27b-lmcache` serving `qwen3.6-27b-fp8`), the old id 404s on `:8010` — LiteLLM's health check fails and tool-calls never reach a model.

Stopgap — see the issue for the durable scene-agnostic-alias fix still open for maintainers.

Refs #482

### DIFF
--- a/services/litellm/config.yaml
+++ b/services/litellm/config.yaml
@@ -2,11 +2,14 @@ model_list:
   # === Current serving primaries (one GPU service runs at a time per gpu-mode;
   #     routes to the inactive backends fail until that mode is brought up) ===
 
-  # Qwen 3.6 27B (Lorbus AutoRound INT4) + MTP n=3 + fp8 KV + 262K + vision.
-  # Started by `gpu-mode 27b` → vllm-qwen36-27b-dual at :8010. Stack default.
+  # Qwen 3.6 27B (fp8) + LMCache + fp8 KV + 262K + vision.
+  # Live scene: vllm-qwen36-27b-lmcache at :8010, serving id `qwen3.6-27b-fp8`.
+  # Public model_name kept as -autoround for client back-compat; upstream `model:`
+  # repointed to the real served id (the old dual/autoround scene was retired —
+  # mismatch was 404ing LiteLLM's 30-min health check). Stack default.
   - model_name: qwen3.6-27b-autoround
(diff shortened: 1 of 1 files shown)


noonghunna:
Thanks @alexpolo1 — great diagnosis on the drift, but I'm going to hold this one. The `gpu-mode 27b` ⭐ default scene still serves **`qwen3.6-27b-autoround`** (`vllm-qwen36-27b-dual`, the `autoround-int4` compose) — it isn't retired. The lmcache/fp8 scene (`qwen3.6-27b-fp8`) is 🐣 incubating + opt-in (`--force`, hidden from `--list`), which is the one you had live when you hit the 404.

So repointing the LiteLLM upstream to `fp8` would unblock the opt-in scene but **404 the default for everyone else** — it just flips the coupling. The committed config (autoround) is correct for the default.

I'd rather land the **durable scene-agnostic fix** you flagged in #482: have `gpu-mode` rewrite the LiteLLM upstream `model:` to the active scene's served-name on each switch (+ restart litellm if it's up), so it's always in sync regardless of which scene is live. Putting that up shortly and linking it here. Closing this in favor of that — but the catch + the LMCache repro were spot-on, thank you.

alexpolo1:
Makes total sense — thanks for the correction, and you're right: I was on the opt-in fp8 scene and mistook it for the default being retired. Pointing the shared route at `fp8` would just flip the 404 onto everyone running the autoround default. Agreed the committed config is correct as-is.

The `gpu-mode`-rewrites-the-upstream-on-switch approach in #482 is clearly the right shape — keeps the route in sync with whatever scene is actually live. Happy to test it on the dual-3090 rig once it's up. Thanks for the quick, careful review. 🙏

noonghunna:
Follow-up: the proper fix landed in #490 — standardized `--served-model-name` to `qwen3.6-27b-autoround` across all 27b scenes (the shared-name direction you pointed at in your reply). Thanks again — your report is what surfaced it.

PR #549, opened 2026-07-04
tools: add HTTP model-switch service (thin wrapper over switch.sh)

titan550:
## What
A HTTP control service that wraps `scripts/switch.sh`, so an experiment harness can `POST /switch` and block until the new model is serving. Only one model fits in VRAM at a time, so cross-model runs mean repeated switches — this makes that programmatic.

## Design
- **Wraps, doesn't reimplement** — shells out to `switch.sh` (registry lookup → down → up → readiness). No new orchestration.
- Mirrors the repo's stdlib `ThreadingHTTPServer` style (`services/studio/*`). **Zero new deps.**
- Registry-validated slugs; `/health` readiness so it works with or without `VLLM_API_KEY`; single-flight lock; refuses to start unauthenticated on a non-loopback bind.

## Endpoints
`GET /healthz` (no auth) · `GET /status` · `GET /models` · `POST /switch {slug|model}` (blocks until ready).

## Scope
No changes to `switch.sh`; no compose changes. Marked experimental/opt-in until cross-rig validated.

## Test
`bash scripts/tests/test-model-switch.sh` — hermetic (stubbed switch + docker), asserts the HTTP/auth/validation/lock contract.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -240,3 +240,10 @@
 # GENESIS_ENABLE_P40=1
 # GENESIS_ENABLE_P82=1
 # SOAK_TIMEOUT_S=3600
+
+# --- model-switch service (tools/model-switch) — optional HTTP model swapper ---
+# Bearer token for the control endpoint. Falls back to VLLM_API_KEY if unset;
+# if BOTH are empty the endpoint is unauthenticated (loopback only).
+# CLUB3090_API_TOKEN=
+# MODEL_SWITCH_PORT=8099
+# MODEL_SWITCH_BIND=127.0.0.1
--- a/docs/EXAMPLES.md
+++ b/docs/EXAMPLES.md
@@ -366,8 +366,33 @@ Or override at run-time with `--host 127.0.0.1` (llama.cpp) / by editing the com
 
 ---
 
+## Switch the served model over HTTP
+
+On a 1–2 GPU rig only one model fits in VRAM at a time. The optional
+[`tools/model-switch`](../tools/model-switch/README.md) service wraps
+`scripts/switch.sh` behind an HTTP endpoint so an experiment harness can swap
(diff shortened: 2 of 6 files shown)


noonghunna (approved):
This is a model community PR — thank you. Reviewed + fully validated on the maintainer rig before merging:

- **Hermetic test**: `bash scripts/tests/test-model-switch.sh` → ok (and it slots straight into the `scripts/tests/*.sh` gate).
- **Live end-to-end** (2× 3090): idle `/status` nulls · 401 without/with wrong token · `/models` lists the registry · **`POST /switch {"slug":"vllm/minimal"}` blocked 272.6s → `{"ok":true}`**, `/status` then correct (model id, slug via the container reverse-lookup, port, ready) · repeat call hit the `already-running` fast path at 0s · unauthenticated non-loopback start refused.

Design notes that stood out: it *actually* wraps `switch.sh` (no re-orchestration — exactly the layering we hold our own tools to), the argv-list subprocess with registry-validated slugs closes the injection question, `hmac.compare_digest` on the token, and the no-op fast path comparing **by container** rather than model id dodges a slug-masquerade trap we fixed in our own estate detection just this week (#554). The test asserting the current curated gemma default was a nice touch.

Two tiny non-blocking follow-ups (merging as-is; happy if you want to pick them up, otherwise we will):
1. `_running_model()`'s name-prefix tuple misses `sglang-` — an sglang container reads as "nothing serving" (fail-safe direction, just a status blind spot).
2. One README line about writer coordination: this daemon is a third writer alongside the CLI and the c3 cockpit — `switch.sh`'s own teardown/preflight keeps the rig safe, but don't drive switches mid-validation-run.

Merging as 🧪 experimental per your own marking. Welcome aboard.

PR #650, opened 2026-07-09
feat(c3): GGUF route-G fit + compose emit

noonghunna:
## Summary
Phase 4 of the c3 UI/UX plan — **GGUF route-G**.

1. **GGUF fit** — `CockpitData.byo_check_gguf` (selected quant size + card VRAM headroom); never hits the vLLM/safetensors deriver `unsupported-format` wall
2. **Download** — reuses existing Bring `[D]` / pull lock / % / cancel (#643–#645); size from GGUF inventory pick
3. **Serve compose emit** — `emit_gguf_compose` clones a GGUF-engine sibling compose (`llama-cpp` / `ik-llama` / `beellama`), rewrites `--model` → brought `.gguf` (+ optional mmproj)

## Stack
- Branched from **`c3-uiux-phase3`** tip (phases 1–3 not yet on master); PR base = `c3-uiux-phase3` so the stack stays linear.
- After phases 1–3 merge, this can be rebased onto `master`.

## Maintainer-gated (not in this PR)
- Live boot + verify of the emitted GGUF compose on GPU

## Test plan
- [x] `pytest tests/ -q` → **835 passed**
- [x] `tests/test_uiux_phase4_gguf.py` — fit verdicts + emit rewrites model path / non-vLLM image

Co-Authored-By: Grok <noreply@x.ai>

### DIFF
--- a/tools/serve-cockpit/club3090_cockpit/app.py
+++ b/tools/serve-cockpit/club3090_cockpit/app.py
@@ -8084,10 +8084,16 @@ class CockpitApp(App):
                 pass
             return
         # Phase 4 route-G: GGUF pick → bypass vLLM/safetensors deriver.
+        # Budget = 24 GiB × topology cards of the chosen sibling (dual/multi).
         gguf_quant, gguf_gb = self._funnel_gguf_selection()
         if gguf_quant:
+            cards = self._data.topology_cards_for_profile(profile_like)
             res = self._data.byo_check_gguf(
-                repo, profile_like, quant=gguf_quant, size_gb=gguf_gb or 0.0,
+                repo,
+                profile_like,
+                quant=gguf_quant,
+                size_gb=gguf_gb or 0.0,
+                card_vram_gb=24.0 * max(1, cards),
             )
         else:
             res = await self._data.byo_check(repo, profile_like)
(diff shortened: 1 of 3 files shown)


noonghunna:
## Review: GGUF route-G — needs changes before merge

Rebased onto master (phases 1–3 landed as #647/#648/#649), so this is now a clean single commit on master and the fixes below apply directly.

The route-G **foundations are sound**: it correctly bypasses the deriver's `unsupported-format` wall, constrains GGUF picks to the llama.cpp/beellama engine family (never vLLM), routes serve through the normal confirm/reconcile gate, and makes no false "validated serve" claim (live boot stays maintainer-gated). But `emit_gguf_compose` does not produce a bootable compose for any *shipped* GGUF sibling, and the fit-check is single-card-only. Four items:

### 1. BLOCKER — `emit_gguf_compose` explodes `command:` into per-character args
`services.py` (`emit_gguf_compose`) does `cmd = list(svc.get("command") or [])`. Every shipped GGUF sibling declares `command:` as a **folded block scalar** (`command: >-`), which `yaml.safe_load` returns as a `str`, so `list(str)` yields one element *per character*. Reproduced against `models/qwen3.6-27b/ik-llama/compose/single/ubergarm-iq4ks/mtp.yml`:

```
type(command) == str
list(command)  -> 652 single-char elements: ['-','-','h','o','s','t',' ','0', ...]
no token ever == '--model' / '-m'  ->  the rewrite loop never fires
```

Result: the emitted `command:` is ~650 one-char argv tokens and `--model` is never repointed at the brought GGUF — the compose can't boot. **Fix:** detect the `str` form and `shlex.split()` it before iterating (then re-emit as a list). All five shipped siblings use `>-` — llama-cpp `mtp.yml` / `mtp-vision.yml`, ik-llama `mtp.yml`, beellama `dflash.yml` — so this hits every real path.

### 2. Sibling drafter flags leak into the brought model's compose
The rewrite only touches `--model`/`-m`, leaving the sibling's `--spec-type draft-mtp` / `--spec-draft-model /models/…anbeeld-dflash…` pointing at the **Qwen/Anbeeld** drafter — wrong tokenizer/vocab for an arbitrary brought model, and (beellama) a draft path that won't exist in the brought container. **Fix:** strip the sibling's drafter flags when emitting for a foreign model. Bonus: if the brought repo ships its own `mtp-*.gguf`, wire *that* as the drafter instead — the `migtissera/Tess-4-27B-GGUF` repo is the canonical multi-artifact case (main GGUF + `mtp-*.gguf` + `mmproj-*.gguf`).

### 3. mmproj is appended-if-absent, not rewritten
`if not any(x.startswith('--mmproj') …): out_cmd += ['--mmproj', '/models/brought.mmproj']`. A **vision** sibling (`mtp-vision.yml`) already carries `--mmproj`, so the brought projector is mounted but ignored — the sibling's projector is used instead. **Fix:** rewrite the existing `--mmproj` value to the brought projector, not only append when missing.

### 4. Fit-check is single-card-only, contradicting the funnel's own dual/multi filter
`byo_check_gguf` defaults `card_vram_gb=24.0` and the caller never passes topology, so a large GGUF is always judged against one 24 GiB card. But `funnel_slug_options` already offers **dual/multi** siblings for large GGUF (`_FUNNEL_WEIGHTS_VRAM_FRAC`), so a 34 GB Q8 paired with a dual sibling returns `wont-fit` → route blocked. **Fix:** thread the chosen sibling's topology → card budget (24 × N). Also the KV term is a flat `+2.0 GiB`; the plan asked for a ctx-aware llama.cpp KV estimate (reuse `tools/kv-calc.py`) and to **sum** mmproj + drafter file sizes — the inventory currently carries mmproj *names* without sizes and has no drafter-size field, so this needs a small model extension.

### Test gap that masked #1
`test_uiux_phase4_gguf.py`'s emit test feeds a **list-form** command (`["--host", …, "--model", "/old.gguf"]`) — a shape no shipped compose uses — so it round-trips fine and hides the `>-`→`str` explosion. Please add: (a) a case using a real `command: >-` sibling (folded scalar); (b) a vision-sibling case asserting `--mmproj` is *rewritten*, not appended; (c) a case asserting no sibling drafter flags survive into the emit.

Engine-choice, no-false-boot, gitignored output, and the deriver-wall bypass are all correct — no changes needed there.

Issue #665, opened 2026-07-09
[bench] Tess-4-27B on single RTX 4090 (llama.cpp DFlash n=4, 96K) — 117/150 off · 125/150 on, beats dual-3090 ref


seanyourhighness:
> ### ⚠️ Read this before parsing the pasted report
>
> My llama-server runs `--alias qwen3.6-27b` for downstream-routing reasons, so every
> auto-captured line in the report below reads `model=qwen3.6-27b`. **The weights are
> `Tess-4-27B-Q4_K_M.gguf`** (migtissera). These are **Tess** numbers, not Qwen3.6
> numbers. Sorry for the trap — the alias is load-bearing in my stack and I didn't want
> to restart the server mid-benchmark and invalidate the run.

## Exact serving command

```bash
llama.cpp b9932 (a646006f0), host build, no container

llama-server \
  -m Tess-4-27B-Q4_K_M.gguf \
  --model-draft Qwen3.6-27B-DFlash-Q4_K_M.gguf \
  --spec-type draft-dflash --spec-draft-n-max 4 \
  --alias qwen3.6-27b \
  --ctx-size 98304 --gpu-layers all \
  --cache-type-k q4_0 --cache-type-v q4_0 --flash-attn on \
  --jinja --chat-template-file qwen36-vanilla-embedded-unsloth.jinja \
  --batch-size 2048 --ubatch-size 512 --parallel 1 \
  --threads 16 --threads-batch 16 \
  --temp 0.6 --top-p 0.95 --top-k 20 --min-p 0.75
```

GPU: 1× RTX 4090 (24564 MiB), driver 610.62, WSL2.
club-3090 @ `977d34f` · benchlocal-cli `0.9.7`.

Note the DFlash drafter is the **stock Qwen3.6-27B** DFlash head running against a
**Tess** target — an off-distribution pairing. It still works (aggregate draft
acceptance ~0.70 on short-ctx agent traffic), which may itself be of interest given
`INTERNALS.md`'s DFlash forensics.

Quality packs ran at their canonical **temp 0** (I did *not* pass
`--sampling-from-server`), so the server's `--temp 0.6 --min-p 0.75` defaults did not
affect the scores.

## What this contributes

A **single-card 4090** data point for **Tess-4-27B**, which the catalog's Tess compose
explicitly calls out as missing:

> *"a single-card long-ctx variant is architecturally valid but not yet benched."*
[...]

PR #674, opened 2026-07-11
BENCHMARKS: single-4090 Tess-4 + Eagle3 8-pack quality (124/150 on · 115/150 off)

seanyourhighness:
Adds a **`### Single-card (1× RTX 4090) — llama.cpp`** subsection under `## Tess-4-27B` — the first single-4090 data point for Tess-4-27B with an **Eagle3 draft** on mainline llama.cpp b9932.

### 8-pack (`quality-test.sh --full`, 150 scenarios, packs v1.0.x)
| | think-on | think-off |
|---|---:|---:|
| **TOTAL** | **124/150 (82.7%)** | **115/150 (76.7%)** |

Per-pack ON: toolcall 15·instructfollow 15·structoutput 14·dataextract 14·reasonmath 13·bugfind 13·hermesagent 15/20·cli-40 25/40
Per-pack OFF: 14·13·14·12·11·12·15/20·24/40

### Config
`Tess-4-27B-Q4_K_M` + Tess-matched Eagle3 draft (`…-EAGLE3-Q4_K_M`, 0.39 GB) via `--spec-type draft-eagle3 --spec-draft-n-max 2`, pinned vanilla Unsloth template, q4_0/q4_0 KV, ctx 98304, b4096/u1024, `--reasoning on`, sampler `min-p 0` (temp 0.6/top-p 0.95/top-k 20). Single RTX 4090 (WSL2).

### Notes
- **vs same-rig DFlash baseline (2026-07-09, min-p 0.75): 125 on / 117 off** → within 1–2 scenarios (single-run noise). Eagle3 is quality-equivalent to DFlash (both output-lossless); the `min-p 0.75→0` sampler change costs no measurable quality.
- **`--spec-draft-n-max` sweep (this rig):** n_max=2 optimal (~76 code TPS), 4→68, 6→54, 8→47 — deeper drafting is strictly worse for Tess-4.
- Quality-only submission — throughput cells left un-rebenched (`~76` code TPS carried from the n_max sweep, not a formal `bench.sh` run).


noonghunna:
This is a genuinely important data point, beyond the numbers — thanks for the careful labeling (informal TPS flagged, "no compose yet", full config inline; exactly the shape we want these rows in).

**Why it matters: your run completes the EAGLE3 mechanism story.** In vLLM, this same head accepts at ~40% — we traced that to trunk-feature misalignment (the head is byte-identical to the base Qwen3.6 head; the fine-tune shifted the hidden-state space it was trained to read — forensics in #662). Your external EAGLE3 on mainline llama.cpp is **token-fed**: it never sees the drifted representations, exactly like the external MTP (62% accept) and DFlash paths. With your row, all three token-fed drafters (MTP, DFlash, EAGLE3) are now measured lossless-and-paying on the GGUF path, while every hidden-state-fed variant fails on this fine-tune. That's the cleanest cross-engine confirmation of the mechanism we have — and your n_max=2 sweet spot matches our MTP n-sweep on the dual (also n=2, same deep-draft decay shape).

**One ask before/after merge (either way):** where does `Tess-4-27B-EAGLE3-Q4_K_M` come from? Self-converted from `migtissera/Tess-4-27B-EAGLE3` — and if so, with which convert script/commit? A one-liner in the row (or a link if you've published the GGUF) makes it reproducible for the next 4090 owner; we'd fold the recipe into the model README.

**Engine context worth knowing (post-dates your run):** the 118-vs-125 chase in #665 resolved to the engine build — think-ON totals read ~4 points higher on newer llama.cpp builds (greedy think-OFF ties exactly), and we've bumped the fleet pin to `server-cuda-b9967` (#680). No change needed here: your row is dated and build-labeled (b9932), which is precisely why the convention exists. Just expect think-ON comparisons against newer-pin runs to skew a few points.

Merging — thanks again. Your OFF 115 landing exactly on our dual-rig OFF 115 (two rigs, two drafters, two builds) is about as solid as a cross-rig anchor gets.

seanyourhighness:
Glad the row slots into the mechanism story — the token-fed vs hidden-state-fed framing also retroactively explains my n_max sweep shape better than my "deep drafts waste verification" hand-wave did.

Provenance, fully reproducible:

**Source:** `migtissera/Tess-4-27B-EAGLE3` (HF, `LlamaForCausalLMEagle3`, 610M, BF16 safetensors). Self-converted — no published GGUF exists yet (happy to push BF16/Q8_0/Q4_K_M to HF if you'd rather link than recipe).

**Recipe** (llama.cpp **b9932 / `a646006f0`** — deliberately the same checkout that served the benchmark, so one pin reproduces the whole row):

```bash
# target dir needs ONLY the base model's config.json + tokenizer files —
# no 27B weights required (~20 MB download, not ~55 GB)
python convert_hf_to_gguf.py <Tess-4-27B-EAGLE3-dir> \
  --target-model-dir <Tess-4-27B-metadata-dir>        # → BF16 GGUF (1.2 GB)
llama-quantize Tess-4-27B-EAGLE3-BF16.gguf Tess-4-27B-EAGLE3-Q4_K_M.gguf Q4_K_M   # → 403 MB
```

The converter reads target hidden size 5120 from the target config, selects extraction layers `[2, 32, 61]`, and writes the 32K→248K `d2t` vocab map into the GGUF (`eagle3.target_layers` / `eagle3.target_hidden_size` metadata, verifiable with `llama-gguf`). Q8_0 (629 MB) same way; my sweep found Q4_K_M draft quantization quality-free and fastest (acceptance is flat across draft quants — 27.4–28.0% at n=4 — so the smallest draft wins on latency).

One row-level suggestion when you fold this in: worth noting the metadata-only `--target-model-dir` trick explicitly — it's the difference between "clone two repos, 20 min" and "download the full BF16 base" for the next 4090 owner.

Noted on the b9967 think-ON skew from #665 — my next full-pack runs will be on the new pin, ×3 per the ON contract, so the dated b9932 rows stay as-is and the next rows will be directly comparable to fleet numbers.

noonghunna:
Folded — the row now carries the full provenance including the `--target-model-dir` metadata-only trick, called out exactly as you suggested (6da2456), with your comment linked as the canonical recipe.

**And yes please on pushing to HF** — a linkable artifact beats a recipe for most users. All three (BF16/Q8_0/Q4_K_M) would be ideal: your own sweep shows Q4_K_M is the one people should serve, but the BF16 is the one a future re-quant or a different engine's converter wants to start from. If you put the recipe + the acceptance-flat-across-draft-quants finding in the model card, we'll repoint the BENCHMARKS link at the repo and it becomes the reference EAGLE3-GGUF for Tess. Worth a heads-up to @trinity-cloud too — it's his head finding a consumer path.

Two things your numbers quietly settled, for the record:

- **The baked extraction layers are (2, 32, 61)** — the same taps vLLM defaults to when the config omits `eagle_aux_hidden_state_layer_ids`. So the llama.cpp and vLLM paths were reading the *same taps* all along, and the acceptance split (your working ~28% token-fed vs the ~40%-but-never-paying hidden-state path) is purely the feed mechanism plus draft economics — a 403 MB draft is cheap enough that 28% pays, while vLLM's in-graph path can't win at 40% against a fast quantized trunk. Cleaner than any explanation we had.
- **Draft-quant quality-freeness** extends the `mtp.fc` law from #662 in an interesting direction: *target-side* head precision is critical (quantize it → 0% accept), *draft-side* precision is free (BF16 ≡ Q4_K_M). Precision matters where the hidden states live, not where the tokens live — one more datapoint for the token-fed framing.

Good call keeping the b9932 rows dated and immutable — ×3 on b9967 whenever the card frees up, and between that and our shipped-config refresh the engine-vs-numerics residual from #665 closes itself.

noonghunna:
Recipe reproduced end-to-end, and we owe you one — it went from your comment to a serving artifact in under 20 minutes, with zero local toolchain (the `full-b9967` CPU docker image bundles both the converter and `llama-quantize`, so the whole thing ran as two `docker run`s). Every value you stated verified at our pin: BF16 1.22 GB, Q4_K_M 402.5 MiB, `eagle3.target_layers [2,32,61]`, `target_hidden_size 5120`, the 248320-vocab d2t map. The metadata-only `--target-model-dir` trick worked exactly as advertised — 12 MB of config+tokenizer instead of a 55 GB base pull. Recipe is version-robust across the b9932→b9967 gap.

Then we ran the head-to-head that was still missing: **EAGLE3-external vs MTP-external, same dual-3090, same session, fresh boots, n=2 both, only the draft file + `--spec-type` changed:**

| arm | accept (aggregate) | narr / code wall TPS |
|---|---:|---:|
| MTP external (shipped) | **0.737** | **63.4 / 73.5** |
| EAGLE3 external (your recipe, our build) | 0.408 | 47.5 / 61.3 |

**MTP wins decisively here — no EAGLE3 compose ships**, artifact kept for the day a stronger head lands (it drops into the same compose shape unchanged). Your 27.4–28.0% at n=4 and our 0.408 at n=2 are the same story once per-position decay is accounted for.

The part worth keeping: **token-fed EAGLE3 accepts ~0.40 — statistically the same as the ~0.40 the identical head scores hidden-state-fed in vLLM.** Feed mechanism changes nothing for this head. Contrast the MTP head: 0% hidden-state-fed, 0.74 token-fed. So the two drafters fail in opposite ways — **EAGLE3 is feed-robust but capability-limited; MTP is capability-strong but alignment-limited** — and the winner is the strong head with its feed fixed. It also cleanly explains your rig: EAGLE3 viable at ~76 code TPS, but your DFlash (a ~0.70-accept head) at 118 — this head works everywhere and wins nowhere on Tess until someone trains a stronger one.

@trinity-cloud this one's for you as much as for Sean — it sharpens the #662 pitch by one notch: your EAGLE3's ~40% ceiling is now confirmed **feed-independent** (same acceptance in-graph and token-fed, two engines), so more EAGLE3 training data would lift it *everywhere at once* — and a realigned MTP-class head remains the bigger prize (the 0.74-token-fed head shows what the trunk supports; realignment unlocks that class of acceptance *inside vLLM* where it's currently 0%). Either path drops into composes that already exist on both engines, benchable against exactly these numbers.

Also incidental but real: the MTP arm doubled as a fresh b9967 re-bench of the shipped compose — 63.4/73.5 wall vs the 52/68 on the b9246 row (+22% narr / +8% code). The pin bump bought throughput as well as the think-ON quality. Both rows are in BENCHMARKS now (dated, pin-labeled), your recipe linked as canonical.

Standing offer from earlier still applies: push your GGUFs to HF and we'll repoint the row link at the repo.

seanyourhighness:
Twenty minutes and zero local toolchain is better than the recipe deserved — good to know the full-b9967 CPU image bundles the converter, that's worth a line in the README recipe section too.

GGUFs are up: **https://huggingface.co/YourHighnessLA/Tess-4-27B-EAGLE3-GGUF** — all three quants plus the recipe as the model card, with your dual-3090 head-to-head numbers and the #662/#665 links included so the card tells the honest story (accelerates Tess-4, loses to any ~0.70-accept head, kept for the day a stronger head lands). Repoint the row link whenever convenient.

One caveat before "feed mechanism changes nothing" hardens into lore: my rig's n=2 number was **0.491** (SPEED-Bench coding+reasoning, OSL 512, Q4_K_M target) — ~9 points above both your 0.408 and the vLLM ceiling. Same story directionally (this head tops out far below the 0.70 class no matter how you feed it), but the tidy ≈0.40 equality may be partly workload/target-quant flavored rather than a constant of the head. Your aggregate mixes narration in; mine was code+reasoning only, which this head's training mix favors. Worth keeping the claim as "same band, feed-independent" rather than "same number."

The opposite-failure-modes framing (feed-robust-but-capability-limited vs capability-strong-but-alignment-limited) is the crispest one-line summary of this whole thread — if trinity-cloud only reads one sentence of #662, it should be that one.

And +22% narr on the shipped compose from the pin bump alone is a nice bonus to fall out of a provenance question. b9967 ×3 quality rerun still on deck from #665 — same watch list as agreed there.

seanyourhighness:
GGUFs are up on HF: [YourHighnessLA/Tess-4-27B-EAGLE3-GGUF](https://huggingface.co/YourHighnessLA/Tess-4-27B-EAGLE3-GGUF) — BF16, Q8_0, and Q4_K_M all there with the conversion recipe and benchmark numbers in the card.

@trinity-cloud this is your head finding a consumer path — the token-fed external path on a single 4090 is lossless and pays at ~76 code tok/s (1.55× over no-spec), and the acceptance ceiling is confirmed feed-independent so more training data would lift it everywhere at once.

Issue #187, opened 2026-05-22
[bug] Issues getting started with Windows


rhossack:
### What happened?

Here are the commands I ran after installing WSL, pipx, and hf:

./scripts/setup.sh qwen3.6-27b
./scripts/launch.sh --variant llamacpp/mtp
hf download unsloth/Qwen3.6-27B-GGUF Qwen3.6-27B-UD-Q3_K_XL.gguf mmproj-F16.gguf --local-dir /mnt/d/LLMs/qwen3.6-27b-gguf/unsloth-q3kxl
CTX_SIZE=131072 UBATCH_SIZE=1024 docker compose -f models/qwen3.6-27b/llama-cpp/compose/single/mtp.yml up -d
hf download unsloth/Qwen3.6-27B-GGUF Qwen3.6-27B-Q4_K_M.gguf mmproj-F16.gguf --local-dir /mnt/d/LLMs/qwen3.6-27b-gguf/unsloth-mtp-q4km

At this point, I noticed that the model the variant I chose is looking for is not the one installed by the hf, which is why I manually changed the command to get the Q4 instead of the Q3. I then manually moved the mmproj file to the location that was complained about in launch.sh.

Now, when I run launch.sh with the above variant, I see this:

[switch] bringing up: llamacpp/mtp  (models/qwen3.6-27b/llama-cpp/compose/single/mtp.yml)
[+] up 2/2
 ✔ Network single_default         Created                                                                           0.1s
 ✔ Container llama-cpp-qwen36-27b Started                                                                           0.7s
[switch] waiting for http://localhost:8020/v1/models (container=llama-cpp-qwen36-27b, timeout 600s)...
[switch]   60s elapsed, still waiting...

But I'm not able to access that models endpoint in my browser, and the curl test command returns nothing when I run it in another wsl terminal, and the wsl terminal that the model is running under doesn't indicate that there's anything trying to hit it.

I then noticed that the curl test specifies a model different from the one I launched. I'm not sure what exact model string to replace it with, though.

curl -sf http://localhost:8020/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"qwen3.6-27b-autoround","messages":[{"role":"user","content":"Capital of France?"}],"max_tokens":200}'

[my-rig.md](https://github.com/user-attachments/files/28160432/my-rig.md)

### Steps to reproduce

```bash
I put that info in the above box
```

[...]

PR #680, opened 2026-07-11
Bump llama.cpp pin to server-cuda-b9967 (+4 think-ON, validated A/B)

noonghunna:
## Why

Our pin `server-cuda-b9246` was **52 days / 721 builds old** (built 2026-05-20). Validated the current build with a full quality A/B on the Tess dual serve (identical config, only the image changed):

| leg | b9246 (pin) | **b9967** |
|---|---|---|
| think-OFF (greedy) | 115 / 113 | **115** — total tie (4 scenarios flipped each way; engine changes greedy outputs on specific prompts but nets zero) |
| think-ON (+16K budget) | **118 ×3 (tight)** | **122** — incl. 3 previously-100%-reproducible failures now passing: **CLI-31** (safety refusal, was a hard 0), **CLI-20** (byte-exact archive), **RM-04** (the math scenario cross-rig report #665 always passed and we never did) |

Boots clean, external `draft-mtp` works, all 8 packs ran. The think-ON lift also explains most of #665's cross-rig gap (his b9932 sits between our pins).

## What

- All 8 llama.cpp compose image defaults → `server-cuda-b9967`, **unifying a drifted pin landscape**: qwen/tess were b9246, Deckard was b9570, and gemma-12b + vibethinker rode the **rolling tag** — a violation of the #187 policy (rolling crash-looped at b9282; we pin validated builds).
- Engine profile `install.spec` + notes aligned with the composes (spec previously said rolling while composes pinned).
- `test-preflight-compose-deps` fixture, llama-cpp README, WSL_SETUP doc.
- NOT touched: registry status_notes and BENCHMARKS (historical validation records stay dated to their measured pin); `baselines.yml` rows go stale by design — the staleness badge is the feature, and the off-hours re-bench queue picks them up.

## Validation status per slug

Tess dual: full A/B (above). Others: single-boot smoke on the qwen single-card default posting to this PR shortly; remaining slugs boot-revalidate opportunistically — no vendored patches exist on llama.cpp, so risk is flag-compat only (the A/B exercised the full flag family incl. reasoning, draft-mtp, cache-type, ts).

Guards: preflight-compose-deps · status-drift · both parities · registry-disk · mounts-resolve — all PASS.

https://claude.ai/code/session_01EfF565T9eSLaqGzidyJ1Pm


noonghunna:
**Smoke: qwen single-card default family PASS on b9967** — `single/unsloth-q4km/mtp.yml` boots via compose with the new default (`docker inspect` confirms `server-cuda-b9967`), coherence clean ("Paris"/stop), spec-dec path initialized. Combined with the Tess dual full A/B, that's the two highest-traffic llama.cpp shapes (dual layer-split + external draft, single + embedded MTP family) validated on the new pin.

Issue #710, opened 2026-07-14
[bug] MTP n=3 corrupts tool calls on chained-agent workloads (poisons the prefix cache; survives until restart) — parser is NOT the variable, and MTP costs ~44% TPS on 2×3090


alexpolo1:
Opening the separate thread requested in [#178](https://github.com/noonghunna/club-3090/issues/178) ("On the **MTP × qwen3_coder tool-call corruption A/B** — that's the most valuable part, and **yes, please open the separate issue**"). @mgabor3141 reported this mode; this is an independent reproduction on a second rig with a **2×2 A/B that isolates the variable**.

**TL;DR — three findings:**
1. **MTP n=3 corrupts tool calls on chained-agent workloads.** The agent itself poisons the KV/prefix cache: a probe goes **6/6 → 0/6** after 3 autonomous agent runs on the same server.
2. **The tool-call parser is NOT involved.** `qwen3_coder` vs `qwen3_xml` makes **zero** difference (both corrupt with MTP on, both clean with MTP off). So the `qwen3_xml` half of the known workaround is **unnecessary** — MTP-off alone is sufficient.
3. **MTP n=3 is also a large TPS *regression* on this rig: 36.9 → 65.1 tok/s with it off (+79%).** It costs correctness *and* throughput.

---

## Rig

- 2× RTX 3090 (Ampere sm_86), **PCIe-only, no NVLink**, power caps 260/280 W
- `qwen3.6-27b` AutoRound INT4, TP=2, `--kv-cache-dtype fp8_e5m2`, `--enable-prefix-caching`, `--enable-chunked-prefill`, `max-model-len 262144`, `max-num-seqs 4`
- froggeric chat template, `--reasoning-parser qwen3`, `--default-chat-template-kwargs {"enable_thinking": false}`
- vLLM `v0.24.0` (`vllm-stable` pin), compose = **shipped default** `models/qwen3.6-27b/vllm/compose/dual/autoround-int4/fp8-mtp.yml`
- Agent: Nous **hermes-agent** CLI (chained file/terminal tools) against `:8010`

## Symptom

The model stops emitting the template's XML tool format and instead emits unparseable markup, so vLLM returns **no `tool_calls`** and the agent silently ends the task having done nothing (`exit 0`):

```
<arg_key>thinking</arg_key><arg_value>Let me start by reading the log file to understand the symptom...</arg_value>
```
```
search_files(target='files', path='/tmp', pattern='*')      # naked call, no XML wrapper
```
[...]

PR #735, opened 2026-07-17
Add vllm/tess-dual-w4a16: the MTP-revival slug (262K, MTP n=5, #662)

noonghunna:
## What

Promotes `LeaderboardModel1/Tess-4-27B-AutoRound-W4A16-Tuning` into the catalog as **`vllm/tess-dual-w4a16`** — dual 3090, TP=2, **262K ctx**, fp8_e4m3 KV, **built-in MTP n=5** (`SPEC_N` overridable). This is the #662 resolution: the checkpoint keeps `mtp.fc` + `linear_attn.in_proj` BF16 via `extra_config`, reviving the MTP head that the migtissera NVFP4 export killed (4-bit'd `mtp.fc` → 0% accept). klement's hypothesis confirmed — the export was the killer, not head misalignment.

## Validation (reference 2×3090, rebench-full 131 min)

| Gate | Result |
|---|---|
| bench | **73.5 narr / 108.2 code** tok/s decode (NVFP4 spec-off: 62.4) — MTP accept-len 5.0 / 80% warm |
| verify-stress | **8/8** incl. NIAH ceiling ladder to 0.92×262K (2.2 GB VRAM margin) |
| soak | PASS — 0/100 silent-empty, p50 110.6 tok/s, 105% retention |
| 8-pack | **108/150 off / 115/150 on** — beats the NVFP4 A0 (106/113) on both legs |
| compose boot | via `switch.sh` (not hand-run): verify-full **9/9**, MTP contributing |
| catalog suite | full sweep **83/83 green**; `diagnose-profile` GREEN |

MTP n-sweep (single-variable boots): knee at **n=5** — prose break-even (69.9 vs 70.1 no-spec), code **+49%** (104.3 vs 70.0); n=6/8 regress as tail acceptance collapses. Artifacts: `results/rebench/tess-4-27b-20260717-0529`.

Status ships **⚠️ caveats**: think-OFF still undershoots the GGUF tier (108 vs 116, cli-40-concentrated); think-ON ties (115/117, cli-40 25/40 = parity) — the known 4-bit precision gap from the 07-12 arms, not a serving-path issue.

## Changes

- **New compose** `models/tess-4-27b/vllm/compose/dual/leaderboard-w4a16/fp8-mtp.yml` + registry entry + `leaderboard-w4a16` weights variant + `tess-mtp-builtin` drafter profile (c3 spec column).
- **⚠️ DEFAULTS decision to review:** adds `(tess-4-27b, vllm, dual)`. Because `vllm` leads `ENGINE_PREFERENCE[dual]` and `caveats` is functional, this **flips the curated `tess-4-27b/default` from llamacpp to vllm** (73/108 TPS vs 52–63/67; think-ON ties, think-OFF −8). The DEFAULTS comment documents the one-line revert if the GGUF should stay the model default.
- **compat.py C5**: `qwen35` family added to the Ampere fp8_e4m3 FlashInfer allowance (same DeltaNet-hybrid class as qwen3-next, live-validated today). Gemma stays correctly rejected; guard fixtures unchanged.
- **NVFP4 sibling** header: drafter forensics re-attributed to the export (the "feature-space misaligned" claim is retracted); slug repositioned for native-FP4 rigs.
- **UPSTREAM.md**: new **vllm#40898** row — stock-vLLM DFlash hard-`NotImplementedError`s on hybrid `layer_types` (verified on v0.25.1 during this work); the retired 82/125 DFlash numbers were purged-nightly overlays, not stock.
- **BENCHMARKS.md**: full row (n-sweep, gates, DFlash rule-outs — including the cross-model Anbeeld drafter losing to Tess's own MTP GGUF on the llama.cpp lane by ~20pp acceptance).
- Test count bumps: registry 66, disk composes 67, drafters 14.

https://claude.ai/code/session_01EfF565T9eSLaqGzidyJ1Pm


kevinb361:
Field report from an attended dual-3090 canary of the merged Tess recipe.

Exact model: `LeaderboardModel1/Tess-4-27B-AutoRound-W4A16-Tuning` at revision `4004cd9a5e1ac009ce31358e1fb77b99a58dfc68`.

The recipe worked technically: the model loaded, MTP n=5 engaged, plain and native-tool smokes passed, final-state checks passed, and there were no OOMs or unexpected restarts. MTP acceptance measured 49.1% (2,188/4,460 accepted draft tokens).

The attended comparison did not justify advancement over the existing Qwen control:

| Measure | Qwen control | Tess MTP5/APC-off |
|---|---:|---:|
| Code decode | 91.32 tok/s | 52.39 tok/s |
| Narrative decode | 80.93 tok/s | 40.02 tok/s |
| Repeated tool-loop median | 2.749 s | 14.398 s |
| MTP acceptance | 61.5% | 49.1% |
| Safety/minimality lexical pass | 0.733 | 0.667 |

An external pairwise judge liked Tess on the general corpus (13 wins, 13 ties, 4 losses), but it lost the safety corpus gate (5 wins, 2 ties, 8 losses; mean 0.433). The repeated-turn penalty was 5.24x in the shipped APC-off shape.

I stopped there under the preregistered fail-closed sequence, so the conditional spec-off and APC-on arms were not used to rescue the candidate. This is not a claim that Tess is generally bad; it is a field report that the exact dual-3090 MTP5/APC-off recipe works but was not competitive for an agentic deep lane where repeated tool turns and safety/minimality matter.

noonghunna:
@kevinb361 — a preregistered fail-closed canary with an attended pairwise judge is the reference standard for field reports here; thank you. Three responses, one of which materially changes the picture:

**1. Your worst finding was measured on a shape that was superseded hours later.** The canary ran the #735 ship (APC-off — the #720 interim default). Later the same day, **#736 flipped prefix-caching back ON as the shipped default** — with the vllm#48375 fix vendored and boot-applied on this exact compose — largely *because of your own production data* (the 89.5% hit rate / 7.4× TTFT numbers you posted on #710). The 5.24× repeated-tool-loop penalty you measured is the APC-off re-prefill cost at 262K, and it's precisely what the flip removes: our warm-TTFT matrix on the qwen sibling shows 13.3s → 1.3s on held-prefix turns with APC on. Your prereg reserved an APC-on arm as a "rescue" you rightly refused to use — but as of #736 that arm *is the shipped default*, so a re-run of just the repeated-tool-loop probe on current master would be measuring the recipe as shipped, not a rescue. If it still loses there, that's a finding we'd act on.

**2. The decode gap is the accept-rate collapse on your workload — and the recipe has a knob for it.** You measured 49.1% accept on attended agent traffic vs the 80% we measured warm on code generation — MTP acceptance is strongly workload-shaped, and at ~49% accept, **n=5 is past its knee and becomes an anchor** (our n-sweep: the tail positions only pay for themselves above ~55–60%). That's most of your 52/40 tok/s vs our 108/73. For an agentic deep lane, try `SPEC_N=3` (the compose seam) or comment the spec block out entirely — at your acceptance profile, spec-off may genuinely win on short tool turns, consistent with the two-regime split from #710.

**3. The safety-corpus loss is the model, not the recipe — and your gate is reading it correctly.** Tess is a Hermes-lineage agentic fine-tune: deliberately low-refusal, compliance-optimized. Our own quality history says the same thing — the cli-40 safety-refusal scenario has been a chronic Tess failure across every quant and both engines (one earlier trace executed a destructive command outright; a later apparent refusal turned out to be engine-side, not model-intrinsic). No quant, template, or serving recipe moves this; it's equally true of the GGUF lane. For a lane where safety/minimality gates matter, the Qwen control is the right pick and your fail-closed result is the catalog working as intended — the slug ships ⚠️ with the agentic undershoot documented, and the recipe's honest win is speed + 262K + the revived head for code-gen-shaped work.

We'll fold this report into the BENCHMARKS Tess section as the first cross-rig W4A16 row (attended-canary caveat noted). If you do re-run the tool-loop probe on the post-#736 default, that number lands there too — either way it sharpens the lane guidance.

Issue #724, opened 2026-07-16
c3 catalog: column show/hide + reorder, persisted across restarts


noonghunna:
## What

The Run-catalog table is now 14 columns (#723) — narrow terminals and per-user workflow differ, so make the column set user-configurable:

- **`[|]` hotkey** (and **clicking the table header**) opens a columns picker modal
- picker: `space` show/hide · `shift+↑/↓` reorder · `r` reset to canonical · `⏎` apply · `esc` cancel
- `slug` is pinned (identity — never hideable)
- **persisted across restarts** via `c3-settings.json` (`catalog_columns: {order, hidden}`), applied at launch through `apply_persisted_settings` (same hermetic hook as `[S]` settings, so directly-constructed apps — tests — always start canonical)
- forward-compat: when a future release ships a NEW column, a saved layout gets it inserted at its canonical position, visible — no wiped layouts

## Design notes

- Canonical column set becomes a single module-level spec (`_CATALOG_COLUMNS`) — one source for headers, row cells, and the picker.
- The modal never touches the pane (repo modal rule): it dismisses `{order, hidden}` and `CatalogPane.set_columns` applies + persists.
- `c` and `C` are taken (power-cap / lean-view); `|` reads as column bars and sits on shift+\\ next to the `[\\]` model toggle.

Follow-on to #723.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #723, opened 2026-07-16
c3 catalog: add provider, weight-size, and activations columns


noonghunna:
## What

Three new columns in the c3 (serve-cockpit) Run-catalog table, placed:

| new column | position | shows |
|---|---|---|
| **provider** | before `weights` | the weights author/org (Lorbus, cyankiwi, Avesed, nvidia, ubergarm, …) |
| **GB** (weight size) | before `kv` | the artifact size in GB |
| **act** (activations) | before `spec` | the activation compute format (`16bit` today; `int8` for W4A8-class configs) |

New column order: `model · slug · provider · weights · GB · kv · act · spec · ctx · TPS (rig) · 8pk (rig) · topo · engine · status`.

## Why

- **provider**: quant provenance matters when picking a slug (recipe quality varies by author — see the NVFP4 recipe findings in #662 and the W4A8 scale-sign findings in #609), and today it's only discoverable by opening the model YAML.
- **weight size**: the download/VRAM planning number, already shown in the download flow but not at pick time.
- **activations**: the W4A16 / W4A8 / W8A8 axis became real on this stack (#609 — +52% prefill on dual-fast with int8 activations). The catalog needs to show which configs run 16-bit activations vs int8 once the W4A8 knob ships; today every slug is `16bit`, so the column starts uniform by design.

## How (traced)

- **provider + GB need no contract change**: `CatalogEntry.weights` (a `WeightsMeta`) is already joined at catalog build from `weights.py list --json` and carries `hf_repo` + `size_gb`. Provider = the org segment of `hf_repo`; size = `size_gb`. Render-only.
- **activations is a new registry facet**, following the exact `kv_format` pattern: `_entry(..., act_format="16bit")` default in `compose_registry.py` → emitted in `registry-emit.sh --json` variants → attached in `services.py::_variant_row_from_dict` via `object.__setattr__` → `_act_label()` in `app.py`. No per-entry edits needed (default covers the whole catalog); future W4A8 slugs set `act_format="int8"`.

Headers kept short (`provider`/`GB`/`act`) for narrow-terminal column budget; status stays last (emoji-width alignment rule).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #757, opened 2026-07-24
c3: add catalog "offload" column (weight-offload backend facet)

noonghunna:
## What

Adds a catalog **`offload`** column to c3 — the weight-offload backend a slug uses to serve a model that doesn't fit VRAM. Mirrors the `act_format` / #723-#724 facet pattern exactly.

Vocabulary:
- **`None`** — fully resident (the default; **every existing slug** → column shows `—`)
- **`uva`** — vLLM zero-copy demand-paged expert offload (GPU computes, routed experts stream over PCIe)
- **`n-cpu-moe`** — llama.cpp CPU-computed expert offload (weights stay in RAM, only activations cross PCIe)
- **`prefetch`** — vLLM bulk layer prefetch

## Why

The catalog had **no dimension to represent weight offload**. It's needed by the first offloaded model on the stack — Poolside Laguna S 2.1 (118B MoE), served via vLLM UVA expert offload on 2×3090 — where the offload backend is a load-bearing serving fact (UVA vs prefetch is a ~25× decode difference). The existing LMCache KV-offload slug can adopt the facet retroactively.

## Change (5 additive touch-points)

| File | Change |
|---|---|
| `compose_registry.py` `_entry` | new `offload=None` field (+ facet comment) |
| `registry-emit.sh --json` | emit `offload` (mirrors `act_format`) |
| c3 `services.py` | thread `offload` onto the row |
| c3 `app.py` | `_offload_label()` + `("offload","offload")` in `_CATALOG_COLUMNS` + row cell |

Every existing registry entry defaults to `offload=None`, so all 68 current slugs render `—` and nothing else moves. Backward-compatible with older emit (`""` → `—`).

## Tests

- Registry imports clean; all 68 entries carry the key.
- Emit-parity: `test-switch-registry-parity`, `test-launch-registry-parity`, `test-registry-emit-no-yaml`, `test-compose-status-drift` — green.
- c3 pytest: **248** `test_services` + `test_registry_parser`, **81** catalog/app-headless (incl. full app boot) — green.

## Not in this PR

The Laguna catalog entry itself is **held** (drafted, staged out-of-repo) pending the `67dbeda4` re-quant re-measure and the llama.cpp `--n-cpu-moe` head-to-head that decides the engine default. This PR is just the enabling column.

### DIFF
--- a/scripts/lib/profiles/compose_registry.py
+++ b/scripts/lib/profiles/compose_registry.py
@@ -60,6 +60,14 @@ def _entry(
     # its weights are positive-symmetric int4 (the c3 serve-confirm checkbox reads
     # this to offer VLLM_MARLIN_INPUT_DTYPE=int8 per-launch). #609.
     act8_capable=False,
+    # Weight-offload backend when the slug serves a model too large to fit VRAM
+    # by paging expert/layer weights to host RAM. None = fully resident (the
+    # default — every existing slug). "uva" = vLLM zero-copy demand-paged expert
+    # offload (GPU computes, weights stream over PCIe); "n-cpu-moe" = llama.cpp
+    # CPU-computed expert offload (weights stay in RAM, only activations cross
+    # PCIe); "prefetch" = vLLM bulk layer prefetch. Surfaced as the c3 catalog
+    # "offload" column. First used by the Laguna 118B-MoE offload slugs.
+    offload=None,
     chat_template="native",
     tp,
     max_ctx,
@@ -92,6 +100,7 @@ def _entry(
         "kv_format": kv_format,
         "act_format": act_format,
         "act8_capable": act8_capable,
+        "offload": offload,
         "chat_template": chat_template,
         "tp": tp,
         "pp": 1,
--- a/scripts/lib/registry-emit.sh
+++ b/scripts/lib/registry-emit.sh
@@ -670,6 +670,9 @@ for vr in _tui_registry.parse_variant_rows(tab):
             # W4A8-int8-activation capability (c3 serve-confirm checkbox, #609) —
             # True when the compose is wired + weights are positive-sym int4.
             "act8_capable": bool((COMPOSE_REGISTRY.get(d["slug"], {}) or {}).get("act8_capable")),
+            # Weight-offload backend (catalog "offload" column) — None = resident
+            # (default) / "uva" / "n-cpu-moe" / "prefetch". Laguna 118B-MoE slugs.
+            "offload": (COMPOSE_REGISTRY.get(d["slug"], {}) or {}).get("offload"),
             # Weights quant_label + FORMAT from the model profile (catalog
             # Weights column fallbacks) — see _weights_meta() above.
             "weights_quant_label": _weights_meta(
--- a/scripts/tests/test-registry-json.sh
+++ b/scripts/tests/test-registry-json.sh
@@ -74,6 +74,9 @@ VARIANT_KEYS = {
     "chat_template",
     # c3 serve-confirm W4A8 checkbox capability (#609).
     "act8_capable",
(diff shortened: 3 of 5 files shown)


noonghunna:
Reviewed. **Found and fixed one blocking regression** (pushed as `10f672c`); the column itself is sound. Two staleness notes below that don't block the merge but should be settled before the first offload slug lands.

## Blocking regression — `test-registry-json` (fixed)

`test-registry-json.sh` asserts an **exact** key set for the emitted variant JSON:

```
ASSERTION FAILED: variant keys mismatch (got [... 'offload' ...])
```

Adding `offload` to the emit without adding it to `VARIANT_KEYS` fails that test. Confirmed it's a genuine regression, not pre-existing: **PASS on pristine `origin/master`, FAIL on this branch.** It isn't in the PR description's tested list, which is presumably how it slipped — the four guards listed (`test-switch-registry-parity`, `test-launch-registry-parity`, `test-registry-emit-no-yaml`, `test-compose-status-drift`) all pass, so the emit change looked fully covered.

Worth noting for the next facet: `test-registry-json` is the only guard that pins the emit's key set exactly, so **every** `--json` contract addition must touch it. The AGENTS.md c3 note ("emit changes also need the `test-switch-registry-parity` / `test-launch-registry-parity` guards green") lists the parity guards but not this one — that's an accurate-but-incomplete instruction that will keep catching people.

## Verification after the fix

| Check | Result |
|---|---|
| Full `scripts/tests` suite (branch + master) | **84 / 0** |
| c3 `test_services` + `test_registry_parser` | **248 passed** |
| c3 `test_app_headless` (full app boot) | **568 passed** (13 min) |
| Registry import, all entries carry the key | 68 entries, 0 missing, distribution `{None: 68}` |
| Merges onto current master | clean |

So the "every existing slug renders `—` and nothing else moves" claim checks out literally — all 68 entries are `None`.

## Note 1 — `n-cpu-moe` names a flag we deliberately rejected

The vocabulary defines `n-cpu-moe` as "llama.cpp CPU-computed expert offload". But the Laguna llama.cpp config we actually settled on does **not** use `--n-cpu-moe` — it uses interleaved `-ot` regex overrides, *because* `--n-cpu-moe` generates contiguous layer overrides and therefore cannot balance two GPUs (it left one card at 3.2 GB and the other at 12 GB; `-ot` got them to 22.7 / 21.8). `--n-cpu-moe` is on the list of llama.cpp defaults that were actively wrong for this workload.

So a slug labelled `n-cpu-moe` would name the mechanism we rejected. Suggest either a flag-agnostic value (`cpu-moe`, which honestly covers both `--n-cpu-moe` and `-ot`) or an explicit `tensor-override` value, so the facet describes the offload *strategy* rather than a CLI flag we don't pass.

## Note 2 — the "Why" rationale predates the llama.cpp arm winning

The PR says Laguna is "served via vLLM UVA expert offload on 2×3090", and the held slug is `vllm/laguna-dual-uva`. That was true when this was opened, but the llama.cpp arm now **beats** vLLM UVA on decode — 31.5 TPS vs 25.4 — so the first offloaded slug may well be a llama.cpp one, and `uva` may not be the value it carries.

Doesn't affect the column (it's engine-agnostic and defaults to `None`), but the held entry's name and the PR narrative should be re-derived from the final numbers rather than the state at authoring time. Both arms are still gated on the `67dbeda4` re-measure, so nothing needs deciding yet.

## Verdict

**Mergeable now** — `MERGEABLE/CLEAN`, +28/−0, additive, fully backward-compatible with older emit (`""` → `—`). The vocabulary question is worth resolving before a slug actually adopts a non-`None` value, but it costs nothing to leave the enabling column in place meanwhile, since no entry uses it yet.

PR #109, opened 2026-05-09
fix: BIND_HOST opt-in + localhost script fixes

easel:
## Summary

Adds `BIND_HOST` env var to all 18 vLLM compose files so the API port can be restricted to localhost, and fixes two scripts that broke when `BIND_HOST=127.0.0.1` was set.

**Motivation**: When running `max-num-seqs=1` variants (bounded-thinking, default, long-*, tools-text), any IDE agent (Cline, Cursor, Copilot) that probes the same port will preempt an in-flight verify-stress or soak run, causing spurious HTTP 000 / 429 failures. Binding to 127.0.0.1 prevents that without requiring firewall rules.

**Changes**:
- All 18 `docker-compose.*.yml` files: port binding changed from `"${PORT:-NNNN}:8000"` to `"${BIND_HOST:-0.0.0.0}:${PORT:-NNNN}:8000"` — default unchanged, opt-in via `.env`
- `scripts/preflight.sh`: port-detection regex broadened from `0.0.0.0:PORT->` to `[any IPv4]:PORT->` — was silently producing empty output (exit 1) when `BIND_HOST=127.0.0.1`
- `scripts/report.sh`: forward `SOAK_TIMEOUT_S` to `soak-test.sh` invocation — was silently using the 1800s default even when `.env` set `SOAK_TIMEOUT_S=3600`
- `.env.example`: docs for `BIND_HOST` and `SOAK_TIMEOUT_S`

## Test plan

- [ ] `docker compose up -d` with no `.env` — port binding unchanged (`0.0.0.0:8020->8000/tcp`)
- [ ] `BIND_HOST=127.0.0.1 docker compose up -d` — port binds to `127.0.0.1:8020->8000/tcp`; `preflight.sh` detects port correctly; `bash scripts/report.sh --full` runs to completion with correct soak timeout

Tested on RTX 5090 Laptop + WSL2 (driver 596.36) with bounded-thinking.yml + `BIND_HOST=127.0.0.1` + `SOAK_TIMEOUT_S=3600` — all 10 verify-full checks pass, soak 5/5 PASS.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #764, opened 2026-07-25
Route every compose port through ${BIND_HOST:-0.0.0.0} + guard test

### LINKED ISSUE #763: BIND_HOST knob missing from 46 of 86 composes (documented contract, not followed)
## The gap

`scripts/lib/generate_compose.py` emits port mappings as:

```yaml
    ports:
      - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-<port>}}:8000"
```

and `docs/ADDING_MODELS.md` documents that exact shape as the compose contract. But **46 of the 86 composes under `models/` don't follow it** — they publish bare:

```yaml
    ports:
      - "${ESTATE_PORT:-${PORT:-8060}}:8080"
```

The split is almost perfectly along authorship lines: every compose carrying the knob is a vLLM one (`:8000`), and the ones missing it are the hand-written llama.cpp / ik-llama / beellama / SGLang composes (`:8080`, `:30000`) plus a few vLLM Gemma ones. It reads like drift from composes written outside the generator, not a deliberate distinction.

## Why it matters

None of the engines we ship have auth. On a compose without the knob there's no way to narrow the bind short of editing the file — and because that edit is to a tracked file, `git checkout <tag>` silently reverts it on the next update. Concretely, on my own fleet: a box serving `beellama/dflash` got re-exposed on `0.0.0.0` on a LAN by a routine tag update, because the loopback bind only existed as a local commit that the checkout detached away from. That's a footgun that a one-line default-preserving change removes.

Operators on a trusted LAN are unaffected either way — the default stays `0.0.0.0`.

## Proposed fix

1. Prefix the host side of every published port under `models/` (excluding `_archive/`, matching `test-compose-gpu-mask-passthrough`'s convention) with `${BIND_HOST:-0.0.0.0}`. **Default unchanged**, so no behaviour change for anyone who doesn't set the variable.
2. Add `test-compose-bind-host.sh` so a new compose copied from an old one reds instead of silently shipping an unnarrowable bind. It also asserts `generate_compose.py` still emits the prefix.
3. Expand the `.env.example` note — the knob currently reads as a benchmarking aid (avoiding IDE-agent contention); worth stating the security reason too, now that it applies everywhere.

Happy to reshape if you'd rather scope it differently — e.g. live composes only, skip the guard test, or split the engine families into separate PRs. PR follows, referencing this issue.

datanerdie:
## Summary

Closes #763. `generate_compose.py` emits `${BIND_HOST:-0.0.0.0}` on the host side of every port mapping and `docs/ADDING_MODELS.md` documents that as the compose contract — but **46 of the 86 composes under `models/` don't follow it**, publishing a bare `"${ESTATE_PORT:-${PORT:-N}}:8080"` instead. The split falls along authorship lines: everything with the knob is a generator-shaped vLLM compose (`:8000`); everything without it is hand-written llama.cpp / ik-llama / beellama / SGLang (`:8080`, `:30000`) plus a few vLLM Gemma composes. This brings all of them onto the documented contract and adds a guard so they can't drift off it again.

Since no engine we ship has auth, a compose without the knob gives the operator no way to narrow the bind except editing a tracked file — which `git checkout <tag>` then silently reverts on the next update. That happened on my own fleet: a box serving `beellama/dflash` was re-exposed on `0.0.0.0` on a LAN by a routine tag update, because its loopback bind only existed as a local commit the checkout detached away from.

**The default does not change.** It stays `0.0.0.0`, so this is behaviour-preserving for everyone who doesn't set `BIND_HOST` — it only gives `BIND_HOST=127.0.0.1` in `.env` something to bind on the other 46.

## Type of change

- [x] Other — cross-cutting compose consistency + a guard test. No serving flags, images, weights, ports, or defaults are touched.

## What's in it

1. **46 composes** get the `${BIND_HOST:-0.0.0.0}:` prefix on the host side. One line each, no other edits. `_archive/` is left alone, matching `test-compose-gpu-mask-passthrough`'s convention.
2. **`scripts/tests/test-compose-bind-host.sh`** — asserts every live compose publishing a port routes it through `BIND_HOST`, and that `generate_compose.py` still emits the prefix. Verified it bites: reverting the compose commit makes it fail with all 46 named.
3. **`.env.example`** — the knob currently reads as a benchmarking aid (avoiding IDE-agent contention for the `max-num-seqs=1` slot). Added the security rationale and noted it now applies to every compose.

## Verification

No rig report or bench: this change cannot move TPS or alter serving behaviour — the container-side port, all flags, and the `0.0.0.0` default are untouched, and with `BIND_HOST` unset the rendered compose is byte-identical to before. So the meaningful verification is the test suite, run on Linux (2× 3090, Ubuntu, bash 5.2):

```
with this PR:            PASS=83  FAIL=1
pristine upstream master: PASS=82 FAIL=1
```

Same single failure on both — `test-submit-bench.sh` (`expected at least 6 submit-bench fixtures, found 0`), pre-existing and unrelated. The `+1` is the new guard.

All YAML re-parsed cleanly (90 files). Spot-checked the rendered mapping across all three shapes:

```
vLLM     - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-8032}}:8000"
llama.cpp- "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-8060}}:8080"
SGLang   - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-8041}:30000"
```

## Notes for review

- Happy to reshape: live-composes-only, drop the guard, or split by engine family if a 46-file diff is unwelcome as one unit. It's mechanical and regenerable either way.
- I left `_archive/` untouched for consistency with the existing GPU-mask guard, but can include it if you'd rather the invariant be exceptionless.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -175,9 +175,14 @@
 # PORT=9876
 
 # Host interface to bind the API port on. Default is 0.0.0.0 (all interfaces).
-# Set to 127.0.0.1 to restrict to localhost — prevents IDE agents (Cline,
-# Cursor) from competing for the max-num-seqs=1 slot and causing verify-stress
-# HTTP 000 failures during benchmark runs.
+# Honored by every compose under models/ (enforced by test-compose-bind-host).
+#
+# Set to 127.0.0.1 to restrict to localhost. Two reasons you might:
+#   - Security: none of the served engines have auth, so on 0.0.0.0 anyone who
+#     can reach the host on that port gets an unauthenticated API. If the box
+#     isn't on a trusted network, bind loopback and reach it over an SSH tunnel.
+#   - Benchmarking: keeps IDE agents (Cline, Cursor) from competing for the
+#     max-num-seqs=1 slot and causing verify-stress HTTP 000 failures.
 # BIND_HOST=127.0.0.1
 
 # Docker restart policy for launched model containers. Default: unless-stopped
--- a/models/diffusiongemma-26b-a4b/vllm/compose/dual/fp8/base.yml
+++ b/models/diffusiongemma-26b-a4b/vllm/compose/dual/fp8/base.yml
@@ -62,7 +62,7 @@ services:
     container_name: "${ESTATE_CONTAINER:-vllm-diffusiongemma-26b-a4b-fp8-tp2}"
     restart: ${CLUB3090_RESTART:-unless-stopped}
     ports:
-      - "${ESTATE_PORT:-${PORT:-8042}}:8000"
+      - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-8042}}:8000"
     volumes:
       - ${MODEL_DIR:-../../../../../../models-cache}:/root/.cache/huggingface
       # Google Gemma 4 CANONICAL chat template (published 2026-07-09) — SHARED
--- a/models/gemma-4-12b/beellama/compose/single/beellama-q8kxl/base.yml
+++ b/models/gemma-4-12b/beellama/compose/single/beellama-q8kxl/base.yml
@@ -65,7 +65,7 @@ services:
     container_name: "${ESTATE_CONTAINER:-beellama-gemma4-12b}"
     restart: unless-stopped
     ports:
-      - "${ESTATE_PORT:-${PORT:-8067}}:8080"
+      - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-8067}}:8080"
     volumes:
       - "${MODEL_DIR:-../../../../../../models-cache}:/models:ro"
     # server target ENTRYPOINT is /app/llama-server — args only below.
--- a/models/gemma-4-12b/llama-cpp/compose/single/unsloth-q8kxl/base.yml
+++ b/models/gemma-4-12b/llama-cpp/compose/single/unsloth-q8kxl/base.yml
@@ -45,7 +45,7 @@ services:
     container_name: "${ESTATE_CONTAINER:-llama-cpp-gemma4-12b}"
     restart: unless-stopped
     ports:
-      - "${ESTATE_PORT:-${PORT:-8069}}:8080"
+      - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-8069}}:8080"
     volumes:
       - "${MODEL_DIR:-../../../../../../models-cache}:/models:ro"
     command: >-
--- a/models/gemma-4-12b/vllm/compose/dual/bf16/mtp.yml
+++ b/models/gemma-4-12b/vllm/compose/dual/bf16/mtp.yml
@@ -45,7 +45,7 @@ services:
     container_name: "${ESTATE_CONTAINER:-vllm-gemma-4-12b-dual-mtp}"
     restart: ${CLUB3090_RESTART:-unless-stopped}
     ports:
-      - "${ESTATE_PORT:-${PORT:-8036}}:8000"
+      - "${BIND_HOST:-0.0.0.0}:${ESTATE_PORT:-${PORT:-8036}}:8000"
     volumes:
       - ${MODEL_DIR:-../../../../../../models-cache}:/root/.cache/huggingface
       # Google Gemma 4 CANONICAL chat template (published 2026-07-09) — SHARED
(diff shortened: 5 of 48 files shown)


noonghunna:
Reviewed and verified independently. **This is ready to merge as-is** — don't reshape it. Details below, plus confirmation that your one test failure is ours, not yours.

## Verification

Reviewed in a clean worktree off your branch, not my working tree, so local state couldn't mask anything.

**1. The diff is purely mechanical.** I asserted every changed mapping equals master's string with exactly `${BIND_HOST:-0.0.0.0}:` prepended and nothing else:

```
composes changed: 46
port mappings that are EXACTLY master + '${BIND_HOST:-0.0.0.0}:': 46
anything else: 0
```

Zero non-port lines changed under `models/`. So the behaviour-preserving claim holds structurally, not just by inspection — with `BIND_HOST` unset the host side renders `0.0.0.0`, which is what Docker already does for a bare host port.

**2. YAML validity:** 46/46 parse cleanly.

**3. The guard bites.** Reverted one compose (`gemma-4-31b/vllm/compose/dual/qat-w4a16/int8.yml`) to its master form and re-ran:

```
FAIL: models/gemma-4-31b/vllm/compose/dual/qat-w4a16/int8.yml publishes a port without a BIND_HOST knob:
      expected the host side to route through ${BIND_HOST:-0.0.0.0}
1 BIND_HOST check(s) failed
```

Names the offending file, which is what makes it useful to whoever trips it. And I like that it also asserts `generate_compose.py` still emits the prefix (L49-50) — that closes the loop where the generator drifts and the composes silently follow.

**4. Full sweep:** `PASS=83 FAIL=1`, identical to your numbers.

## Your `test-submit-bench` failure is our bug, and you found it

You called it "pre-existing and unrelated" and were right on both counts — but it's worth stating exactly why, because it affects every contributor:

`bench_row_fixtures()` in `scripts/lib/bench-row-formatter.sh` looks for six **`results/rebench/<tag>` directories**:

```
qwen-int8-pth-n4-2026-05-10, qwen-bf16-n4-2026-05-11, qwen-int8-tq3-n3-2026-05-11,
qwen-tq3-mtp-genesis-2026-05-11, gemma-int8-pth-n4-2026-05-11, gemma-bf16-n4-2026-05-11
```

Those are **local benchmark output directories, not committed fixtures** — `git ls-files results/rebench` returns **0 tracked files**. They exist on the maintainer rig only because we've run those benchmarks. So the test passes here and **cannot pass on any fresh clone**.

I confirmed the mechanism while reviewing you: in my normal working tree the suite is 83/0; in the clean worktree I made for this review it's 83/1 with exactly your failure. Same commit, different local artifacts.

That's a genuine defect — a test that silently depends on untracked local state means **the suite is red for every external contributor**, and "is the suite green?" is unanswerable for anyone outside this rig. It should skip cleanly when the fixtures are absent, or the fixtures should be committed. Filing that separately; it is not yours to fix and should not hold this PR.

## On your three offered reshapes — no to all three

- **Split by engine family?** No. The value is the invariant holding *everywhere*; a partial rollout recreates exactly the drift you're fixing, and the guard would have to be disabled until the last part landed.
- **Drop the guard?** Definitely not — the guard is the most valuable part. The composes are a one-time fix; the test is what stops the next hand-written compose from silently shipping an unnarrowable bind. This is the second time `BIND_HOST` has needed attention (#109 introduced it as opt-in and the rollout was left partial), which is precisely the argument for enforcing it mechanically.
- **Include `_archive/`?** Leave it. Matching `test-compose-gpu-mask-passthrough`'s existing convention is the right call — archived composes aren't launchable, and an exceptionless invariant that covers dead files buys nothing while making future archiving noisier.

## One note for the changelog

Worth being precise that this is **not** a security fix in the sense of changing a default — the default stays `0.0.0.0` and anyone not setting `BIND_HOST` is unaffected. What it fixes is **reachability of the knob**: on 46 composes there was no supported way to narrow the bind, and the unsupported way (editing a tracked file) is silently reverted by `git checkout <tag>`. Your beellama LAN re-exposure is the concrete failure mode, and it's the kind of thing that only shows up once someone runs a fleet through a tag update — thank you for reporting it with the mechanism rather than just the symptom.

Nice, disciplined PR: one-line-per-file, default preserved, guard included, and honest verification including a failure you didn't cause.

noonghunna (approved):
Verified independently in a clean worktree: 46/46 mappings are exactly master + `${BIND_HOST:-0.0.0.0}:` with zero non-port lines changed, 46/46 YAML parse cleanly, the guard bites by filename when reverted, and it also pins `generate_compose.py`. Full sweep matches your numbers (83 pass / 1 fail), and that one failure is our pre-existing `test-submit-bench` fixture bug — reproduced it on your commit in a clean worktree vs 83/0 in a tree that happens to have untracked `results/rebench/` dirs. Filing that separately.

Merging as-is — no reshape. Thanks for the guard; that's the part that stops this drifting a third time.

PR #801, opened 2026-07-27
c3: add opt-in master logging

### LINKED ISSUE #795: c3: master logging setting ([S], default OFF) — app log + all invoked-script logs
### Motivation

The 2026-07-27 download triage (#792 / #793 / #794) started with a user report that ended in "i did not found any logs for c3 🙁" — and they were right. #793 fixed exactly one slice (downloads now tee to `~/.config/club-3090/logs/`, pane-announced), but everything else in c3 still evaporates with the pane: app exceptions, action dispatch errors, serve/switch/verify/bench subprocess output, GPU-claim decisions. The next non-download incident will be the same remote guessing game.

### Proposal

A **master logging setting** in c3:

- **Toggle in [S] Settings**, persisted to `c3-settings.json` (same mechanism as MODEL_DIR / HF_TOKEN), plus an env override (`C3_LOG=1/0`) that wins per-launch, matching the settings precedence rule.
- **Default OFF** (maintainer decision).
- When ON, capture to `<C3_CONFIG_DIR|~/.config/club-3090>/logs/`:
  1. **c3 app log** — one session file (`c3-session-<ts>.log`): Python `logging` from the app (screen/action transitions at INFO, warnings, and **unhandled exceptions with tracebacks** — the crash case is the highest-value line), Textual log integration where useful.
  2. **Subprocess logs** — generalize #793's `DownloadLog` to every runner invocation:
     - write/stream runner (`SubprocessRunner.start_raw`): full line stream + argv header + exit/verdict footer (serve, switch, verify, bench, gpu-mode, downloads — the download path already does this).
     - read runner (`RealRunner.run`): argv + exit code always; stdout/stderr **only on failure or non-parseable output** (registry-emit/kv-calc JSON blobs are huge and useless when healthy — logging them would drown the signal).

### Design points / open questions

- **Relationship to #793's download logs**: they shipped **always-on** because the motivating incident was precisely "nothing to paste." Folding them under a default-OFF master switch regresses that win. Options: (a) downloads stay always-on, the switch governs everything else; (b) everything under the switch, default OFF, and the download preflight prints "enable logging in [S] for a persistent trace" as a breadcrumb; (c) switch default ON for downloads only via a tri-state. Maintainer call — the ticket should not ship without an explicit decision here.
- **Redaction**: existing rule holds everywhere — argv yes, **env never** (HF_TOKEN). Session log must also avoid echoing settings values (the token field).
- **Retention**: reuse the newest-N prune (downloads keep 30); session logs probably fewer (10?).
- **Announcement**: when ON, the footer / [S] modal shows the active log path so discoverability matches the download logs' `[log]` line.
- **Hermeticity**: suite-wide `C3_CONFIG_DIR` tmp + default-OFF means tests stay clean; add explicit-enable tests mirroring the #793 pattern.

### Acceptance

- [ ] [S] Settings toggle, persisted + env override, default OFF
- [ ] Session app log with exception tracebacks when enabled
- [ ] All write-runner invocations logged when enabled (download parity)
- [ ] Read-runner failures logged when enabled
- [ ] Decision recorded on the #793 always-on question
- [ ] Redaction + prune + tests

noonghunna:
## Decided behavior

The new master switch defaults OFF and governs the app session log plus non-download subprocess logs. Download logs remain always-on exactly as shipped in #793.

## Acceptance

- [x] Add a persisted Settings toggle with per-launch C3_LOG=1|0 precedence.
- [x] Create one discoverable UTC-stamped session log and capture screen/action transitions plus full unhandled-exception tracebacks.
- [x] Generalize RunLog for streamed writes and read-runner command/exit records while omitting healthy JSON output.
- [x] Never serialize child environments or Settings values; add token/env redaction regressions.
- [x] Retain the newest 10 session logs while downloads keep their existing newest-30 policy.
- [x] Keep logging default-off and cover explicit enable, footers, failures, pruning, and callback restoration hermetically.

## Tests

- 262 passed: test_services.py + test_registry_parser.py
- 88 passed, 483 deselected: test_app_headless.py filtered to settings or log
- Python syntax compile and git diff --check passed

No GPU or live-service validation was run or required.

Fixes #795

### DIFF
--- a/tools/serve-cockpit/club3090_cockpit/__main__.py
+++ b/tools/serve-cockpit/club3090_cockpit/__main__.py
@@ -55,7 +55,7 @@ def save_settings(settings: dict) -> None:
 
 
 def apply_persisted_settings(app, environ) -> None:
-    """Apply MODEL_DIR / HF_TOKEN to the app + process env before run().
+    """Apply MODEL_DIR / HF_TOKEN / C3_LOG to the app before run().
 
     Precedence (highest first): an explicit shell env var > the persisted
     setting (``c3-settings.json``) > the bundled default.  An env var is a
@@ -79,6 +79,15 @@ def apply_persisted_settings(app, environ) -> None:
     cols = s.get("catalog_columns")
     if isinstance(cols, dict):
         app.catalog_columns_pref = cols
+    # Master logging: strict C3_LOG=1|0 shell override wins for this launch.
+    # Invalid/absent values fall back to the persisted boolean, default OFF.
+    from .session_logging import env_log_override
+
+    env_override = env_log_override(environ)
+    persisted_log = s.get("logging_enabled")
+    enabled = env_override if env_override is not None else persisted_log is True
+    app._c3_log_env_override = env_override is not None
+    app.configure_session_logging(enabled)
 
 
 def load_surface_setting() -> "str | None":
--- a/tools/serve-cockpit/club3090_cockpit/app.py
+++ b/tools/serve-cockpit/club3090_cockpit/app.py
@@ -42,6 +42,7 @@ NEVER executed live — tests inject fakes and conftest blocks the real spawn.
 from __future__ import annotations
 
 import dataclasses
+import logging
 import re
 from collections import OrderedDict
 from pathlib import Path
@@ -61,6 +62,7 @@ from textual.widgets import (
     Label,
     Select,
     Static,
+    Switch,
     TabbedContent,
     TabPane,
     Tabs,
@@ -4348,11 +4350,24 @@ class SettingsScreen(ModalScreen):
         Binding("escape", "cancel", "Cancel", show=True),
     ]
 
-    def __init__(self, model_dir: str, hf_token_set: bool, director_device: str = "gpu0", **kwargs):
+    def __init__(
+        self,
+        model_dir: str,
+        hf_token_set: bool,
+        director_device: str = "gpu0",
+        *,
+        log_enabled: bool = False,
+        log_path: str = "",
+        log_env_override: bool = False,
+        **kwargs,
+    ):
         super().__init__(**kwargs)
         self._model_dir = model_dir or ""
         self._hf_token_set = hf_token_set
         self._director_device = director_device if director_device in ("gpu0", "gpu1", "cpu") else "gpu0"
+        self._log_enabled = log_enabled
+        self._log_path = log_path
+        self._log_env_override = log_env_override
 
     def compose(self) -> ComposeResult:
         with Vertical():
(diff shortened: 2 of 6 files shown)


noonghunna:
**Review: NEEDS-CHANGES — one blocker, everything else ships as-is.** (Independent review, isolated worktree; suites re-run and verified: 262 fast + 88 targeted match the PR claims, PLUS a full 571-test headless run — green — since the `_dispatch_action`/`_handle_exception` overrides execute in every test. The two non-negotiables both PASS: default-OFF verified at init/env/persisted layers; redaction verified — env never reaches a log, argv only, settings values and HF_TOKEN asserted absent.)

**BLOCKER (empirically reproduced):** read-runner logs share the write-runner prune pool. `services.py:237` creates a `RunLog("read", …)` file per `RealRunner.run` call in the same `c3-run-*` pool (KEEP=30) as the serve/switch/verify/bench logs. The estate poll fires ~8+ reads per batch — every 3s during the 90s burst a serve action itself triggers — so **a crashed serve's log was unlinked after 35 reads (~13 s of burst polling)** in the repro. That's the exact scenario #795 exists for: serve crashes, user opens the log, the log is gone. Fix: give reads their own category + KEEP (e.g. `c3-read-*`), or fold read records into the session log instead of per-call files.

**Should-fix alongside it:** `services.py:341-360` treats "non-parseable" as "not JSON", so healthy `df`/`meminfo`/`nvidia-smi`/`docker ps` poll output gets fully dumped every batch (confirmed) — compounds the churn. Suppress healthy non-JSON for the polling read class.

Noted, non-blocking: overlapping write runs can cross-wire log attribution (pre-existing single-callback architecture, extended not introduced); Textual private-API overrides verified against pinned 8.2.8 but unguarded for future bumps; `C3_LOG` env disables the [S] switch entirely (deviates from the MODEL_DIR editable-under-override precedent — communicated in-modal, defensible); download log filename gained microseconds (needed for collision-avoidance — worth a PR-body line); unwritable log dir → silent ON-with-no-logging.

Fix the blocker (± the should-fix) and this merges.

noonghunna:
Blocker addressed in 56fef24; re-review requested.

- Read calls now use a separate c3-read-* pool with KEEP=100, so estate polling cannot prune c3-run-* serve/switch/verify/bench logs.
- Successful read bodies, including healthy non-JSON df/meminfo/nvidia-smi-style output, are suppressed; argv and exit status remain.
- Failed reads still retain stdout/stderr.
- Added a churn regression that creates 105 reads, verifies the crashed serve log survives, verifies read retention, and verifies healthy poll bodies remain absent.

Validation:
- focused blocker regressions: 2 passed
- required fast suite: 263 passed
- targeted headless settings/logging suite: 88 passed, 483 deselected
- py_compile and git diff --check passed

No GPU or live-service operations were run.

noonghunna:
**Re-review of 56fef24: MERGE.** Both flagged issues cleanly resolved with a minimal delta: reads now live in their own `ReadLog` subclass with a `c3-read-*` pool (KEEP=100) — **the empirical churn repro was re-run against the fix: a completed serve/verify write log survives 150 reads intact** (previously unlinked at 35), with the `c3-run-*` pool untouched. Healthy read bodies suppressed (argv/exit retained), failed reads keep full stdout+stderr, and the new churn regression test is non-vacuous (constructs KEEP+5 ReadLogs and asserts the write log survives *with content*). Zero redaction or default-OFF drift — the commit has no env access and `ReadLog` sits behind the same default-False gate. Suites: **263 fast + 88 targeted headless green.** Remaining LOW/nit findings from the first pass stand as non-blocking follow-up candidates.

PR #802, opened 2026-07-27
power: fix multi-GPU sweep metrics and limits

### LINKED ISSUE #796: power-cap-sweep.sh: multi-GPU TPS/W presentation, 2.6× decode-single vs bench.sh gap, heterogeneous-ceiling guard (#773 findings)
Three findings from @alesha-pro's unmodified `power-cap-sweep.sh` run on a 4× 3090 rig against a live TP=2 service ([disc #773](https://github.com/noonghunna/club-3090/discussions/773), §4 of the 2026-07-27 01:17 comment), plus one doc scoping item the same run produced.

### 1. Multi-GPU presentation: summed draw vs per-card cap

The power column sums across active cards (438 W at a 220 W cap = 2×219) while the cap column is per-card, and the header renders a single card (`GPU: NVIDIA GeForce RTX 3090, Stock TDP: 350.00W`). TPS/W on a multi-GPU sweep is therefore silently computed against **combined** draw. Arithmetic correct, presentation misleading. Fix: label draw as `combined (N cards)` and either normalize TPS/W per-card or state the basis in the summary line.

### 2. `decode-single` vs `bench.sh` disagree ~2.6× on the same service

Same container, same session, ~10 minutes apart: `bench.sh` measured 78–107 wall TPS; the sweep's `decode-single` reported 28–31. The sweep's *shape* remains readable (the docs already say absolutes aren't the cross-rig signal), but a 2.6× gap between two of our own tools needs a root cause — candidate suspects: no warmup / MTP cold-start inside the sweep's probe, different prompt/max_tokens, or measurement window including non-decode time. Needs a side-by-side of the two probes' protocols.

### 3. Heterogeneous per-card power ceilings

`power.max_limit` on the reporting rig reads **350 / 385 / 365 / 350 W** across four same-model 3090s. Every cap in the run was below all ceilings, so that sweep stayed uniform — but a sweep reaching e.g. 370 W would silently become non-uniform (two cards capped, two not). Fix: read `power.max_limit` per card up front and WARN (or clamp with a note) when a requested cap exceeds the minimum ceiling.

### Related doc scoping (HARDWARE.md, separate touch)

The same run found **no knee above 220 W on `dual-fast` TP=2** (220→330 W: TPS flat within noise, draw +44%, SM clock +24% while mem clock sits pinned at 9501 — bandwidth-bound decode can't spend core clock). This doesn't contradict the single-card 290 W sweet spot; it scopes it. HARDWARE.md's power section should label the existing curves **single-card** and can take the TP=2 datum as a new anchor row (credited); a `--load-mode decode-concurrent` sweep was offered by the reporter as the missing cell.

noonghunna:
## 1. Multi-GPU presentation

- [x] Enumerate every active card with its model, VRAM, min/default/max limit.
- [x] Label caps as W/card and sampled draw as combined across N cards.
- [x] Define efficiency as narrative TPS per combined watt across N cards.
- [x] Identify primary-card clock/P-state fields and any-card throttle aggregation.

## 2. decode-single versus bench.sh

Static protocol comparison found a measurement-basis bug, not a 2.6x engine regression:

| Basis | Old sweep probe | Canonical bench.sh |
|---|---|---|
| Warmup | None despite stale comments | 3 completed warmups per prompt |
| Shape | max_tokens=99999, cut at 10s; code prompt differed | completed 1000/800-token requests with canonical prompts |
| Token count | final usage was normally unavailable after timeout, so content-bearing SSE chunks were counted as tokens | final usage.completion_tokens |
| Window | chunks / full wall, including TTFT | completion_tokens / (wall - TTFT) |

The dominant error was treating SSE transport chunks as tokens; one chunk may contain multiple tokens. First-cap cold start and TTFT inclusion added smaller downward biases.

The corrected probe uses the canonical prompts/request shape, three completed warmups per prompt once at the highest cap, stream usage tokens, and weighted completion_tokens / decode-seconds. It repeats completed 500/400-token requests until each prompt covers the requested wall budget, so fast cards still provide a stable power-sampling window and slow requests are never truncated. The reduced request size preserves practical sweep runtime.

## 3. Heterogeneous ceiling guard

- [x] Capture power.max_limit for every active card.
- [x] Keep auto-derived caps inside the intersected envelope.
- [x] WARN and clamp explicit caps to min(per-card max_limit), removing duplicates created by clamping.
- [x] Verify and print the applied limit on every participating card.

## 4. HARDWARE.md scope

- [x] Scope existing efficiency curves and 3090 recommendations as single-card.
- [x] Add the credited @alesha-pro discussion #773 TP=2 anchor: 220 to 330W/card stayed flat within noise while combined draw rose 44%, SM clock rose 24%, and memory stayed at 9501 MHz.
- [x] Record decode-concurrent on the same TP=2 service as the missing cell.
- [x] Flag legacy time-bounded absolute TPS as a historical metric basis while retaining its curve-shape evidence.

## Tests

- bash -n passed for the sweep and its test.
- 28 mocked multi-GPU and source-protocol checks passed.
- test-locale-utf8.sh passed over 90 scripts.
- Embedded decode-probe Python compiled successfully.
- git diff --check passed.

## Deferred live validation

Per the task constraint, no sweep, model request, service boot, Docker GPU workload, or other GPU operation was run. A maintainer should run a live same-service comparison and the missing TP=2 decode-concurrent sweep when the GPUs are available.

Fixes #796

### DIFF
--- a/docs/HARDWARE.md
+++ b/docs/HARDWARE.md
@@ -138,7 +138,7 @@ then `bash scripts/launch.sh --gpus 1,2` as normal (the composes pass `CUDA_VISI
 
 ## Power
 
-Production target: **290W (air-cooled) / 330W (water-cooled) per card** is the sweet spot — peak TPS/W efficiency and only ~5-7% TPS loss vs unrestricted stock.
+Single-card production target: **290W (air-cooled) / 330W (water-cooled)** is the sweet spot — peak TPS/W efficiency and only ~5-7% TPS loss vs unrestricted stock.
 
 Power lever:
 ```bash
@@ -152,9 +152,9 @@ Past the sweet spot: diminishing returns (SM clocks saturate near 1.9 GHz on 309
 
 **The "230W is the sweet spot" lore is wrong** — it traces back to early thermal-constrained recommendations and 3-cap-resolution data. Dense 10W sweeps show 230W costs ~16% efficiency vs 290W (decode) and ~4% vs 250W (prefill) on this rig. 230W is a *low-power / quiet* cap, not an efficient one. Use it only if your goal is thermal/acoustic, not perf-per-watt.
 
-**Sweet spot varies by workload class** — same card, same engine, same model:
+**Single-card sweet spot varies by workload class** — same card, same engine, same model:
 
-| Workload | Air-cooled 3090 sweet spot | Notes |
+| Single-card workload | Air-cooled 3090 sweet spot | Notes |
 |---|---:|---|
 | Decode-single (chat / IDE agent) | **290W** (0.111 TPS/W) | -7% TPS vs 370W stock for -22% wattage |
 | Decode-concurrent (multi-stream) | **290W** (0.110 TPS/W) | Same knee as decode-single — concurrency doesn't move it |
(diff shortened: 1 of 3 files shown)


noonghunna:
**Review: MERGE-WITH-NITS** (independent review; all checks re-run in an isolated worktree). All four spec items verified in the actual code paths — the multi-GPU labeling, the per-card ceiling guard with clamp+warn, and the HARDWARE.md anchor cross-checked number-by-number against the #773 source. The §2 root cause was verified strictly against both scripts' sources and is correct; **28/28 mocked tests green and non-vacuous**, locale gate green, `bash -n` clean.

One sentence would make the §2 writeup airtight: the reporter's service runs **MTP spec-dec, and vLLM streams all accepted tokens of a step in one SSE delta** — chunk-counting therefore undercounts by ≈ the mean acceptance length, which is exactly the ~2.6× observed.

Non-blocking findings for a follow-up (or pre-merge if trivial):
1. `clamp_caps_to_ceiling` (~:348-370): validate the ceiling argument — an empty string clamps every cap to 0 (degenerate sweep); a `[N/A]` max_limit silently no-ops the guard.
2. `:1590`: the `[result]` line still prints `cap=%sW` while every other surface says `W/card`.
3. `:998-1010`: the warmup `nvidia-smi -pl` lacks the `|| _cap_ok=0` tolerance the cap loop has (`set -e` exits unfriendly).
4. Behavioral note: `decode-single` now hard-requires streaming `usage` — fail-fast is defensible (and the message is explicit), but older llama.cpp builds without `include_usage` will refuse the sweep.

Mergeable as-is; branch is 1 commit behind master (no conflicts in these files).

PR #821, opened 2026-07-29
fix(spark-dashboard): enable multi-GPU support and fix Docker GID detection

BlackBox-Labs:
## Problem

The spark-dashboard service only reports stats for 1 of 2 RTX 3090 GPUs. The vLLM engine was also not being detected until the DOCKER_GID was fixed.

### Root Causes

1. **`SPARK_DASHBOARD_GPU_INDEX=0`** — hardcoded in the original commit (9ce4adf7), forces the dashboard to only monitor GPU 0. The upstream spark-dashboard repo's default is 0, and this repo inherited it.

2. **Image pinned to v0.11.0** — the multi-GPU selector UI was added in spark-dashboard v0.12.0 (commit a34c31a). v0.11.0 has no way to view GPU 1 even if the backend detects it.

3. **DOCKER_GID default mismatch** — the upstream commit 4ed7833 changed the default from 988 → 999 (community common). This rig's Docker GID is 984. When GID mismatches, the container can't access the Docker socket, so engine discovery fails silently with "permission/socket issue".

### Evidence

- `docker logs spark-dashboard` shows: `NVML initialized: 2 GPU(s) available` — backend sees both GPUs
- But the UI only shows one GPU's hardware panels (no selector in v0.11.0)
- Engine detection logs: `Failed to list Docker containers (permission/socket issue?)` until GID was corrected to 984
- After fixing GID: `Detected engine: vLLM at http://localhost:8010`

### services/spark-dashboard/docker-compose.yml

| Setting | Before | After |
|---------|--------|-------|
| `image` | `v0.11.0` | `v0.13.0` ✅ |
| `SPARK_DASHBOARD_GPU_INDEX` | `=0` (hardcoded) | removed ✅ (defaults to all) |
| `DOCKER_GID` | `999` | `984` ✅ (matches rig) |

### Verification

After fix, the dashboard should show:
- Both GPU 0 and GPU 1 with a clickable selector
- vLLM engine auto-detected under "LLM Engines" section


noonghunna:
Merged — thanks for the systematic root-cause work: all three diagnoses (the inherited GPU_INDEX=0 hardcode, the v0.12.0 selector-UI requirement, and the GID/socket silent-discovery failure) reproduced exactly on our 2×3090 reference rig, and v0.13.0 came up clean with both GPUs visible and the socket accessible. One fixup pushed to your branch before merge (kept your authorship on the fix itself): the DOCKER_GID default stays parameterized at 999 rather than moving to 984 — 984/988/999 are all real-world docker GIDs, so any specific default is wrong for someone; the compose header now says explicitly to set yours via `getent group docker | cut -d: -f3`. The unrelated UPSTREAM.md row also dropped out of the diff. Dashboard on this rig is now running your config.

Issue #824, opened 2026-07-30
offload-matrix: test suite for the sweep harness (it has already shipped silent defects)


noonghunna:
Follow-up to #823, which adds `scripts/offload-matrix.sh` + `scripts/offload-matrix-render.py` + `docs/OFFLOAD_MATRIX.md`.

## Why this needs tests more than most scripts

The harness exists to produce numbers that decide serving configs. A bug in it does not crash — it produces a plausible wrong number, and the config decision downstream is silently wrong. That is not hypothetical: **six defects of exactly that shape shipped in it during development and were caught by luck**, not by any check.

1. **Silent column misalignment.** The TSV `printf` had 32 format specifiers for 33 arguments. Bash re-runs the format string for surplus args, so every arm wrote a *short* row plus a *junk* row containing only the status. Fixed by routing all rows through one emitter that derives width from the header.
2. **Early-exit rows had a different column count** than measured rows (`BOOT_FAIL`/`TIMEOUT` omitted `shape` and were 4 short).
3. **Zero throughput reported as `OK`.** An empty `SHAPE` made the driver exit; the arm recorded `agg=0.00`, status `OK`. Now `NO_TOKENS`.
4. **Renderer referenced columns the script never emits** — one crashed with `KeyError`, the other silently rendered `-` forever.
5. **Invalid pairing.** Spec arms pair against a no-spec baseline; the pairing key initially omitted `shape`. Workload shape *flips the sign* of speculation results on the reference rig, so a mispair manufactures a gain that does not exist.
6. **`CACHE_DISABLED`.** The engine's default expert-cache reserve can consume the whole budget at high context, so the pool is never allocated and the run measures the no-cache path — 12-22% slow, reported healthy. It logs "no cache budget", which no bypass-only detector matches.

## Scope

A tiered suite, maintainer-side plan already written (61 cases, 28 risks tagged SILENT/LOUD, 15 row invariants, plus a fake-`llama-server` contract table giving each log scrape its exact pattern and a satisfying example line):

| tier | needs | budget |
|---|---|---|
| **0 — static/lint** | nothing | < 5 s |
| **1 — mocked server** | nothing (fake `llama-server` + fake `nvidia-smi`) | < 90 s, ~25 fake arms |
| **2 — real short run** | GPU, on demand, kept out of the default `scripts/tests/*.sh` sweep | ~15-20 min |

[...]

PR #825, opened 2026-07-30
offload-matrix: Tier 0 + Tier 1 test suite, and fix a partial-cache blind spot

noonghunna:
Closes the first two tiers of #824, and fixes a detector defect found while validating the engine work that motivated it.

## The fix: `CACHE_PARTIAL`

`CACHE_DISABLED` keyed on the presence of a `[moe-cache] enabled:` line. A **per-device** budget failure still prints that line while one GPU runs with no pool at all — measured live on the reference rig: CUDA0 got no budget, CUDA1 allocated a 69-slot pool, `enabled:` printed, and the arm reported **`OK`**. Throughput on such a run blends a cached and an uncached device and is comparable to neither.

Detection now counts **devices that actually got a pool** against the device count:

| pooled devices | status |
|---|---|
| 0 | `CACHE_DISABLED` (unchanged) |
| 0 < n < NGPU | **`CACHE_PARTIAL`** (new) |

## Tier 0 — static, 9 checks, ~1.2 s

No server, no GPU, no model. Each check guards a defect that actually shipped in this harness during development (all six are listed in #824):

- **column parity** — header width == `emit_row` values + status, and the width must derive from `$HDR`. Guards the `printf` that carried 32 specifiers for 33 arguments: bash re-ran the format for the surplus arg, so every arm wrote a short row *plus* a junk status-only row.
- **status-enum parity**, sweep ↔ renderer. This one has already earned itself — `CACHE_DISABLED` shipped with no renderer annotation within an hour of being added, and it caught `CACHE_PARTIAL` in this very PR.
- **pairing key includes `shape`** — workload shape flips the *sign* of speculative results, so a mispair manufactures a gain that does not exist.
- plus: syntax both halves, `MODEL` mandatory, engine-scope refusal, `PLAN=1` boots nothing, renderer survives a mixed TSV, and the suite leaves no artifacts in the repo.

**Mutation-tested**: deleting the `CACHE_DISABLED` annotation reds the suite with the right message. A green run is not evidence that a check can fail.

## Tier 1 — mocked server, 5 scenarios, ~15 s

Runs the **real sweep end-to-end** against a fake `llama-server` that emits the exact log lines the script scrapes and serves the two endpoints it calls. No GPU, no model, no engine — but boot → drive → scrape → emit → render all execute, which is where the silent defects live.

| scenario | asserted status |
|---|---|
| healthy | `OK` |
| **partial** | **`CACHE_PARTIAL`** — the regression guard for the fix above |
| nobudget | `CACHE_DISABLED`, remedy named |
| bypass | `INVALID_BYPASS` |
| notokens | `NO_TOKENS` |

Each case also asserts row invariants (field count == header count; `OK` implies non-zero throughput; `bypass>0` implies not `OK`), and a final pass proves the renderer surfaces every failure status and separates them from the conclusions.

**Throughput values are deliberately never asserted.** They are not portable across rigs, and the failure mode that matters is a healthy status on an unhealthy run.

The fixture documents the scrape contract by construction — it emits only what the script actually reads, so an accidental dependency on some other log line fails the suite rather than working by luck.

## Notes

- `test-locale-utf8` caught the missing `PYTHONUTF8` export in both suites before this PR opened — exactly its job.
- The mocked suite's EXIT trap needed a bracketed process pattern: the unbracketed form matches the killing command's own command line, self-kills with exit 144, and leaves the fixture it was meant to reap still running. Observed, then fixed.
- Tier 2 (short real-GPU run) is still to come and will stay **out** of the default `scripts/tests/*.sh` sweep so the suite remains fast.

## Tests

`test-offload-matrix` ✅ · `test-offload-matrix-mocked` ✅ · `test-locale-utf8` ✅ (93 scripts) · `test-artifact-inventory` ✅. Scoped change — no registry, compose or profile touched.

### DIFF
--- a/scripts/offload-matrix-render.py
+++ b/scripts/offload-matrix-render.py
@@ -129,6 +129,7 @@ if bad:
                 "BOOT_FAIL":      "  server never started (OOM at this ctx/pool is the usual cause)",
                 "TIMEOUT":        "  server did not reach /health before BOOT_TIMEOUT",
                 "CACHE_DISABLED": "  expert cache requested but NEVER allocated -- try RESERVE_MB=1536",
+                "CACHE_PARTIAL":  "  cache allocated on only SOME devices -- throughput blends cached and uncached",
                 "PORT_BUSY":      "  port was already served by another process; arm did not boot",
                 "MB_CLAMP_CONFLICT": "  n-max too high for the [1,8] MAX_BATCH clamp; cache would bypass"}.get(r["status"], "")
         print(f"   {r['arm']}: {r['status']}{note}")
--- a/scripts/offload-matrix.sh
+++ b/scripts/offload-matrix.sh
@@ -565,8 +565,15 @@ PY
   # Found live 2026-07-30: the engine's DEFAULT reserve (3072 MiB) left no budget at
   # 262K ctx, logging "CUDA0 has no cache budget after 3072 MiB reserve" -- a string the
   # bypass grep does not match. Both arms reported OK and ran ~12-22% slow.
-  local cache_enabled=0 reserve_seen=""
-  command grep -q '\[moe-cache\] enabled:' "$log" && cache_enabled=1
+  # Count DEVICES THAT ACTUALLY GOT A POOL, not the presence of the "enabled:" line.
+  # A per-device budget failure still prints "enabled:" while one GPU runs with no
+  # pool at all — measured 2026-07-30: CUDA0 got no budget, CUDA1 allocated 69 slots,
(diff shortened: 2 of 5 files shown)


noonghunna:
Superseded by #835, which grew this into the 3-tier suite (static + mocked + tier2-real, both fake-binary fixtures) and folded the partial-cache blind spot into the per-device CACHE_DISABLED fix (#824). Every file here landed there in evolved form. Closing to keep the queue honest — the test coverage this PR pioneered is on master.

Issue #841, opened 2026-08-01
bench.sh: capture backlog from the first bare-metal CPU-offload bench (11 items)


noonghunna:
Collected during the **DeepSeek-V4-Flash day-1 bring-up** — the first bench on this stack of a
*bare-metal*, *CPU-offloaded*, *chat-tuned* model. That combination exposed every soft spot in
`scripts/bench.sh` at once, because each of the three breaks a different assumption the script was
written under (a docker container to scrape, weights on the GPU, a model that answers at length).

A bench harness does not fail by crashing. It fails by printing a **plausible wrong number** — the
class [#824](https://github.com/noonghunna/club-3090/issues/824) catalogued for `offload-matrix.sh`.
Every item below is one of those, observed on a real run.

**Relationship to other work:**

- **[#249](https://github.com/noonghunna/club-3090/issues/249) (JSON measurement-record producer)** is
  the natural long-term home for most of this. Each capture below becomes a record field, and the
  `docs/BENCH_CARD.md` integrity panel fills itself instead of relying on whoever ran it being
  vigilant. Landing the captures first means #249 has something real to serialise.
- **[#835](https://github.com/noonghunna/club-3090/issues/835)** established the mocked-scrape test
  pattern (fake server + fake `nvidia-smi`, asserting status and structure rather than throughput).
  This builds directly on it — the same fixtures, extended.

---

### 1. Silent decode-scrape zero (NO_TOKENS class)

`wall_TPS` / `decode_TPS` printed `0.00` with **no warning**, while TTFT printed fine. A zero
throughput next to a healthy-looking TTFT is a scrape failure, not a slow model, and nothing said so.

Fix: (a) loud warning when measured tokens fall below a floor or computed TPS is 0; (b) when
engine-side timings are readable, cross-check them against the client-observed rate and report
divergence above 5%. The `offload-matrix` suite's invariant — *`status=OK` implies `agg>0`* — belongs
here too.

### 2. Chat-shaped canonical prompts

Chat-tuned models answer briefly and EOS: **under 100 tokens on 4 of 5 runs per shape**. The
per-shape decode statistics then degenerate to n=1 while still printing `n=5`, and the CV printed
next to them is meaningless.

Fix: an `ENDPOINT` knob so the endpoint being driven is explicit and a raw-completion mode exists for
base models; and *regardless of mode*, detect `predicted_n << max_tokens`, warn, and report
**n-usable per shape**.

### 3. Conditional cache/offload capture

[...]

PR #842, opened 2026-08-01
feat(bench): shared capture lib + bench.sh capture backlog (#841)

noonghunna:
Implements the capture backlog in #841: a shared capture library
(`scripts/lib/capture.sh`) plus the `bench.sh` wiring that consumes it, with mocked tests extending
the #835 pattern.

**Everything is additive.** An existing invocation with no new env vars produces everything it
produced before, plus the new always-on captures. `BENCH_MOCK` output is byte-identical — verified by
diffing both mock branches against `master`, and guarded by a new test, because `test-submit-bench`
and `test-measurement-record` parse it.

## Per-item status

| # | Item | Status | Notes |
|---|---|---|---|
| 1 | NO_TOKENS class | ✅ done | Loud warn naming the failure class (scrape, not a slow model) + "do not quote this run". Engine-side cross-check warns above 5% divergence. |
| 2 | Chat-shaped prompts / n-usable | ✅ done | `ENDPOINT=chat\|completion`. **`chat` is the default** — `bench.sh` already drove `/v1/chat/completions`, so the default is unchanged and numbers stay comparable; the knob makes it explicit and adds the raw `/v1/completions` counterpart for base models. n-usable + short-EOS warning fire in **both** modes. |
| 3 | Conditional cache/offload capture | ✅ done | KV type + fingerprint always on; offload gates the moe-cache capture. (a) config fingerprint incl. `--moe-cache` cap + the five `GGML_CUDA_MOE_CACHE_*` vars; (b) marginal rate from start/end snapshots **per device**; (c) health counters, all-zero = pass; (d) per-device rows kept in the **output**. |
| 4 | Bare-metal server scrape | ✅ done | `SERVER_LOG=<path>` scrapes `print_timing` for decode *and* prefill, and feeds the item-1b cross-check. Also fills the summary `PP tok/s` field that previously read "log scrape unavailable". |
| 5 | Config fingerprint in output | ✅ done | `CONFIG FINGERPRINT` block: `/props` (ctx, slots, ftype, drafter) + argv fingerprint from `/proc`. No operator input required. |
| 6 | Live progress | ✅ done | Per-run one-liners, flushed, on **stderr**; stdout keeps its exact historical shape. |
| 7 | Haystack sizing | ✅ done | Token-ratio probe via `/tokenize`, else a tiny `max_tokens=1` request, else the old word heuristic. Test asserts a 2000-token target lands within ±10% where word-sizing overshoots ~37%. |
| 8 | System RAM + swap | ✅ done | Always on. `MemTotal` / `MemAvailable` / server RSS, plus a swap verdict that distinguishes *serving-process* pages in swap (run is suspect) from host-wide swap with no attributable pid. |
| 9 | RAM-bandwidth utilization | ✅ done | (a) derived `ram_rd_mbps` with the "DERIVED, NOT A PERF COUNTER" caveat and its arithmetic printed; (b) `STREAM_CALIB=1` triad ceiling, opt-in. (c) contention probe **deliberately not here** — it perturbs the run by construction and stays a separate opt-in arm, as the item specifies. |
| 10 | PCIe | ✅ done | Unconditional. Per-phase decode/prefill means **and peaks**, per device; link gen/width sampled **under load** with a warn when current < max; utilization against a **stated** denominator. |
| 11 | Import from offload-matrix | ✅ done (lib) / ⏸️ deferred (adoption) | (a)–(e) all landed in the lib and wired into bench. The lib is consumed by `bench.sh` only — see below. |

## Deferred, deliberately

**`offload-matrix.sh` is not touched by this PR** (`git diff` against `master` confirms it).
A 12-commit feature branch (`feat/offload-matrix-prefix-cache`) is in flight on that file; rewiring
it underneath would conflict for no measurement benefit. Until that lands, the sweep keeps its own
inline copies of these scrapes and **the duplication is accepted**. The reason and the expiry are
recorded at the top of `capture.sh`, and a test asserts that note still exists — so the rationale
cannot quietly rot into "why are there two of these?".

Because of that, item 11's "consumed by **both** scripts" clause is not yet satisfied, so this is
**Progress on #841**, not a close. Remaining work on the issue is exactly one step: point
`offload-matrix.sh` at the lib and delete its inline copies (delete, not leave alongside — two copies
of a scrape is how a fixed defect comes back).

Also worth noting: (c) of item 9 is out of scope by the item's own text, not by omission.

## Test evidence

New: `scripts/tests/test-bench-capture.sh` — **27 checks**, three tiers, asserting structure, status
and warnings, never a throughput value (numbers from a fake server mean nothing).

- **Tier 0** — syntax of both scripts plus the embedded python heredoc; no `pgrep`/`pkill -f`;
  `command grep` throughout the lib; the degradation-notice shape; the deferral note; idempotent
  sourcing; and an AST parse of every inline `python3 -c '...'` snippet.
- **Tier 1a** (fixture logs, real telemetry line shapes) — multi-pool-per-device census aggregation;
  the #824 hole; marginal-vs-cumulative constructed to **disagree** (48/58 marginal vs 19/23
  cumulative); health counters; acceptance with fire count and drafts-attempted; prefill/decode
  timing split; status-enum precedence; derived-demand arithmetic.
- **Tier 1b** (end-to-end, real `bench.sh` against the fake server) — healthy; `no_tokens`;
  `short_eos`; `low_fire`; full degradation (no GPU, no log, no pid); `BENCH_MOCK` shape unchanged;
  `CAPTURE=0`; both `ENDPOINT` modes verified against the server's own request log; haystack sizing;
  `STREAM_CALIB` opt-in.

Suites run (no GPU, no model, no tier-2, live services untouched):

| Suite | Result |
|---|---|
| `test-bench-capture` (new) | **PASS** — 27 checks |
| `test-offload-matrix-static` | **PASS** — 9 checks |
| `test-offload-matrix-mocked` | **PASS** — 10 checks |
| `test-measurement-record` | **PASS** — all assertions |
| `test-submit-bench` | **PASS** — 6 synthetic fixtures |

`bash -n` clean on every touched shell script; `py_compile` clean on the fixture. `shellcheck` is not
installed on this rig, so it was not run (advisory only).

The last two are the shipped consumers of `bench.sh`'s output — they are in the list because this PR
touches the script they parse, and they pass unchanged.

## Findings worth their own follow-ups

Four things surfaced while building this that are not in #841:

1. **The #824 detector generalises further than the fix that closed it.** Counting *pool lines*
   against device count is not sufficient: the current census emits one line **per pool**, and a
   device can hold several (`pool[0]`, `pool[1]`). Two pools on one device make `pool_lines >= NGPU`
   while another device holds nothing at all — so a line-count detector scores that half-cached run
   healthy. The lib counts **devices with a pool** instead, and there is a fixture (`multipool`) that
   fails a line-count detector and passes a device-count one. **`offload-matrix.sh` still uses the
   line-count check** and has this latent hole; it should be fixed when the sweep adopts the lib.
2. **Inline `python3 -c '...'` is a silent-failure hazard.** A single quote inside the single-quoted
   shell string ends it early, python receives a truncated program, and a trailing
   `2>/dev/null || true` swallows the error completely — the capture simply never prints. This bit
   this PR once. The new tier-0 guard covers `bench.sh` and `capture.sh`; a repo-wide version would
   be cheap.
3. **`$!` is not the server's pid** when a process is launched through an `env` prefix and a
   `#!/usr/bin/env` shebang — the chain re-execs and the recorded pid can already be gone. Any
   harness pinning `SERVER_PID` from `$!` will silently fall back to `pgrep -x` and adopt whatever
   real server the rig is running. The fixture now reports its own pid, as real servers do with
   `--pid-file`.
4. **A STREAM ceiling must be multi-threaded to be a ceiling.** A single-threaded triad measured
   ~3.7 GB/s on a host that sustains ~35 GB/s across 16 workers while loaded. Reporting the
   single-threaded figure as "the ceiling" would have inflated the derived utilization by ~10×,
   which is worse than reporting nothing. The calibration runs multiple workers and prints the
   worker count next to the number.

## Docs

`docs/BENCH_CARD.md` gains an appended section (the existing templates are untouched): which card
fields now auto-fill, four additions to the integrity panel (status enum incl. `NO_TOKENS`, all-zero
cache health counters, the swap check, n-usable per shape), the provenance rules for the two numbers
easiest to quote wrongly — **quote the marginal hit rate**, **never quote acceptance without its fire
rate** — and the knob table. New env vars are documented in `bench.sh`'s own header block.

Progress on #841.

---

# Update — card rendering + live-run calibration

Two additions since the first review: a **card-render toggle**, and **four fixes from the first live
bare-metal run** against a real CPU-offloaded MoE.

## 1. `CARD=` — the run emits its own BENCH_CARD

Both modes default OFF; with `CARD` unset the output is byte-identical to before (asserted, because
two shipped tests parse it).

```bash
CARD=snapshot bash scripts/bench.sh | tee run-A.log      # Template 1
CARD=ab BASELINE=run-A.log KNOB=moe-cache bash scripts/bench.sh   # Template 2
```

Rendering lives in a new `scripts/lib/card.sh`; `bench.sh` only gains the toggle wiring. A/B knobs:
`KNOB=` (attributes a deliberate difference instead of confounding the card), `NOISE_BAND=`,
`EXPECTATION=`, `BOOT_POLICY=`.

**What it refuses to do is the design.** Human judgement — verdict, A/B decision, knob name, "quiet
box" — stays an explicit `<fill: …>` placeholder. An unparseable baseline fails with
`baseline unparseable: <why>` rather than rendering half a delta (and the run's own numbers still
print). Resource rows (pool slots, VRAM) are **reported, not scored**: spending pool to buy
throughput may be the intended trade, and only the operator knows the VRAM budget.

**The A/B's teeth are the invariants.** Model, ftype, ctx, KV, drafter, moe-cache config, offload
method, argv — plus pool census per device (±5%), checked **only when the cache config matched**
(a different pool is the intended *consequence* of a cache-sizing knob, not a confound). Any
unattributed mismatch renders **⛔ CONFOUNDED as line one**, not buried in a table.

**Two producers, one schema.** The current run's record is built in-process; a baseline is parsed
back out of a saved log. The suite renders a run, re-parses its own log, and diffs the two records
field by field — **50 fields agree**. That guard caught a real disagreement during development
(the printed line and the record used different phrasings for the same fact), which would have made
every A/B compare two subtly different measurements.

JSON stays **out of scope** — that's #249's job. This emits markdown for a human; the record schema
is the seam where the two should meet, and that's noted in the file.

## 2. Live-run calibration — one required fix, three refinements

A real run against a 162 GB CPU-offloaded MoE on 2× consumer GPUs. The capture layer performed well
(0.3% client-vs-engine divergence, correct 90K-depth skip, correct marginal-vs-cumulative split), and
surfaced four cases where a number was right but its **label** was not:

| # | Fix | Why |
|---|---|---|
| **Required** | **Cache-health PASS condition corrected** | The run FAILED health on a healthy server: `skips=13955` / `2330`. `skips` is `insert_skips`, dominated by admission THROTTLE / queue exhaustion — under a low `ADMIT_AFTER` every miss is an insertion candidate and `THROTTLE` caps insertions per step, so non-zero is normal steady-state pressure (0.5% of ~3.0M misses, tracking the per-device miss asymmetry exactly). PASS is now `fill-fail == 0 AND dispatch-fail == 0 AND collect-fail == 0`; `skips`/`admission` report as load metrics with the reason they aren't defects. **The backlog spec (item 3c, "all-zero is the pass condition") was wrong and the measurement corrected it.** |
| 1 | KV type infers the engine default | Printed "unavailable" where llama.cpp's default is f16. Now `f16 (engine default; no -ctk/-ctv in argv)` — and only once the engine family is identified, so no other engine's default is guessed. |
| 2 | Derived RAM demand uses the **decode window** | Dividing whole-run misses by whole-run wall diluted it with prefill (~40% of the wall), where the cache isn't on the hot path. The counters carry no timestamps, so python now records the server-log line position at each phase boundary and the counters are read truncated there. Without `SERVER_LOG` the whole-run figure still prints, explicitly labelled diluted. |
| 3 | "leak delta" → "growth delta" under cache | +1036 MiB on a healthy run: an expert cache allocates its pool and graphs lazily on first inference, so one run grows VRAM by design. Only a monotonic rise across REPEATED runs is a leak. |
| minor | Engine-side prefill split by prompt size | Short canonical prompts (~30 tok/s) and deep haystacks (~127 tok/s) are different populations; one mean over both described neither. |

Also fixed, found by the same run: value-returning capture functions printed their own degradation
notice, which the caller's `$(...)` swallowed into the label — the run literally printed
`KV cache type  : capture: KV cache type unavailable (…)`. Callers now emit the notice, and a test
asserts no notice ever appears inside a labelled field.

### What the live run measured (validating the capture design)

- **Per-phase PCIe split is real and large**: decode rx mean **231 MB/s** vs prefill **3031 MB/s**
  (13×), peaks 1941 vs 28182 MB/s. A whole-run mean would have described neither.
- **Per-device asymmetry is larger still**: prefill rx GPU0 **5978** vs GPU1 **85 MB/s** (70×), and
  cache marginal hits CUDA0 **60.3%** vs CUDA1 **80.4%**. Averaging either away destroys the signal.
- **Client-vs-engine cross-check agreed to 0.3%** once short probe completions were excluded.
- **Derived RAM demand** ~14 GB/s of a ~36 GB/s measured STREAM ceiling.

## 3. Templates grown (supersedes "render only what the template has")

`docs/BENCH_CARD.md` Template 1 gains a **conditional** "CPU offload + expert cache" block —
rendered only when the run offloaded, replacing the generic cache line — with offload method, RSS vs
host RAM, swap, the cache config fingerprint, a per-device pool/marginal/cumulative/evictions/skips
table, per-phase PCIe + link state, and the derived RAM demand with its window stated. Template 2
gains per-device marginal-hit Δ rows and a real invariants table. Template comments mark which fields
auto-fill vs stay `<fill: …>`.

## Test evidence (this update)

| Suite | Result |
|---|---|
| `test-bench-card` (new) | **PASS** — 19 checks |
| `test-bench-capture` | **PASS** — 27 checks |
| `test-offload-matrix-static` | **PASS** — 9 checks |
| `test-offload-matrix-mocked` | **PASS** — 10 checks |
| `test-measurement-record` | **PASS** |
| `test-submit-bench` | **PASS** |

New fixtures: two hand-written baselines pinning the parser's expected shapes, plus
`live-offload-run.log` — the **real** run, redacted. It is deliberately kept in its *pre-correction*
form so the parser stays able to read logs written by older versions, which is exactly what a user
A/B-ing against an old saved log needs.

`scripts/offload-matrix.sh` remains untouched; the adoption deferral is unchanged.

https://claude.ai/code/session_01B4LeYTBtjVXgpDgXSNUSgr

### DIFF
--- a/docs/BENCH_CARD.md
+++ b/docs/BENCH_CARD.md
@@ -104,3 +104,56 @@ Three rows in this template exist because their absence has repeatedly cost real
 **See also:** [CONTRIBUTING.md](../CONTRIBUTING.md) (bench + verify protocol) ·
 [OFFLOAD_MATRIX.md](OFFLOAD_MATRIX.md) (multi-dimension sweeps — use its TSV output rather than
 hand-running many A/Bs) · [RESULTS_CARD.md](RESULTS_CARD.md) (the model-level card these bench cards feed into).
+
+---
+
+## What `bench.sh` now fills in for you
+
+`bench.sh` carries a capture layer (`scripts/lib/capture.sh`) that auto-fills most of the
+**Fingerprint** and **Integrity** sections above. Copy these straight off the run instead of
+reassembling them from memory afterwards — the reason the panel exists is that reconstructing it
+later is exactly when the awkward facts get dropped.
+
+| Card field | Where it comes from now |
+|---|---|
+| **Config** — KV, ctx, drafter, key flags | `CONFIG FINGERPRINT` block: KV cache type, served ctx + slots, weights ftype, drafter, and a `-m/-ot/-t/-ub/-ct/-ngl/-ts` argv fingerprint read from the serving process. No operator input needed. |
+| **Env** — non-default vars | `moe-cache cfg` line: the `--moe-cache` cap plus `GGML_CUDA_MOE_CACHE_{RESERVE_MB,ADMIT_AFTER,THROTTLE,MAX_BATCH,STATS}`. The **resulting pool size does not identify the arm** — the census self-limits below the cap — so record the config, not just the census. |
+| **GPU** — VRAM / util | `CAPTURE: VRAM` (idle / peak / post, per device) and `CAPTURE: PCIe` (sm%, memctl%, host cpu%). |
+| **Cache** — hits start → end | `CAPTURE: EXPERT CACHE`, per device, with **marginal** and cumulative rates. |
+| **resource-cost row** (A/B) | pool slots + total MiB per device, straight from the census. |
+
+### Additions to the Integrity panel
+
+Add these four lines to the checklist in both templates. Each is printed by the run.
+
+- [ ] **`status: OK`** — the capture layer's own verdict (`NO_TOKENS` / `REQ_ERRORS` /
+      `CACHE_DISABLED` / `INVALID_BYPASS` / `OK`). **`NO_TOKENS` means the run measured nothing** —
+      a `0.00` TPS printed next to a healthy-looking TTFT is a scrape failure, not a slow model.
+      Do not quote a throughput number from a non-`OK` run.
+- [ ] **cache health counters all zero** — `fill-fail` / `dispatch-fail` / `collect-fail` / `skips`.
+      All-zero is the pass condition; anything else invalidates the cache reading for that run.
+- [ ] **swap check PASS** — pages of the *serving process* in swap make every number in the run
+      suspect, and nothing else in the pipeline notices. Needs `SERVER_PID` (auto-detected on bare
+      metal).
+- [ ] **n-usable == n per shape** — chat-tuned models EOS early, and a 5-run shape silently
+      degenerates to n=1 while still printing `n=5`. When these differ the CV is not trustworthy;
+      re-run with `FORCE_TOKENS=<n>`.
+
+Two numbers on the card need their provenance stated, because both are easy to quote wrongly:
+
+- **Cache hit rate — quote the MARGINAL one.** Cumulative rates embed the cold-fill phase, so a
+  *bigger* pool reports a *lower* cumulative rate on the same traffic. Only the marginal rate across
+  the measured window is comparable across boots. Keep the per-device split (a CUDA0/CUDA1
+  asymmetry is routing-entropy signal; averaging it away hides it).
+- **Draft acceptance — never without its fire rate.** A drafter measured at 0.992 acceptance that
+  fired on 5 of ~20 requests contributes almost nothing end-to-end. The run prints both.
+
+### Knobs worth setting
+
+| Env | Why |
+|---|---|
+| `SERVER_LOG=<path>` | Bare-metal (`CONTAINER=none`) runs. Unlocks engine-side `print_timing` TPS, the client-vs-engine cross-check, expert-cache telemetry and drafter acceptance — all of which otherwise read "log scrape unavailable". |
+| `ENDPOINT=chat` \| `completion` | `chat` (default) applies the model's template — the historical behaviour, so numbers stay comparable. `completion` drives raw `/v1/completions` with no template, for base models. Numbers are **not** comparable across modes. |
+| `STREAM_CALIB=1` | ~3 s STREAM-triad ceiling on the host, so the derived miss-path RAM demand can be stated as a fraction ("~29 of ~99 GB/s"). Off by default. The *contention* probe (co-running STREAM during decode) is deliberately not included — it perturbs the run by construction. |
+| `FORCE_TOKENS=<n>` | The fix when `n-usable < n`. |
+| `CAPTURE=0` | Suppress the capture layer entirely (for a harness parsing the older output shape). |
--- a/docs/BENCH_CARD.md
+++ b/docs/BENCH_CARD.md
@@ -42,11 +42,44 @@ Both templates share the same four-part contract:
 **GPU:** <VRAM/card> · <power> · <util>
 **Cache (if the engine has one):** hits <start>% → <end>% (<cold | warming | plateaued>)
 
+<!-- The block below is CONDITIONAL: include it only when the run offloaded to CPU
+     (same gate the capture layer uses). It REPLACES the generic Cache line above.
+     On a GPU-resident run every row here would be a dash pretending to be data.
+     `bash scripts/bench.sh CARD=snapshot` fills all of it automatically. -->
+
+### CPU offload + expert cache
+
+**Offload:** detected via <argv -ot | --n-cpu-moe | boot log> · **RSS** <X> GiB / <Y> GiB total · swap <PASS/FAIL>
+**Cache config:** cap=<MiB> · RESERVE_MB/ADMIT_AFTER/THROTTLE/MAX_BATCH=<…>
+
+> The cache config **is** the arm identity — the pool self-limits below the requested cap, so
+> pool size alone does not identify which arm a run was.
+
+| device | pool (slots · MiB · type) | marginal hits | cumulative | evictions | skips† |
+|---|---|--:|--:|--:|--:|
+| CUDA0 | 3038 slots · 12911 MiB · mxfp4 | **60.3%** | 59.3% | 42171 | 13955 |
+| CUDA1 | 3126 slots · 13285 MiB · mxfp4 | **80.4%** | 79.4% | 9158 | 2330 |
+
+†`skips` = admission throttling (the insert queue exhausted under the configured `THROTTLE`) — a
+**load signal, not a defect**. The hard-failure counters are fill / dispatch / collect-fail, and
+those are on the integrity panel.
+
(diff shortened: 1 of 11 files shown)


noonghunna:
Review note — known behavior when rendering cards from **pre-correction saved logs** (e.g. a baseline captured before this branch's calibration fixes):

1. The integrity line "cache hard-failure counters zero" carries the old log's verdict string verbatim (which may say `FAIL — skips=…`). Old logs printed only nonzero counters, so the renderer cannot recompute the corrected pass condition from them. New runs print the corrected verdict natively.
2. The derived RAM-demand caveat states the decode window, but an old log's value was whole-run-diluted — treat old-log demand figures as lower bounds.
3. Old logs say `leak delta` where new offload runs print the cache-aware `growth delta` label.

All three converge as new logs replace old baselines; listing them here so an A/B against an old baseline is read with the right squint. Rendered-from-real-log sample verified locally (51-line snapshot card, 6 `<fill:>` placeholders, all measured fields populated).

Issue #865, opened 2026-08-02
gpu-mode status: "(no inference endpoint responding)" while a scene is live — the warning port list drifted from the probes


alexpolo1:
## What's happening

`scripts/gpu-mode.sh status` prints **"(no inference endpoint responding)"** on a rig that is actively serving. On the reference rig right now:

```
═══ Service Status ═══
NAMES                            STATUS        PORTS
vllm-thinkingcap-27b-dual-w4a8   Up 2 hours    0.0.0.0:8099->8000/tcp
litellm                          Up 2 hours    0.0.0.0:4000->4000/tcp

═══ Active Model(s) ═══
  ▶ LiteLLM @ :4000  → qwen3.6-27b, …, thinkingcap-27b
  (no inference endpoint responding)          ← while :8099 serves fine
```

Both GPUs were at 100% / 22.8 GB, and a real completion through the gateway returned HTTP 200 in 0.10s. The engine is healthy — only the status view is wrong.

## Root cause

The `Active Model(s)` section keeps **two independently-maintained hardcoded port lists**: the per-port `if curl …` probe blocks, and the port list inside the final `(no inference endpoint responding)` guard (`scripts/gpu-mode.sh:447` at `87ad57d4`). Nothing ties them together, and they have drifted apart **in both directions**:

| Port | Probe block | In the warning's list | Consequence |
|---|---|---|---|
| `:8051` qwen3.6-35b-a3b | ✅ | ❌ | prints its scene line **and** the warning, simultaneously |
| `:8020` llamacpp/single | ✅ | ❌ | same |
| `:8033` gemma-4-31b QAT | ❌ | ✅ | silences the warning while displaying nothing |
| `:8099` thinkingcap-27b | ❌ | ❌ | invisible; always trips the warning |
| `:8199` qwen3.6-40b-deckard | ❌ | ❌ | invisible; always trips the warning |

`:8099` is the one that surfaced it — ThinkingCap now runs under systemd on the reference rig (`87ad57d4`), and that port has never been in either list.

Same failure family as #843 (a live serving path diverging from the catalogue that describes it) and #830 (status output that keeps asserting something the engine contradicts). Here the divergence is internal to one function.

Secondary: the guard re-issues **9 fresh curls** at `-m 2` against ports already probed a few lines above — up to ~18s of dead wait on an unreachable host.

## Reproduction

No hardware needed — stub `curl` so only `:8051` answers:

```bash
[...]

PR #867, opened 2026-08-02
Probe the unlisted scene ports in gpu-mode status (#865 follow-up)

alexpolo1:
Follow-up to #865, rebased onto `0cf0dd78`.

**Scope changed after 0cf0dd78 landed.** This PR originally carried the same structural fix you shipped there — the `_endpoint_up` flag replacing the second hardcoded port list. That half is yours now and I've dropped it; the branch is rebased onto your commit and keeps only the part 0cf0dd78 explicitly left open.

## What's left

0cf0dd78 fixed the *drift* but not the *coverage*, and flagged that gap itself. The consequence on the reference rig is that the reporting case from #865 still shows nothing:

```
# at 0cf0dd78
  ▶ LiteLLM @ :4000  → qwen3.6-27b, …, thinkingcap-27b
        ← thinkingcap-27b is live on :8099, Up 2h, both cards at 100%

# with this PR
  ▶ thinkingcap @ :8099  → thinkingcap-27b (AutoRound W4A16 served W4A8 + built-in MTP + 262K)
  ▶ LiteLLM @ :4000      → qwen3.6-27b, …, thinkingcap-27b
```

The warning is correctly suppressed either way now (`:4000` sets the flag), so the failure is quieter than before but the operator still can't see what's serving — which was the original question that led to the issue.

Adds probe blocks for the three functional scene ports a rig in this repo can actually be serving on today, in the same shape as the twelve already present:

| Port | Model | Compose |
|---|---|---|
| `:8033` | gemma-4-31b | `dual/qat-w4a16/int8.yml` |
| `:8099` | thinkingcap-27b | `dual/w4a16/w4a8.yml` |
| `:8199` | qwen3.6-40b-deckard | `llama-cpp dual/piehsoft-q6k/mtp.yml` |

## This is a stopgap, and I agree it isn't the fix

Your note on 0cf0dd78 is right: 34 of 53 functional slugs are unprobed, and a hardcoded list cannot track a 53-slug catalog. Registry-derived discovery from `compose_registry.py`'s `default_port` is the durable answer.

I'm proposing this anyway on the narrow grounds that it costs 18 lines, unblocks the three ports that are live on real rigs now, and doesn't make the rewrite harder — discovery deletes these blocks wholesale along with the other twelve. **If you'd rather hold out for the registry-derived version, close this** and I'll take a run at that instead; I'd just want your read on one design question first: whether `status` should probe all 53 declared ports on every invocation (~53 curls, though they can be parallelised) or narrow the set first from `docker ps` port bindings and only probe those.

I dropped `:11434` (ollama) from the earlier revision of this branch — the old guard referenced it with nothing printing it, but that stack retired 2026-06-22 and reviving a line for it seemed like the wrong direction.

## Verification

Live on the reference rig (2× 3090, thinkingcap-27b on `:8099` under systemd): reports as shown above. `bash -n` clean. `test-gpu-mode-list`, `test-locale-utf8`, `test-studio-derig`, `test-setup-ai-studio-skip-path` all pass. Scoped-change guards only per AGENTS.md — one script, not a catalog-shape change.

### DIFF
--- a/scripts/gpu-mode.sh
+++ b/scripts/gpu-mode.sh
@@ -450,6 +450,24 @@ show_status() {
         m=$(curl -sf -m 2 http://localhost:8038/v1/models | python3 -c "import sys,json;d=json.load(sys.stdin);print(', '.join(x['id'] for x in d.get('data',[])))" 2>/dev/null)
         echo -e "  ${GREEN}▶${NC} gemma-12b @ :8038        → ${m:-unknown} (gemma4_unified, INT8 + bf16 KV + MTP, single card)"
     fi
+    if curl -sf -m 2 http://localhost:8033/v1/models >/dev/null 2>&1; then
+        _endpoint_up=1
+        local m
+        m=$(curl -sf -m 2 http://localhost:8033/v1/models | python3 -c "import sys,json;d=json.load(sys.stdin);print(', '.join(x['id'] for x in d.get('data',[])))" 2>/dev/null)
+        echo -e "  ${GREEN}▶${NC} gemma-qat @ :8033        → ${m:-unknown} (QAT W4A16 + INT8 PTH KV)"
+    fi
+    if curl -sf -m 2 http://localhost:8099/v1/models >/dev/null 2>&1; then
+        _endpoint_up=1
+        local m
+        m=$(curl -sf -m 2 http://localhost:8099/v1/models | python3 -c "import sys,json;d=json.load(sys.stdin);print(', '.join(x['id'] for x in d.get('data',[])))" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


noonghunna:
Merged — thanks for the courteous rebase onto `0cf0dd78` and for dropping the half that had landed there. To set expectations for the file's future since #865's design ruling: gpu-mode's probe list stays **curated and finite by design** — it's the legacy quick-status, and additions land the way this PR did (a maintainer-accepted lane at a time, following the existing pattern), never via registry-derived discovery. The full estate view is c3's job. These three were easy accepts: all real catalog lanes, one of them ✅ production, and your reference-rig before/after made the operator case plainly.

Issue #779, opened 2026-07-26
[bug] half the script layer breaks on non-UTF-8 locales — 42/84 tests fail under LC_ALL=C


noonghunna:
Found while fixing #777. That issue is one instance; this is the extent of it.

## Half the test suite does not run on a non-UTF-8 locale

```
$ for t in scripts/tests/*.sh; do PYTHONUTF8=0 PYTHONCOERCECLOCALE=0 LC_ALL=C bash "$t"; done
PASS=42  FAIL=42
```

Same tree, normal locale: **84 / 0**. Verified pre-existing on pristine `origin/master` (`36455e5`) — sampled `test-patch-attribution`, `test-compose-image-drift`, `test-hwdetect`, all fail there too.

## These are script bugs, not test fragility

Error signatures across the 42 failures:

| count | error |
|--:|---|
| 34 | `UnicodeEncodeError` |
| 29 | `UnicodeDecodeError` |
| 1 | `ASSERTION FAILED` |

Only **one** is an assertion. The rest are the scripts themselves dying while reading or printing repo files — the files being read when they die are compose YAMLs (`mtp.yml`, `dflash.yml`, `base.yml` …) and `scripts/lib/profiles/patch_attribution.py`, i.e. exactly the unicode-bearing sources the rule below is about.

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 16
```

`0xe2` is the lead byte of `—`/`→`/`⚠` — the profile-schema headers every compose carries.

## Why this is worth its own issue

[`AGENTS.md`](../blob/master/AGENTS.md) already has the standing rule ("Any Python read of a repo source file — **including python heredocs inside shell scripts** — MUST pass `encoding="utf-8"`", plus the stdout corollary), and #599 fixed the launcher table path with `test-registry-emit-no-yaml` guarding it. The guard covers a narrow slice; the rule is **widely unenforced everywhere else**.

The affected surface isn't obscure — the failing list includes `test-launch-compat`, `test-profiles-compat`, `test-pull`, `test-pullgate-*`, `test-generate-compose`, `test-diagnose-profile`, `test-kv-calc-fit`, `test-hwdetect`. So on a minimal VM or container with `LC_ALL=C`, a user plausibly cannot pull, generate a compose, diagnose a profile, or run the launch-compat path. Nobody noticed for the same reason as #776: nothing here has ever been executed off a UTF-8 rig.

## Suggested approach

[...]

PR #868, opened 2026-08-03
Add Qwen AgentWorld 35B production profile

Whamp:
## Summary

Add Qwen-AgentWorld-35B-A3B as an opt-in production model for dual RTX 3090 rigs. The shipped profile serves cyankiwi's AWQ INT4 checkpoint with TP=2, 262K context, FP8-E4M3 KV, four active serving slots, checkpoint-native thinking, and `qwen3_coder` tool parsing on stock vLLM v0.25.1.

FP8 preserved the BF16 baseline's short-context decode and long-context correctness while expanding the reported KV pool to 1,795,289 tokens. Four simultaneous ~261,529-token prompts completed for six rounds with no errors, silent outputs, or post-warm VRAM growth. INT8-PTH remained correct but slowed deep-context prefill and decode; INT4-PTH corrupted recall; TurboQuant K8/V4 failed to boot on the pinned stock image.

The model remains opt-in: this PR does not add it to `DEFAULTS` or `RECOMMENDED_DEFAULT_MODELS`.

## Type of change

- [x] New compose variant (`models/<model>/<engine>/compose/`)
- [ ] New patch / sidecar (`models/<model>/<engine>/patches/`)
- [ ] New script or tool (`scripts/`, `tools/`)
- [x] New model (`models/<new-model>/`)
- [ ] Doc-only / typo
- [x] Other: profile catalog, calibration, KV calculator, and compatibility documentation

## Verification

- [x] **Full rig + validation report attached** — attached as a PR comment; canonical rebench plus final FP8-specific reruns are identified separately.
- [x] **Profile header complete** — `✅ Production`; no unresolved production caveat.
- [x] **BENCHMARKS row added** — includes FP8 production results and the BF16 A/B baseline.
- [ ] **CHANGELOG entry added** in `models/<model>/CHANGELOG.md`.

Final FP8 configuration on 2× RTX 3090:

- `verify-full.sh`: PASS, including ordinary and streamed tool calls
- `verify-stress.sh`: 8/8; exact recall through 240,634 tokens
- Canonical bench: 147.12 / 147.24 narrative/code decode TPS
- C=4 × ~261,529 prompt tokens: 24/24 complete across six rounds
- Fresh-session soak: 100/100 HTTP 200, zero silent outputs, zero VRAM growth
- Thinking-ON quick quality: ToolCall 14/15, InstructFollow 15/15
- Affected catalog/profile/registry/launcher/submission guards: PASS
- `kv-calc.py --calibration`: 8/8

### N/A justifications

- Per-model CHANGELOG: N/A for this initial model addition; `models/qwen-agentworld-35b-a3b/` has no prior changelog. The measured history is recorded in `BENCHMARKS.md`, the compose header, and calibration source.
- `report.sh --full`: the canonical full production run used `rebench-full.sh --with-8pack-thinking=both`, which supersedes `report.sh --full`. After selecting FP8/C=4, the affected operational, concurrency, quality, and soak gates were rerun on the final configuration.

## Cross-links

- Related upstream: none; stock vLLM v0.25.1 serves this checkpoint without an overlay.


Whamp:
## Final-config validation report

**Rig:** 2× RTX 3090, PCIe x4/x16, 230 W/card  
**Engine:** stock `vllm/vllm-openai:v0.25.1`  
**Profile:** TP=2 · PP=1 · 262,144 context · FP8-E4M3 KV · `max_num_seqs=4` · no MTP

### Correctness and stability

| Gate | Result |
| --- | --- |
| `verify-full.sh` | PASS; completion, SSE, ordinary/streamed `qwen3_coder` tool calls, separated reasoning, 2K output |
| `verify-stress.sh` | 8/8; exact recall through 240,634 tokens (91% of n_ctx) |
| C=4 full-context | 6 rounds × 4 simultaneous ~261,529-token prompts; 24/24 complete, 0 errors, 0 silent, 0 MiB post-warm growth |
| Fresh-session soak | 20 sessions × 5 turns; 100/100 HTTP 200, 0 silent, 0 MiB growth, 100% TPS retention |
| Thinking-ON quick quality | ToolCall 14/15; InstructFollow 15/15 |

The checkpoint declares one MTP layer in config but ships zero `mtp.*` tensors, so the profile deliberately disables speculative decoding.

### Performance

| Workload | Result |
| --- | --- |
| Narrative | 144.29 wall TPS · 147.12 decode TPS · 133 ms TTFT |
| Code | 142.70 wall TPS · 147.24 decode TPS · 129 ms TTFT |
| Cache-busted prefill | 5,116 tok/s near 10K · 3,788 tok/s near 90K |
| C=4 × 10K | 76.5 TPS/stream · 103.3 aggregate TPS · 6,395 ms TTFT |
| C=4 × ~261.5K | ~920 prefill tok/s/stream · 1.69 decode TPS/stream · 281,416 ms median TTFT |

At full C=4 saturation, peak serving allocation was 46,828 MiB total and remained flat across all six rounds.

### KV A/B

- **FP8-E4M3:** selected; full correctness, concurrency, quality, and soak gates passed.
- **BF16:** correct and similar throughput, but provides roughly half the KV capacity.
- **INT8-PTH:** correct, but deep-context prefill fell to 2,358 tok/s near 90K and decode fell 33% around 35K accumulated context.
- **INT4-PTH:** rejected; recall corrupted near 29K/91K and the ceiling ladder failed at 184,605 tokens.
- **TurboQuant K8/V4:** rejected; stock v0.25.1 failed before readiness with `Unknown TurboQuant cache dtype: 'auto'`.

### Canonical full gate

Before the FP8 A/B, `rebench-full.sh --with-8pack-thinking=both` completed in 121m55s with RC 0 on the BF16 baseline. Thinking ON scored 125/150 versus 101/150 OFF. The final FP8 configuration then reran the affected functional, stress, benchmark, active-concurrency, quick-quality, and soak gates listed above.

BlackBox-Labs:
I am setting it up today. Going to try your compose file as i am hitting loops !

Whamp:
Nice let me know how it goes, mine has been 100% solid. I have run very heavy benchmarking on it in DeepSWE and am comparing it to ThinkingCap-27B and finding it shockingly close.

All that said - just killing time until Qwen3.8-27B is released . . .

BlackBox-Labs:
Independent reproduction on my rig — results vs your final-config report (as promised 👍)

**Rig:** 2× RTX 3090, PCIe x16 x2, **330 W/cap**, no NVLink, custom-AR off
**Engine:** `vllm/vllm-openai:v0.25.1` (same)
**Profile:** TP=2 · 262,144 ctx · FP8-E4M3 KV · `max_num_seqs=4` · no MTP · `qwen-agentworld-35b-a3b` served as `qwen-8011`

### Correctness and stability — matches your report

| Gate | Result |
| --- | --- |
| `verify-full.sh` | PASS — completion, SSE, `qwen3_coder` tool calls, separated reasoning |
| `verify-stress.sh` | 8/8 boundary checks PASS (incl. 0.92×n_ctx ceiling ladder, 5.5 GiB VRAM margin) |
| Fresh soak (20×5) | PASS — 0 errors, 0 silent-empty / 100, p50 145.95 decode, 98.5% TPS retention, 0 MiB growth |
| Stream tool-call probe | PASS (`stream-toolcall-probe.py`) |

### Performance — land within your noise band

| Workload | Yours (Whamp) | Mine (BlackBox-Labs) |
| --- | --- | --- |
| Narrative | 147.12 decode · 133 ms TTFT | **146.98** decode · 162 ms TTFT |
| Code | 147.24 decode · 129 ms TTFT | **147.94** decode · 163 ms TTFT |
| Cache-busted prefill | 5,116 @10K · 3,788 @90K | **7,465 @10K · 5,047 @90K** |

Decode is within run-to-run noise of yours (~±1%). My cache-busted prefill came in notably higher — worth a glance at the prefill harness between us, but I'll take the deeper numbers either way.

### Quality (dedicated --quick, 2 packs)

Thinking-OFF: ToolCall **12/15** · InstructFollow **15/15** → **27/30 (90%)** — IF clean, ToolCall trails your 14/15 by 2, inside noise.

### Heads-up flags (rig-cell honesty)

- My cap is **330 W** vs your 230 W — decode landed about the same, so that checks out.
- Soak and stress clean on my side — no Cliff 2b.
- Full-phase detail: tag `results/rebench/qwen-8011-20260806-0404/REPORT.md` (regen'd, quality folded). Happy to attach the per-phase logs if useful.

Solid compose — consistent across two rigs now. Nice work.

BlackBox-Labs:
I do want to comment, that my gpu is a bit more stressed then with regulay 35b-a3b slug and harness through some tool errors, though they recover automatically. 

⚠️  Stream interrupted by network error (finish_reason='length' on partial-stream-stub)

BlackBox-Labs:
Problem Statement:
The core issue stems from a fundamental architectural mismatch between AgentWorld's Qwen-AgentWorld-35B-A3B model and the agent harness execution environment. Unlike instruction-following models that rely on system prompts and tool schemas to guide behavior, AgentWorld was trained as a world-model where environment dynamics themselves are the pretraining objective, giving it internalized priors about how terminals, web pages, and GUIs transition between states. However, the current harness environment suppresses this native reasoning mode by truncating or discouraging the "extra" reasoning about environment state that is actually load-bearing computation for the model. Additionally, there's a native observation schema mismatch where the harness formats tool outputs according to Claude-Code expectations rather than the specific formats AgentWorld was trained on (such as proper terminal output chunking or GUI state representations), and aggressive context compaction breaks the continuity the model depends on to keep predicting state correctly, resulting in the "drift and thrash" pattern and reasoning spirals where the model enumerates alternatives rather than committing to actions.

Response and Supporting Evidence:
The trace data from the 5-minute capture provides concrete evidence supporting this diagnosis:

1. Evidence of Suppressed Native Reasoning Mode: The model's reasoning traces show it attempting to reason through multiple steps but being constrained, with examples like "Thinking Process: 1. Understand the Goal: The user wants to check if the file 'instrument-vllm-capture-8011.py' exists... 2. Identify the Command: The test command in shell... 3. Formulate the Command: test -f instrument-vllm-capture-8011.py && echo \"exists\" || echo \"does not exist\" or simply test -e instrument-vllm-capture-8011.py; echo $? - showing the model considering multiple command formulations and second-guessing itself, which indicates the reasoning spiral behavior.

2. Evidence of Native Observation Schema Mismatch: The tool execution observations are formatted in a way that may not match what the model was trained on, with structured JSON formats like "actual_tool_observation": {"command": "free -m", "return_code": 0, "stdout": "...", "stderr": "", "timestamp": "..."} rather than the native observation formats AgentWorld was likely trained on.

3. Evidence of Context Compaction Issues: The model shows repeated re-derivation of solutions across turns, with the same reasoning pattern appearing in multiple turns (e.g., "Thinking Process: 1. Understand the Request: The user wants to list the files in the current directory using the ls command. 2. Identify the Command: The command to list files in the current directory is ls..."), suggesting the model is not building on previous context but rather re-deriving solutions from scratch each time, which aligns with the "drift and thrash" pattern caused by aggressive context compaction breaking continuity.

BlackBox-Labs:
Based on the Qwen Code CHANGELOG.md, I can see that they have indeed made several adaptations to support world-model-native architectures and reasoning content. Here are the key adaptations I found:

Qwen Code Harness Adaptations for World-Model-Native Architectures

1. Reasoning Content Preservation
- "Enables support for Qwen 3.8 reasoning effort levels by mapping existing effort flags to the provider-specific reasoning_effort parameter."
- "Transport stream retries are now allowed during the thinking-only phase by distinguishing non-thought content chunks from intermediate processing steps."
- "fix(core): render a thought part's reasoning instead of the boolean flag"
- "fix(core): estimate reasoning_tokens when completion_tokens_details is missing"
- "core: preserve OpenAI reasoning as raw thoughts"

2. Thinking Block Management
- "Expanding thinking blocks with Alt+T during streaming now displays the full reasoning content in real-time instead of a truncated preview."
- "Stabilizes the thinking block height to prevent layout shifts and replaces the full-screen transcript overlay with an inline Ctrl+O toggle."
- "Prevents stale thinking blocks from appearing by removing them when their associated tool calls are orphaned or contain empty text."
- "fix(core): retry malformed repeated thinking tags"
- "fix(core): handle unsigned Claude thinking from proxies"
- "fix(core): sanitize standalone closing thinking tags"

3. Context and Compaction Handling
- "Added a threshold setting to automatically preload deferred tools into the context window when their total size fits within limits."
- "Added the --compaction flag to the /model command to configure a dedicated model for chat compression."
- "Refactors system prompt assembly into a layered builder to explicitly manage stable, context, and volatile instruction layers."

4. API Parameter Adjustments
- "Resolved DashScope API request failures for Qwen models by correcting conflicting reasoning parameters and honoring explicit opt-outs."
- "Fixed DashScope API requests for Qwen 3.8 models to prevent conflicting reasoning parameters that caused request rejections."
- "Resolved DashScope thinking parameter conflicts by correctly honoring explicit opt-outs and preserving legacy budget settings."

noonghunna:
Reviewed and tested **as-if-landed** — I merged current master into your branch locally rather than testing it 49 commits stale, since a stale-branch failure tells us nothing. Result: **99 pass / 0 fail** on the full suite.

This is a well-formed catalog addition: profile, compose, calibration, registry entry, count bumps, profile-compat coverage, and docs. That's the lightweight-path checklist actually followed, which is rarer than it should be — thank you.

## `kv-calc.py` — reviewed closely, it's clean

It's the one file here that affects *every* model, so I checked it specifically. It adds `qwen-agentworld-35b-a3b` alongside the existing `agents-a1` entry, reusing the Qwen3-Next MoE geometry (10 attention + 30 GDN) with its own AWQ INT4 footprint, plus an alias row. Same shape as the `agents-a1` precedent, purely additive — no existing model's math moves. The note that the inherited MTP declaration normalises to zero because no `mtp.*` tensors ship is exactly the right thing to record.

## What needs a rebase

Master has moved (three DeepSeek slugs landed since), so the count bumps collide. After rebasing, the numbers are:

| assertion | value |
|---|---|
| `test-compose-registry-disk`: registry | **73** |
| `test-compose-registry-disk`: disk | **74** |
| `test-profiles-compat`: `len(p.models)` | **16** |
| `test-profiles-compat`: `len(p.drafters)` | **16** ← moved on master, not by you |

I verified these against the loaded profiles rather than arithmetic. The `drafters` one is easy to miss because it isn't a count your PR changes — master moved it underneath you.

## Optional, not a blocker: the `setup.sh` front door

A new model id needs wiring into `scripts/setup.sh` to be downloadable via the guided path, otherwise `setup.sh qwen-agentworld-35b-a3b` is rejected outright. Your PR doesn't touch it.

**I'm explicitly not asking you to fix that**, because I checked before raising it: **9 of our 15 catalog models already lack it**. Holding your PR to a standard two-thirds of the catalog doesn't meet would be unfair. Mentioning it only so you can opt in if you want the guided download — `docs/ADDING_MODELS.md` Step 4c lists the five hand-written lists involved.

## What would help before merge

A `bash scripts/report.sh` from your rig plus whatever validation you ran (verify-full at minimum). The compose header carries a `Status:` claim, and the repo's rule is that first-party status claims are backed by measurements from the rig that made them — yours, in this case, since we haven't run this model.

Rebase + those four numbers and I'll merge it.

Whamp:
Rebased onto current master (`a6385ece`) and force-pushed head `7afc001f`. GitHub now reports the PR **CLEAN / MERGEABLE**.

Resolved counts:

| assertion | value |
|---|---:|
| registry entries | 73 |
| compose files on disk | 74 |
| models | 16 |
| drafters | 16 |

Validation on the rebased branch:

- Full catalog sweep: **99 pass / 0 fail**.
  - `test-bench-capture.sh` ran with NumPy supplied through `uv run --with numpy`.
  - `test-switch-explain.sh` ran with `CUDA_VISIBLE_DEVICES=0,1`, matching its dual-card default-resolution scenario. On the raw host, these two tests also fail on exact base `a6385ece` because NumPy is absent and four visible GPUs select a topology with no vLLM default.
- `diagnose-profile.sh vllm/qwen-agentworld-35b-a3b-dual-awq-int4`: **GREEN**.
- `tools/kv-calc.py --calibration`: **8/8 (100%)**.
- Registry/profile focused checks and locale functional test: PASS.

The rebase exposed one real branch-specific issue: adding the `qwen-agentworld-*` model id made `switch.sh --list` ordering locale-dependent. `7afc001f` pins the two list sorts to C collation; the ISO-8859-1 functional test now passes.

The served compose, calibration row, and model profile are byte-identical to the pre-rebase head that produced the existing measurements:

- [Final-config validation report from my 2×3090 rig](https://github.com/noonghunna/club-3090/pull/868#issuecomment-5161958260)
- [Independent 2×3090 reproduction](https://github.com/noonghunna/club-3090/pull/868#issuecomment-5198937068)

The GPUs aren't available for a fresh serving run right now. The existing final-config report covers verify-full, stress 8/8, bench, C=4 full-context concurrency, soak, and quick quality; the documented `rebench-full` path replaces the overlapping operational gates in `report.sh --full`.

Given the unchanged runtime artifacts plus the existing first-party and independent evidence, this is ready to merge. If you specifically need a newly generated `report.sh` bundle, I'll add it when the GPUs are available.

noonghunna:
Merged. Verification on the reference rig confirmed everything your rebase report claimed: full catalog sweep **99 pass / 0 fail** (no NumPy or GPU-count caveats here — clean on the raw host), all four count assertions land where derived, and the C-collation fix is exactly the right shape for the #779 locale class (nice catch — that bug predates your PR conceptually and your model id merely exposed it).

On the fresh `report.sh` offer: **not needed.** The serving artifacts are byte-identical to the head that carries your full gate run plus the independent 2×3090 reproduction — re-running the same bytes through the same gates adds ceremony, not evidence. The ✅ Production status is backed the way the repo's rule intends: by measurements from the rigs that made them.

That's four merged contributions now, each one arriving pre-validated — the catalog is measurably better for it. AgentWorld is live as of this merge.

Issue #777, opened 2026-07-26
[bug] submit-bench.sh unusable on non-UTF-8 locales — and the obvious encoding= fix truncates BENCHMARKS.md


noonghunna:
Found while fixing #776. Separate bug, separate fix — filing rather than folding it in, because the correct fix is not what it looks like.

## Symptom

`scripts/submit-bench.sh` cannot run at all on a rig with a non-UTF-8 locale. It dies decoding its own artifacts:

```
$ PYTHONUTF8=0 PYTHONCOERCECLOCALE=0 LC_ALL=C bash scripts/submit-bench.sh --tag <tag> --auto-submit --as-pr
Traceback (most recent call last):
  File "<stdin>", line 9, in <module>
  File "/usr/lib/python3.12/pathlib.py", line 1030, in read_text
    return f.read()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 27: ordinal not in range(128)
```

This is the #599 class — the same one [`AGENTS.md`](../blob/master/AGENTS.md) already has a standing rule for ("Any Python read of a repo source file — **including python heredocs inside shell scripts** — MUST pass `encoding="utf-8"`"). Nobody caught it here because, per #776, this whole path had never been executed outside the maintainer rig.

`PYTHONUTF8=0 PYTHONCOERCECLOCALE=0` are required to reproduce: modern Python coerces a bare `LC_ALL=C` to `C.UTF-8`.

## Four sites

| file:line | call | direction |
|---|---|---|
| `scripts/submit-bench.sh:120` | `Path(template).read_text()` | read (PR template) |
| `scripts/submit-bench.sh:127` | `Path(body_file).write_text(text)` | write (PR body) |
| `scripts/submit-bench.sh:185` | `path.read_text().splitlines()` | read (BENCHMARKS.md) |
| `scripts/submit-bench.sh:219` | `path.write_text("\n".join(lines) + "\n")` | write (BENCHMARKS.md) |

(The formatter's own two artifact reads plus its stdout are fixed in the #776 PR, and that PR adds a locale regression guard covering `bench_row_format`/`bench_row_section`. Everything above is untouched by it.)

## ⚠️ The obvious fix is a data-loss trap — read this before patching

Adding `encoding="utf-8"` to those four calls **does not fix it**, and fixing only the *reads* is actively dangerous.

[...]

PR #898, opened 2026-08-05
feat: auto-detect club-3090 endpoint for Claude Code

paulp83:
Auto-detect running club-3090 container, extract port + model name via /v1/models, and update ~/.claude/settings.json. Includes retry/backoff for boot race and a shareable README.


noonghunna:
Reviewed. This solves a real annoyance — `switch.sh` moves both the port *and* the served-model name, so `settings.json` goes stale on every switch — and it's a pure addition (+362/−0, three new files), so there's no regression surface on existing paths.

The engineering in `update-claude-settings.py` is careful, and I want to be specific about what's good since the same care is what makes the one gap stand out: `tempfile.mkstemp` + `os.replace` for an **atomic** write, cleanup on `BaseException`, `/v1/models` retry with backoff for the 10–20 s boot race, container-inspect fallback, and — the detail I liked most — it won't overwrite last-good settings without a valid model name.

## One real defect: the docs teach the unsafe pattern the code avoids

`scripts/claude-auto-settings.md` gives users a copy-paste snippet to install the `SessionStart` hook:

```python
with open(settings_path, "w") as f:
    json.dump(settings, f, indent=2)
```

`open(..., "w")` **truncates on open**. If `json.dump` raises — or the process dies mid-write — the user is left with a truncated or empty `~/.claude/settings.json`, i.e. their Claude Code config is gone rather than unchanged.

This is a house rule here, learned the hard way: our `AGENTS.md` records a case (#777/#780) where exactly this pattern took a file from 180,693 bytes to **0**. The guidance is to write a temp sibling and `os.replace()` — atomic, and a failed encode leaves the original intact.

Your `.py` already does it correctly. The fix is to make the doc snippet match it, or simply have the doc call the script instead of hand-rolling a second writer for the same file.

## Second, smaller: no backup

Nothing takes a copy of a pre-existing `settings.json` before the first write. Given this is a file the user may have hand-tuned, a one-line `shutil.copy2` to `settings.json.bak` on first run would be cheap insurance and would make the tool much easier to recommend.

## One thing that isn't mine to decide

This writes to `~/.claude/settings.json` — a user's editor configuration, which nothing else in this repo touches. Whether club-3090 ships a tool that mutates external config is a scope call for the maintainer rather than a technical one, so I've flagged it for that decision rather than sitting on the PR.

Nothing here is a rejection — the atomic-write fix and the backup are both small, and with them I'd be comfortable recommending it. Thanks for this, and for the DeepSeek fetch report earlier; both landed real fixes.

noonghunna:
Follow-up, and it supersedes most of my previous comment — the maintainer raised a design point I should have led with.

## The two values are just environment variables

The `env` block in `settings.json` exists to inject environment variables into the session. So this:

```json
{"env": {"ANTHROPIC_BASE_URL": "…", "ANTHROPIC_MODEL": "…"}}
```

is a persistent, file-mutating way to do what a shell export does transiently:

```bash
export ANTHROPIC_BASE_URL=http://localhost:8030
export ANTHROPIC_MODEL=deepseek-v4-flash
claude
```

Same two variables, same effect.

## Everything valuable in your PR is upstream of that choice

Reading `update-claude-settings.sh`, the work that's actually hard is the **detection**, and none of it cares where the values end up:

```
docker ps → container id → port binding "0.0.0.0:8077->8000/tcp"
          → HOST_PORT / HOST_IP → BASE_URL
          → curl /v1/models  (retry + backoff for the 10-20s boot race)
          → MODEL_NAME,  with docker-inspect Config.Cmd as fallback
```

That's the part worth shipping. Writing the result into `settings.json` is the last couple of lines — and it's the only part that creates problems.

**Suggestion: emit the exports instead of writing the file.**

```bash
eval "$(bash scripts/club3090-env.sh)" && claude
```

with the script printing:

```
export ANTHROPIC_BASE_URL=http://localhost:8030
export ANTHROPIC_MODEL=deepseek-v4-flash
```

## What that removes

| concern from my earlier review | with the eval form |
|---|---|
| mutating `~/.claude/settings.json` | **gone** — nothing outside the repo is touched |
| non-atomic write in the README snippet | **deleted** — no hook to install |
| no backup before first write | **moot** |
| unguarded `json.load` on a missing/corrupt file | **moot** |
| "is this in scope for club-3090?" | **dissolved** |

Every review point I raised was about the writer, not the detector. Dropping the writer drops all of them at once.

## It's also a better tool

Per-shell rather than global, so two terminals can point at two different composes — which is a normal thing to want on a rig that runs several. And it composes with anything that reads those variables, not just Claude Code.

The one thing you'd give up is automatic refresh on `SessionStart`. That's genuine convenience, but it's precisely the feature that requires the hook, the config write, and the risk. A shell alias — `alias c3claude='eval "$(bash …/club3090-env.sh)" && claude'` — recovers most of it for none of the cost.

**Keep the detection, drop the writer.** Smaller PR, better tool, and no maintainer scope decision needed. If you'd rather keep the settings.json path as an opt-in flag (`--write-settings`) with the atomic write you already have, that also works — the important part is that it stops being the default.

Your call on shape; happy to review either.

paulp83:
---
  Updated per feedback — keeping the detection logic, dropping the file-writer entirely.

  What changed:

  - update-claude-settings.sh now prints export ANTHROPIC_BASE_URL=... and export ANTHROPIC_MODEL=... to stdout (safe to eval), with diagnostics to stderr. No files are written.
  - Deleted update-claude-settings.py — no more Python dependency for the write path.
  - Rewrote docs: alias pattern (alias club3090-env='eval "$(bash ...)"') replaces the hook-based settings.json install.

  Why: As noted, these are just two environment variables. The eval pattern is idiomatic shell, zero file corruption risk, and removes the scope question of modifying an external editor's config.

  Usage:
  # One-time: add to ~/.bashrc or ~/.zshrc
  alias club3090-env='eval "$(bash scripts/update-claude-settings.sh)"'

  # Then:
  bash scripts/switch.sh vllm/dual
  club3090-env    # sets env vars
  claude          # connects to the right endpoint

noonghunna:
Merged — thank you for how you handled both revision rounds; the eval redesign came out exactly right.

Before merging I live-tested your branch on the reference rig against a running llama.cpp DeepSeek container: stdout was precisely the two export lines (`http://localhost:8031` / `deepseek-v4-flash`, both correct), diagnostics clean on stderr, exit 0. First try.

One maintainer touch-up pushed to your branch pre-merge: renamed to `scripts/club3090-env.sh` + `club3090-env.md` — with the settings-writer gone, nothing 'updates settings' anymore, and renaming after people bake the path into shell aliases would have been disruptive. No behavior change.

A few polish items we'll fold into a follow-up on master so you don't need another round: the `'inst'` alias filter (latent — nothing in today's catalog matches, but it would silently exclude a future `*-instruct` id), `printf %q` on the emitted values for belt-and-braces eval safety, the hyphen handling in the container-Cmd fallback regex, and two doc nits (the dry-run grep reads stdout where diagnostics are on stderr; the 'Hook timeout' row is from the deleted design).

PR #928, opened 2026-08-07
docs(p2p): decode gain is the KERNEL and we auto-enable it; add the engine split

noonghunna:
#925 said the decode delta *"sits inside run-to-run noise."* That is true **only with vLLM's custom all-reduce forced off** — which is not what we ship.

## 1. Wrong for the shipped path

`detect_nvlink.sh` sets `_NVLINK_ENABLED=1` for `pcie_p2p`, and the compose then **omits** `--disable-custom-all-reduce`. So any rig with patched P2P runs the kernel **by default**.

Three rigs, once the kernel is accounted for, stop disagreeing:

| rig | custom AR | decode |
|---|---|---|
| @juslex 2×3090 | **OFF** (forced) | inside noise |
| @henrykrinkle01 2×3090 | **ON** (our default) | **+12.5% / +7.1%** |
| @mgabor3141 4×3090 | ON | **+7.7% / +8.9%** |

**DECODE = the kernel. PREFILL = the transport.** Our own rig reports `CNS` and takes the safe branch — which is exactly why the earlier text read the way it did.

## 2. Engine matters, and it was missing entirely

Every figure on that page is vLLM **tensor-parallel**. On llama.cpp / ik-llama `-sm layer`:

```
decode +0.2%   prefill ~+1%     (disc #921)
```

Layer split forwards activations with plain copies — **there is no all-reduce for a peer path to accelerate.** A GGUF-only reader would have rebuilt a DKMS module on every driver bump for nothing. Now stated plainly.

## 3. The wrong-data warning no longer asserts a cause

It claimed the custom kernel returns wrong data. @henrykrinkle01 found [#751](https://github.com/noonghunna/club-3090/issues/751) — **same `!!!!!` signature with P2P off**, resolved by reseating cables — and two patched rigs run the kernel cleanly. Now stated as **unsettled**, with both candidates and the one-line test. A default change is **on hold** pending a reseat + before/after `LnkSta` from #922.

## Also: three measured rows added to §6

4-card TP=4 (**+74.7% / +84.9% prefill** — the transport gain scales with world size), llama.cpp layer split (~0), and the shipped-path decode figure.

## The guard was pinning the claim again

`test-p2p-state` asserted `"INSIDE"` — the exact trap it hit last time, when it asserted `"+2% narrative"`. **Twice now this test has held a misleading claim in place.** It now asserts `AUTO-ENABLE`, `ENGINE MATTERS` and `UNSETTLED`.

Hold the advisory to being **honest**, not to being **unchanged**.

Guards green: `test-p2p-state`, `test-classifier`, `test-locale-utf8`, `test-report-exit-code`.


henrykrinkle01:
Data shared on Discord, copy here. 
Qwen 3.6 27b Q8, MTP n=2 on llamacpp. Prefill is @10k

| Mode | P2P | Narrative wall_TPS | Code wall_TPS | Prefill Speed (tok/s) |
| :--- | :--- | ---: | ---: | ---: |
| **TP=2** | **Off** | 61.52 | 73.48 | 1313.42 |
| **TP=2** | **On** | 64.23 | 72.23 | 1345.68 |
| **PP=2** | **Off** | 43.05 | 50.43 | 1544.78 |
| **PP=2** | **On** | 42.42 | 49.32 | 1548.74 |

Almost no benefit between p2p on or off even with `-sm tensor` (TP=2). Prefill only barely improves. Also the llama-server boot logs say:
```
NCCL not compiled in; falling back to internal AllReduce.  Recompile with -DGGML_CUDA_NCCL=ON for best multi-GPU performance.
```
So probably the default llama image doesn't take advantage of p2p without NCCL compiled in

noonghunna:
That boot line is the mechanism, and it makes my explanation incomplete. Pushed a refinement crediting it.

I'd written that llama.cpp gains nothing because layer split forwards activations with plain copies — no all-reduce, nothing to accelerate. True for `-sm layer` and PP. **But it doesn't cover `-sm tensor`, which does all-reduce**, and a reader given only my reason would reasonably try tensor split expecting it to help. You tested exactly that and it didn't.

Your log says why, and I verified it on our own pinned image:

```
$ strings /app/llama-server | grep -c "ncclAllReduce\|GGML_CUDA_NCCL"   → 0
$ ldd /app/llama-server | grep -ci nccl                                 → 0
```

**Zero NCCL symbols, zero linkage in `server-cuda-b10236`.** So the internal AllReduce fallback is the only path, and it doesn't use peer transfers — which is why your tensor-split arm is noise rather than a win.

So there are two independent reasons, and the second is the load-bearing one because it closes the obvious workaround:

| split mode | your numbers | why |
|---|---|---|
| `-sm layer` / PP | +0.2% decode, ~+1% prefill (disc #921) | plain copies — no all-reduce at all |
| `-sm tensor` TP=2 | 61.5→64.2 narr, 73.5→72.2 code, 1313→1346 prefill | all-reduce exists, but **no NCCL in the image** |

Both now in the doc as a table with your boot line quoted, so anyone can check their own image in one command.

The guidance also changes from flat to conditional: patched P2P isn't worth a DKMS rebuild for a GGUF-only user **unless they're also prepared to rebuild llama.cpp with `-DGGML_CUDA_NCCL=ON`** — which I've stated as *unmeasured*, because nobody has. If you ever feel like building that, it's the one open question in this corner: does a NCCL-enabled llama.cpp actually pick up the peer path, and does it approach vLLM's prefill numbers?

Two smaller things worth noting from your table, since you may not have been looking for them:

**PP=2 prefills faster than TP=2** (1544 vs 1313) while decoding much slower (43 vs 61). That's a real tradeoff and it's not documented anywhere on our side — pipeline parallel avoids the per-token all-reduce entirely but serialises decode across stages. If you have a prefill-heavy workload that's a genuine lever.

**Your code decode went *down* slightly with P2P on** in both modes (73.5→72.2, 50.4→49.3). Small enough to be noise, but it's consistent across two split modes, which is mildly interesting — if P2P is doing nothing useful here, any measurable effect should be zero-mean rather than one-sided. Not worth chasing unless it reproduces.

Thanks — this is the third time today your questions have changed something we ship.

PR #937, opened 2026-08-09
feat(residency): additive auto-sizer calibrated from #931 — replaces the ×0.55 guard

noonghunna:
Closes the calibration arc from #931.

## What

`fit = (FREE − reserve − first_card_extra[card 0] − margin) / bundle`

The ×0.55 multiplicative guard absorbed an **additive** overhead (dense split + drafter half + compute buffers + KV — none of which scales with card size). It was correct on the 24 GB cards it was calibrated on and left ~6 GB/card idle on 32 GB cards — community-measured at **+19% decode** (#931).

## Constants, all field-derived

| constant | value | source |
|---|---|---|
| reserve (per compose) | Q8-dual 18000 (unchanged) · IQ2 14000→16500 · multi4 12000→14500 | re-derived; engine cost measured 16543 MiB (bare card) / 17201 (card 0) on 2×5090 |
| first-card extra | 768 | card 0's larger drafter half + compute buffer (measured 660 MiB asymmetry) |
| margin (`RESIDENCY_MARGIN_MB`) | 1024 | #931 bracket: 556 MiB free **died** at ~90K depth; 896 MiB **survived** a 188K NIAH ladder |

Sizing now reads **detected free** VRAM, so desktops and other consumers fall out automatically and asymmetric per-card counts need no special-casing.

## Reproduces every measured point

Q8 1/card + IQ2 3/card on 2×24 GB · @milano's 2/card on 4×24 · 0 bundles on 16 GB cards · **#931's asymmetric 7+8 on 2×32 GB** — the config field-validated end-to-end (prefill-90K ×3 clean, NIAH to 188K, soak-stable VRAM). Deliberately floor-conservative on borderline cards; `OT_G<i>` overrides (user-wins, gate-priced) remain the escape hatch.

## Tests

`test-preflight-cpu-offload`: fit units rewritten to the additive model, new legs for the 32 GB asymmetric case (must produce 7+8 from identical free readings) and the margin override — 61 checks ok. Status-drift + both registry-parity gates green. IQ2 registry note now cites #931 as first community validation.


noonghunna:
Live-boot validation on the reference rig (2×3090, IQ2 slug via `switch.sh` from this branch):

```
[residency] card0: auto 3 bundles (blk 0|1|2) · card1: auto 3 bundles (blk 42|41|40)
[preflight] cpu-offload: RAM ok — needs ~76 GB on this rig (86 GB worst case, residency keeps ~10 GB of experts on GPU)
[switch] ✓ ready (40s)
```

Same grant as the field-validated counts, gate arithmetic coherent, server healthy.

One trade-off to note for the record: sizing from **free** VRAM (the whole point) introduces a timing sensitivity the total-based formula didn't have — if a previous container's teardown is slow to release VRAM (we've seen kills land ~60 s late on huge-RSS servers), the free reading lands low and the sizer under-grants for that boot. The degradation is in the SAFE direction (fewer bundles, gate prices accordingly), and a re-`switch` fixes it, but a wait-for-free-to-settle loop is a candidate follow-up if it bites anyone in practice.

PR #947, opened 2026-08-10
docs(club3090-env): add Claude Code settings.json configuration instructions

paulp83:
## Summary

This PR adds important documentation to `scripts/club3090-env.md` about configuring `~/.claude/settings.json` to allow the `club3090-env.sh` script to work correctly with Claude Code.

## Changes

- Add important note about Claude Code picking up shell environment variables.
- Specify that the `~/.claude/settings.json` `env` section should only contain `ANTHROPIC_API_KEY: "dummy"`.
- Add a detailed section explaining why `ANTHROPIC_BASE_URL` and `ANTHROPIC_MODEL` must not be hardcoded in `settings.json`.
- Provide instructions on how to fix `settings.json` if it has hardcoded values, using generic placeholders.

## Background

When `~/.claude/settings.json` has hardcoded `ANTHROPIC_BASE_URL` and `ANTHROPIC_MODEL` values in its `env` section, Claude Code uses those hardcoded values and ignores the shell environment variables set by `club3090-env.sh`. Additionally, the `env` section must contain `ANTHROPIC_API_KEY: "dummy"` to prevent authentication errors when working with local endpoints.

The updated documentation ensures users configure their Claude Code settings correctly to allow dynamic endpoint detection via the `club3090-env.sh` script, and uses generic placeholders in examples to avoid exposing local rig information.

### DIFF
--- a/scripts/club3090-env.md
+++ b/scripts/club3090-env.md
@@ -4,6 +4,8 @@
 
 When you switch composes with `switch.sh`, every config changes the port **and** the model name. This tool auto-detects the running endpoint and prints `export` lines so your shell always points Claude Code at the right model — zero manual editing.
 
+> **Important Claude Code setup:** For Claude Code to pick up the shell environment variables (`ANTHROPIC_BASE_URL` and `ANTHROPIC_MODEL`), your `~/.claude/settings.json` file must **not** have hardcoded `ANTHROPIC_BASE_URL` and `ANTHROPIC_MODEL` in its `env` section. The `env` section should **only** contain `ANTHROPIC_API_KEY: "dummy"` to prevent authentication errors, while allowing the shell environment variables to provide the URL and model name.
+
 ## How it works
 
 ```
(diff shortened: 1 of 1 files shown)


noonghunna:
Merged — thank you @paulp83.

This fills a gap our own design creates and never explained. `club3090-env.sh` states plainly that it *"writes no files"* and *"touches no settings.json"* — which is the right contract for something you `eval` — but it means a user with `ANTHROPIC_BASE_URL` / `ANTHROPIC_MODEL` hardcoded in `~/.claude/settings.json` gets their exports silently overridden. The symptom is *"I switched composes and Claude Code is still hitting the old endpoint"*, with nothing in the repo pointing at the cause. Nothing else here even mentions `settings.json`, so there was no thread to pull.

Also appreciated: the generic placeholders. Setup docs are where rig-specific ports and model names usually leak in, and you pre-empted it.

PR #949, opened 2026-08-11
Collapse repeated `-ot` into one comma-separated arg (fixes #948)

### LINKED ISSUE #948: [bug] Repeated `-ot` silently drops GPU expert pinning on 3 DeepSeek composes (only last value is used)
Three shipped DeepSeek-V4-Flash composes pass `-ot` more than once. Current llama.cpp honours **only the last one**, so the GPU expert-pinning rules are silently discarded and every expert falls through to the CPU rule.

Surfaced by a community question on Discord asking whether the `-ot`/`-ts` values were hand-tuned for performance — they are neither hand-tuned nor, as it turns out, applied.

### Affected

| Compose | `-ot` count |
|---|---|
| `models/deepseek-v4-flash-0731/llama-cpp/compose/multi4/unsloth-q8-kxl/offload.yml` | 5× |
| `models/deepseek-v4-flash-0731/llama-cpp/compose/dual/unsloth-q8-kxl/offload.yml` | 3× |
| `models/deepseek-v4-flash-0731/llama-cpp/compose/dual/unsloth-iq2-xxs/offload.yml` | 3× |

### Reproduction

Against the pinned image (`ghcr.io/ggml-org/llama.cpp:server-cuda-b10236`), no GPU needed:

```bash
docker run --rm --entrypoint /app/llama-server \
  ghcr.io/ggml-org/llama.cpp:server-cuda-b10236 \
  -m /nonexistent.gguf \
  -ot 'blk\.0\.ffn_exps\.weight=CPU' \
  -ot 'blk\.42\.ffn_exps\.weight=CPU' \
  -ot 'blk\.[0-9]+\.ffn_exps\.weight=CPU'
```

```
W DEPRECATED: argument '-ot' specified multiple times, use comma-separated values instead (only last value will be used)
W DEPRECATED: argument '-ot' specified multiple times, use comma-separated values instead (only last value will be used)
```

One warning per discarded rule.

### Why this is worse than a dead default

The `blk.99999` values in the file are intentional no-op sentinels, so a bare `docker compose up` is *supposed* to degrade to all-experts-CPU. That part is by design.

The problem is the launcher path. `OT_G0` / `OT_G1` are **injected by the launcher**, resolved from the detected card's VRAM (a 32 GB card holds more layers than a 24 GB one). Those computed values are passed as the first two `-ot` arguments — and then discarded, because a third `-ot` follows them. **The per-card residency policy never takes effect under `switch.sh` either.**

### Impact

- The interleaved dual-GPU expert placement these composes describe has never actually been in force.
- `--n-cpu-moe` cannot balance experts across two GPUs, which is the entire reason these composes use `-ot`; with the rules dropped, that capability is lost silently.
- ⚠️ Benchmarks taken on these composes are still valid *as measurements*, but they characterise **all-experts-on-CPU**, not the interleaved placement the file appears to specify. Any row presented as "interleaved placement" needs re-checking.

The only symptom is a `DEPRECATED` line in the boot log, which is easy to read past.

### Fix

Collapse to a single `-ot` with comma-separated rules, preserving order (earlier rules win, so the CUDA pins must precede the CPU catch-all):

```yaml
- '-ot'
- '${OT_G0:-…=CUDA0},${OT_G1:-…=CUDA1},blk\.[0-9]+\.ffn_(gate|up|down)_exps\.weight=CPU'
```

This is the shape the newer `llamacpp-club3090` moe-cache composes already use, which is why they are unaffected.

⚠️ This is a **behaviour change, not a no-op**: it turns the pinning on for the first time, which shifts VRAM residency and performance. It needs a live boot plus a before/after check, not a blind edit.

### Follow-up worth considering

A guard test asserting no compose passes the same repeatable-but-last-wins argument more than once. This class fails silently and the existing suite has no way to catch it.

noonghunna:
Fixes #948.

`llama.cpp` honours only the **last** `-ot` when the flag is repeated. Three DeepSeek composes passed it 3–5 times, so every GPU expert-pinning rule was silently discarded and all experts fell through to the trailing CPU rule.

### Change

One `-ot` with comma-separated rules, order preserved (earlier rules win, so the CUDA pins must precede the CPU catch-all):

```yaml
- '-ot'
- '${OT_G0:-…=CUDA0},${OT_G1:-…=CUDA1},blk\.[0-9]+\.ffn_(gate|up|down)_exps\.weight=CPU'
```

| Compose | before | after |
|---|---|---|
| `multi4/unsloth-q8-kxl/offload.yml` | 5× `-ot` | 1× |
| `dual/unsloth-q8-kxl/offload.yml` | 3× | 1× |
| `dual/unsloth-iq2-xxs/offload.yml` | 3× | 1× |

### Verified

Against the pinned image `ghcr.io/ggml-org/llama.cpp:server-cuda-b10236`:

- **Repeated form** → one `DEPRECATED: argument '-ot' specified multiple times ... (only last value will be used)` per discarded rule.
- **Collapsed form** → parses clean; no warning, no error, normal startup. The `(gate|up|down)` alternation survives the comma split (the regexes contain no commas).
- `docker compose config` with `OT_G0`/`OT_G1` injected renders all three rules in the correct order in a single argument.
- All 9 `test-compose-*` guards pass.

### ⚠️ Not yet validated — do not merge on the green checks alone

**This is a behaviour change, not a no-op.** It turns the expert pinning on for the first time, which shifts VRAM residency and performance on every affected compose.

Still outstanding:

1. **Live boot** of each compose with a real `OT_G*` injection, confirming the pinned layers actually land on the intended devices (per-device buffer sizes at `-lv 4`).
2. **Before/after bench**, because VRAM residency changes — the fixed config may need its layer counts re-tuned, since the current `OT_G*` resolver was calibrated against a config where its output was being thrown away.
3. **Re-check any BENCHMARKS row** attributed to these composes. Those numbers are valid measurements, but they characterise *all-experts-on-CPU*, not the interleaved placement the files describe.

Boot validation is blocked on GPU availability (a long quality run is occupying both cards). Opening as a draft so the diff and the reproduction can be reviewed meanwhile.

### Follow-up

Worth a guard asserting no compose repeats a last-wins argument. This class fails silently — the only symptom was a `DEPRECATED` line in the boot log, and nothing in the suite reads boot logs.


noonghunna:
Guard added to this branch (`4d3c0d0`), so the fix and the thing that prevents its recurrence land together.

`scripts/tests/test-compose-no-repeated-args.sh` scans every compose `command:` block and fails on any flag token appearing more than once.

Verified **both** directions rather than just the happy path:

- **on this branch** — `ok (62 composes, no repeated flags)`
- **on the unfixed tree** — fails with exactly the three known files and nothing else:

```
FAIL: .../dual/unsloth-iq2-xxs/offload.yml passes -ot 3 times
FAIL: .../dual/unsloth-q8-kxl/offload.yml passes -ot 3 times
FAIL: .../multi4/unsloth-q8-kxl/offload.yml passes -ot 5 times
checked=62 fails=3
```

`-ot` is the only flag repeated anywhere in the repo today, so the allowlist ships **empty** — every repeat found so far has been a bug. If a genuinely additive flag turns up, the test asks for the engine and a link proving the semantics before it can be added.

This moves the item out of "Follow-up" in the description above. The three validation gates listed there (live boot, before/after bench, BENCHMARKS re-check) still stand — they need GPUs that are currently busy.

noonghunna:
## Boot validation — the bug reproduces, the fix eliminates it

Ran both arms on 2× RTX 3090, same rig, same session, **same `OT_G*` values injected into both**, so the only variable is the compose's `-ot` form.

The launcher's residency resolver produced real layers, not the `blk.99999` sentinel — and reproduced the documented `Q8-dual on 2×24 GB → 1 bundle/card` calibration point exactly:

```
[residency] card0: auto 1 bundles (blk 0) · card1: auto 1 bundles (blk 42)

OT_G0=blk\.(0)\.ffn_(gate|up|down)_exps\.weight=CUDA0
OT_G1=blk\.(42)\.ffn_(gate|up|down)_exps\.weight=CUDA1
```

### Result

| | Arm A — unfixed (master) | Arm B — fixed (this PR) |
|---|---|---|
| `-ot` args reaching the engine | **3** | **1** |
| `DEPRECATED: specified multiple times` | **2** | **0** |
| CUDA pins | **discarded** | **applied** |
| Boots + serves | ✅ | ✅ |
| `bench.sh` swap guard | — | PASS |

The engine receives the collapsed rule intact, order preserved:

```
-ot blk\.(0)\.ffn_(gate|up|down)_exps\.weight=CUDA0,blk\.(42)\.ffn_(gate|up|down)_exps\.weight=CUDA1,blk\.[0-9]+\.ffn_(gate|up|down)_exps\.weight=CPU
```

**This confirms the mechanism in #948 end to end.** The residency machinery was computing *correct* values all along — `blk 0` to card 0, `blk 42` to card 1, each card drawing from its own `-sm layer` range — and then the third `-ot` silently discarded both. Not "residency is broken" but "residency was computed correctly and thrown away at the last step."

VRAM comparison (does the pin *place*, not merely *parse*) is in flight; arm B measured **23,392 / 22,954 MiB** and arm A is loading now. Will append.

### ⚠️ Perf delta: NOT measured, and I'd rather say so than publish a bad number

I booted these with raw `docker compose` rather than `switch.sh`, so `THREADS` fell back to the compose's deliberate bare-compose default of **8** instead of the launcher-injected value. Decode came out at **8.31 t/s** against a canonical **21.91** on this hardware — the under-threading dominates everything else.

An A/B at that operating point would say nothing useful about two layers of expert residency, and the bias isn't neutral: at `-t 8` the CPU is a harder bottleneck, so moving expert compute onto the GPU flatters the fixed arm. A real comparison needs both arms through `switch.sh` at the tuned `-t`.

**Worth stating plainly: throughput was never the argument for this PR.** 1 of 43 layers per card is a small share of expert traffic, and a delta inside noise would be an entirely legitimate result. The case for merging is that three shipped composes describe a residency policy the engine silently discards — including under `switch.sh`, where the launcher computes per-card values and has them thrown away.

### Still outstanding before merge

- [ ] VRAM placement confirmation (arm A loading)
- [ ] Perf A/B at tuned `-t`, both arms via `switch.sh`
- [ ] Re-check any BENCHMARKS row attributed to these composes — those numbers are valid measurements but characterise **all-experts-on-CPU**, not the interleaved placement the files describe

noonghunna:
## ⚠️ Correction — VRAM says the pins parse but do **not place**. Do not merge on the evidence above.

Arm A has finished loading. Measured at the same lifecycle point (freshly serving, before any bench load):

| | Arm A — unfixed (pins discarded) | Arm B — fixed (pins applied) |
|---|--:|--:|
| VRAM card 0 | **23,280 MiB** | **23,280 MiB** |
| VRAM card 1 | **22,838 MiB** | **22,838 MiB** |

**Byte-identical.** (Arm B later read +112 MiB/card, but that appeared only after the bench started — compute-buffer growth, not residency. I initially misread it as a placement signal.)

A Q8 expert bundle on this model is **3.19 GiB**. If `blk.0 → CUDA0` and `blk.42 → CUDA1` were placing, each card should gain roughly that. Neither did.

### What this rules out

- **Rule ordering is not the cause.** `llama-model-loader.cpp` breaks on the **first** matching override, so the CUDA pins preceding the CPU catch-all is correct:
  ```cpp
  if (std::regex_search(tensor_name, pattern)) { ... break; }
  ```
- **The architecture is not the cause.** GGUF metadata: `block_count=43`, `expert_count=256`, `expert_used_count=6`, and no `leading_dense_block_count` — all 43 layers are MoE, so `blk.0` does carry `ffn_*_exps` tensors.
- **Injection is not the cause.** Both arms received identical `OT_G*` values, and the fixed arm's single `-ot` reaches the engine intact with the pins first.

### Where that leaves the PR

The **argument-level fix is still correct and still worth having** — 3 `-ot` args become 1, the `DEPRECATED … only last value will be used` warning goes from 2 to 0, and the launcher-computed values stop being silently discarded. That part is verified.

But the *motivating claim* — that fixing this restores interleaved expert residency — is **not demonstrated**. Placement is unchanged, so on current evidence this is a correctness/hygiene fix, not a residency fix. #948's impact section overstates the consequence and I'll amend it.

### Next step

Boot with debug verbosity and read the loader's own override trace:

```
tensor %s (%zu MiB %s) buffer type overridden to %s
```

That line fires per overridden tensor and will say definitively whether the CUDA rules match anything. Until it does, I'd hold this PR — merging it would close #948 while quietly leaving residency just as inert as before, which is the same class of silent failure the issue is about.

noonghunna:
## ✅ Correction to my correction — the pins **do** place. #949 is a real residency fix.

My previous comment concluded "pins parse but do not place". **That was wrong**, and it was wrong because I compared against a baseline that differed in more than one variable. Properly controlled, with the drafter removed from both arms and everything else identical:

| Config (no drafter, otherwise identical) | Card 0 | Card 1 |
|---|--:|--:|
| Real pins — `blk.(0)→CUDA0`, `blk.(42)→CUDA1` | 14,172 MiB | 14,684 MiB |
| Sentinel pins — `blk.99999` (control) | 10,920 MiB | 11,432 MiB |
| **Delta** | **3,252 MiB** | **3,252 MiB** |

A Q8 expert bundle on this model is **3,264 MiB**. Both cards land within 12 MiB of exactly one bundle, and the two deltas agree to the megabyte. The mechanism works.

## So why did the original A/B show no difference?

Because there was nowhere to put it. **These composes park the draft model on the GPUs** — no `-devd` flag, so it takes the default — and that costs:

| | Card 0 | Card 1 |
|---|--:|--:|
| pins + drafter on GPU | 23,280 MiB | 22,838 MiB |
| pins + no drafter | 14,172 MiB | 14,684 MiB |
| **drafter footprint** | **9,108 MiB** | **8,154 MiB** |

That independently reproduces the `~8.9 / 8.1 GB` figure from #931 on different hardware — so that analysis was correct, and I was wrong to doubt it above.

With the drafter resident there is **~1.3 GB free per card** against a 3.26 GB bundle. The pin arrives and cannot be honoured.

## Revised conclusion: two stacked defects

1. **#948, parse level** — three `-ot` args, only the last survives, pins discarded before the engine ever considers them. **This PR fixes it, verified: 3 args → 1, warnings 2 → 0.**
2. **Allocation level, NEW** — the sizer runs against *free* VRAM at resolve time, **before the ~9 GB drafter loads**, so it grants a bundle that cannot fit. Fixing #948 alone just moves the failure one layer down.

**#949 is necessary but not sufficient.** It should merge — it is a genuine residency fix and the guard prevents recurrence — but it will not by itself make residency effective on the Q8 dual compose at this context. I'll open a separate issue for the sizing defect.

⚠️ Also found while testing: **`-devd none` segfaults stock `b10236`** (exit 139, in `common_speculative_init_result → llama_context → sched_reserve`). So "just move the drafter to CPU" is not available as a fix on the stock engine, though our fork handles it. Filing separately.

noonghunna:
Out of draft. Boot validation is complete and the finding held up — with one scope correction now folded in.

**Verified by this PR:**
- 3 `-ot` args → 1; `DEPRECATED … only last value will be used` warnings 2 → 0
- The launcher's computed pins reach the engine intact, order preserved
- **The pins are meaningful**: given headroom they place exactly one bundle per card — measured **3,252 MiB** against a documented 3,264, both cards agreeing to the megabyte

That last point is what makes this a fix rather than tidy-up. Discarding those rules was a real loss of function, not a no-op.

**Explicitly out of scope**, now tracked separately:
- #953 — the residency sizer grants bundles that cannot fit, because it measures free VRAM before the ~9 GB drafter loads. Merging this PR does **not** make residency effective on the Q8 dual compose at 200K; it removes the first of two stacked blockers.
- #954 — `-devd none` segfaults stock `b10236`, which closes off the obvious remedy for #953 on the stock engine.

**Perf A/B is deliberately not included.** Both arms ran at the bare-compose `-t 8` fallback (8.31 t/s vs 21.91 canonical), where under-threading dominates and the bias runs toward the fixed arm. A number from that operating point would be misleading, and throughput was never the argument here.

Ready for review.

PR #970, opened 2026-08-12
docs: restructure SINGLE_CARD + IK_LLAMA after the single-card retirements

### LINKED ISSUE #967: docs: SINGLE_CARD.md and engines/IK_LLAMA.md need a restructure after the 2026-08-12 single-card retirements
Follow-up to #966, which retired every llama.cpp + ik-llama single-card `qwen3.6-27b` slug and removed `ik-llama` from all `ENGINE_PREFERENCE` walks.

#966 did a deliberate **correctness-only** doc pass — retirement banners on the two affected guides, plus repointing the active routing advice in README / GETTING_STARTED / FAQ / QUANTIZATION / WSL_SETUP / BRING_YOUR_OWN. That stops the docs from sending people to `--force`-only slugs, but it leaves both guides structurally stale behind their banners.

## What still needs doing

**`docs/SINGLE_CARD.md`** (~33 slug references) is a front-door guide whose entire recommendation structure — the "simplest vs fastest" framing, the config comparison table, the pick-by-workload decision path — is built on slugs that now need `--force`. It needs restructuring around what's actually functional single-card, which today is `vllm/minimal` at 32K with no vision.

⚠️ The honest version of that page has to say something uncomfortable: **single-card `qwen3.6-27b` no longer has a Cliff-2-immune path.** The old guidance routed accumulating-context agent traffic to `llamacpp/default` precisely because single-card vLLM is unsafe past ~21-26K accumulated context, and `vllm/minimal` *is* single-card vLLM. That's a real gap in the recommendation, not just stale prose.

**`docs/engines/IK_LLAMA.md`** documents an engine with zero functional slugs. The measurements in it are correct and worth keeping (IQ4_KS measured ~18-20% faster than `llamacpp/mtp` at matched 370 W) — the question is what an engine deep-dive should look like when the engine is out of every recommendation walk. Compare how the beellama retirement (2026-07-27) left `docs/INFERENCE_ENGINES.md`.

**The two SVG charts** (`docs/img/performance.svg`, `docs/img/performance-single.svg`) plot retired slugs as current recommendations. Regenerate via `tools/charts/gen-perf.py` once the story settles — note the markdown embeds the `.png` exports, so both need rebuilding.

**Lower-priority stragglers** that only carry passing references: `CLIFFS.md` (10), `COMPARISONS.md`, `COMPOSE_GENERATOR.md`, `EXAMPLES.md`, `HARDWARE.md`, `MULTI_CARD.md`, `PULL_GATE.md`, `STRUCTURED_COT.md`, `engines/SGLANG.md`, `INFERENCE_ENGINES.md`.

## Explicitly out of scope

Measurement tables and benchmark history stay as-is. Those numbers are real and are the historical record — only the *recommendations* are superseded. `BENCHMARKS.md` is append-only and should not be rewritten.

## Resolution trigger

If a functional long-context single-card qwen path ships on any engine, much of this reverses: re-add the engine to `ENGINE_PREFERENCE`, restore the `DEFAULTS` row, reconsider `RECOMMENDED_DEFAULT_MODELS`, and the guides go back to recommending it. Worth checking that before investing in a large rewrite.

noonghunna:
Closes #967. The restructure #966 deferred.

#966 did a deliberate correctness-only pass — banners plus repointed routing advice — which stopped the docs sending people to `--force`-only slugs but left both guides structurally stale behind those banners. This does the actual rewrite.

## `SINGLE_CARD.md` now opens with what runs

Three functional slugs on a single RTX 3090, all vLLM:

| Model | Slug | Max ctx | Vision | Status |
|---|---|--:|---|---|
| **Gemma-4-12B** ⭐ | `vllm/gemma-12b-single-int8-mtp` | **256K** | multimodal arch, served text-only | ⚠️ caveats |
| Gemma-4-26B-A4B | `vllm/gemma-26ba4b-single` | 176K | off | ⚠️ caveats |
| Qwen3.6-27B | `vllm/minimal` | **32K** | ❌ none | ✅ production |

⭐ **The headline: with one 3090 and a long-context need, you no longer run Qwen — you run Gemma-4-12B.** Qwen's single-card ceiling went 200K → 32K and its vision path is gone. The page had been built entirely around Qwen; that is no longer where the capability is.

## Two things the page can now say that it previously couldn't

**The Cliff-2 escape is gone.** The old guidance routed accumulating-context agent traffic to `llamacpp/default` *precisely because* llama.cpp has no DeltaNet GDN cliff. `vllm/minimal` **is** single-card vLLM — so there is no cliff-immune Qwen path on one card any more. Rather than quietly dropping the recommendation, the page now gives three honest answers: two cards (`vllm/dual`), a non-Qwen3-Next model (the Gemma slugs don't carry the cliff), or `--force` a retired llama.cpp slug — listed last, and labelled unsupported rather than pretended away.

**The NVFP4 slugs are arch-gated and unusable here.** `vllm/qwen-27b-single-nvfp4` (64K) and `vllm/qwen-35b-a3b-single-nvfp4` (128K) both declare `required_sm=9.0`; a 3090 is **sm 8.6**. Those are Hopper/Blackwell configs. On a page whose entire subject is a single 3090, that was never called out — a reader scanning for "single-card 64K Qwen" would have found one that cannot run.

## What was deliberately kept

The retired lineup — its TPS table, VRAM figures, per-config tuning rationale, and both cliff write-ups — is **preserved**, retitled as historical and marked per config. Nothing measured was deleted; only the recommendations changed. `BENCHMARKS.md` is untouched.

## `engines/IK_LLAMA.md` becomes a reference document

A status box replaces the interim banner: no functional slug remains, removed from every `ENGINE_PREFERENCE` walk, `ik-llama/default` now errors — but every slug stays `--force` launchable, c3 shows them via `--all`, and the digest-pinned image is untouched.

⭐ Its **~18–20% matched-power finding is explicitly preserved and flagged as still valid.** The engine was not retired for being slow; its qwen slugs went when the catalog surface consolidated onto vLLM. Those are different decisions and the page now says so, because that measurement was itself the subject of a correction earlier in the day and should not be quietly buried by a retirement.

## Charts

Retired configs are marked with a dagger in `gen-perf.py` and regenerated. The measurements stay on the chart; the marker stops it reading as a current recommendation.

⚠️ Worth noting for anyone touching these: my first attempt appended `— RETIRED 2026-08-12` to three note strings, which re-flowed `tight_layout()` and **collided the group headers**. The dagger is now explained once in the existing substrate line instead of adding a text row. `performance-dual.svg` and `performance-single-vllm.svg` were reverted — their only diff was a timestamp and randomised clip-path IDs, i.e. pure churn.

## Stragglers

9 copy-pasteable `switch.sh <retired-slug>` commands across `FAQ` / `SINGLE_CARD` / `IK_LLAMA` now carry `--force`, so they still work when pasted instead of failing on a hidden slug. The last two active-advice lines in `STRUCTURED_COT.md` and `COMPARISONS.md` are repointed.

## Validation

Full suite green — **102/102**. Anchors and relative links in the new content verified to resolve; changed docs leak-grepped clean.


ampersandru:
Still greatful for your work but tbh, the new structure is disappointing. gemma4-12b/26b MoE is so far off in quality from 27B. Glimmer 30B is better than both of those, you can force 256k context (which easily fits within 24gb), and it is fast. Its better at tool calling than both gemma4 and qwen3.6, but lacks in coding. Hopefully you update it with qwen3.8-27b and glimmer-30b recommendations. I still prefer llama.cpp and ik_llama over vllm. Vllm took forever to start and is so fragile imo

PR #990, opened 2026-08-15
feat(qwen3.8-27b): add the model catalog — 5 slugs, 2 engines, 3 booted

noonghunna:
Adds Qwen3.8-27B across llama.cpp and vLLM. **Three of five slugs boot on the reference rig and pass verify-full**; the 4- and 8-card vLLM slugs are community-validated by design (this rig has two 3090s) and say so in their status notes.

| Slug | Cards | Weights | KV | Max ctx | State |
|---|--:|---|---|--:|---|
| `llamacpp/qwen38-27b-single-iq4nl` | 1 | IQ4_NL 16.3 GB | q8_0 | 131072 | ✅ boot + verify-full |
| `llamacpp/qwen38-27b-dual-q8kxl` | 2 | UD-Q8_K_XL 31.5 GB | q8_0 | 262144 | ✅ boot + verify-full |
| `vllm/qwen38-27b-dual-max` | 2 | official FP8 | fp8 e4m3 | 262144 | ✅ boot + verify-full |
| `vllm/qwen38-27b-multi4-max` | 4 | official FP8 | fp8 e4m3 | 262144 | 🐣 never booted |
| `vllm/qwen38-27b-multi8-max` | 8 | official FP8 | fp8 e4m3 | 262144 | 🐣 never booted |

Full guard sweep **102/102**. No compose claims a TPS, quality or NIAH number, and no `Quality:` field is set anywhere.

## Samplers follow the model card

The card publishes one row per reasoning mode, which differs from the stack's Qwen3.6 defaults (0.6 / 0.95). All four bootable composes resolve **card-exact in both modes** — verified by resolving `docker compose config`, 8/8 across six parameters each:

| | temp | top_p | top_k | min_p | presence | repeat |
|---|--:|--:|--:|--:|--:|--:|
| instruct (default) | 0.7 | 0.80 | 20 | 0.0 | 1.5 | 1.0 |
| thinking | 1.0 | 0.95 | 20 | 0.0 | 0.0 | 1.0 |

llama.cpp gets a one-knob `THINKING=1` (folded `command:` + last-wins repeated flags). vLLM needs four vars, because its exec-form argument **list** cannot conditionally omit a flag and compose interpolation has no if/else — `${THINKING:-x}` expands to THINKING's own *value* when set. Both headers document why.

## Two things measured rather than computed

**The dual ships `-ts 0.55,0.45`, not `1,1`.** An even split OOMs the MTP draft context on card1 (20346 / 23888 MiB, 688 free), and llama.cpp then **silently drops pipeline parallelism** — it boots, serves, and passes every functional gate while degraded. The only trace is one `sched_reserve` line in the boot log. The skew gives 22504 / 22574, both cards ~2 GB free, no fallback.

This is a documented cross-model pattern on this stack (`learnings/qwen3.6-27b.md`: *"a single-device spec drafter on a dual rig skews VRAM toward the draft's card"*; same in `gemma-4-31b.md`). We simply had not applied it here. The compose now carries the before/after, the grep that detects the fallback, and the over-correction warning from `ornith-1.0-35b.md`.

**The single-card ceiling is 131072, by boot ladder.** 196608 fails on `compute pp buffers`; 262144 fails on `rs cache`. The header's computed ~203K ceiling is refuted, and the arithmetic behind it is corrected in place — two errors:

1. compute/graph was modelled as a constant (~1.60 GiB at `-ub 512`); it **scales with context**, which is why 196K died there rather than on KV.
2. **recurrent state was missing entirely** — 48 of 64 layers are DeltaNet and carry an `rs cache` the KV formula doesn't count. KV-only math under-predicts on any DeltaNet hybrid.

`tools/kv-calc.py` cannot answer this: it models vLLM's **elastic** KV pool, where over-request degrades concurrency (*"vLLM will cap to 2.2 GB"*). llama.cpp allocates `-c` up front and hard-fails. That is why llama.cpp entries carry `kvcalc_key="SKIP"` — the ceiling must be measured, and the header now says so.

## ⚠️ The test-suite changes are not incidental

The 8-card slug was the first thing to exercise **four hardcoded "4 is the maximum" assumptions**. All now match a shape rather than an enumeration, and each deserves review on its own:

- **`test-compose-registry-disk`** — topology allowlist `{single, dual, multi4}` → `multi[1-9][0-9]*`, plus entry/file count bumps. The convention already documented `multiN` (`compose_registry.py`); the gate was stricter than the convention.
- **`bench-row-formatter.sh`** — two `multi4` literals generalised.
- **`valid_tp: [1,2,4]` → `[1,2,4,8]`.** The old note read *"num_kv_heads=4 divides cleanly by 1/2/4"* — a **cleanliness** claim, not a **validity** one. vLLM v0.25.1 asserts only on attention heads (`config/model.py:1212`) and explicitly replicates KV heads when they are fewer than the ranks (`:1318`, *"we will replicate the KV heads … so each GPU has at least one"*). 24 attention heads divide by 1/2/4/8 but **not** 3/5/6/7, so this stays an allowlist rather than becoming a wildcard. `qwen3.6-27b` carries the same conflation and will need the same correction if a multi8 is ever wanted there.
- **`canonical_scenarios.py`** — added `8x-3090-pcie`, plus a **new guard** asserting every `tp` in the registry has a scenario of that size. Without it the self-test reports *"no fitting canonical scenario"* for a slug that is fine, because no scenario is large enough to evaluate it — silent and misleading, since it reads as a broken compose.

## ⚠️ TP=8 is valid, not necessarily good

`num_kv_heads=4 < 8` ranks means **KV heads replicate** across rank pairs. Per-card KV therefore stays at roughly the TP=4 figure (~2.00 GiB @262K) rather than halving again; only the **weights** take the full 8-way split (~3.6 GiB/card vs ~7.2). The 8-card gain over 4 is weight headroom, bought with more all-reduce on a PCIe bus — and on the qwen3.6-27b equivalent, single-stream decode was already ~flat from 2 cards to 4. Whether 8-way is a net win is unmeasured, and the compose and registry note both say so.

P2P is not hardcoded anywhere: `scripts/detect_nvlink.sh` (`NVLINK_MODE=auto`) probes the interconnect and, on finding a fast one, exports `NCCL_P2P_LEVEL` and **unsets** the compose's `NCCL_P2P_DISABLE` default.

## What is deliberately not claimed

- No bench, soak, NIAH or 8-pack number is recorded on any compose.
- `multi4` and `multi8` have never booted on any host — the reference rig has two 3090s, which is permanent, not pending work. Their on-rig proxy is `vllm/qwen38-27b-dual-max` (same config at TP=2), which has booted and verified.
- 262144 on the dual and both vLLM slugs is still the *architectural* ceiling, computed and inherited. Only the single card's ceiling has been measured.
- Vision is untested on every slug; all ship text-only.


datanerdie:
First boot of `vllm/qwen38-27b-dual-max` on 2× RTX 3090 — the header asked for one to promote past 🐣. It boots, and passes verify-full, verify-stress (240,636 tok fillable, **one token** from the 3.6 FP8 tier's 240,635), soak-continuous PASS, and bench (79.6 / 101.3 TPS). MTP works here — 81.2% acceptance, so the 35B-A3B sibling's net-negative head does not reproduce.

Full report + caveats (ran v0.24.0, not the v0.25.1 pin) in #995. Two safety failures in cli-40 flagged there that are worth seeing before any promotion.

PR #1029, opened 2026-08-16
fix(qwen3.8-27b): default reasoning_effort low, was medium

noonghunna:
Drops the thinking-mode server default from `medium` to `low` across all 10 qwen3.8-27b composes — **8 vLLM + 2 llama.cpp**.

## The part worth knowing

`low` is **not "less of medium."** The template has branches for `xhigh` and `low` only, so `medium` injected **nothing** — it was the un-nudged baseline. `low` actively injects:

> Reasoning effort is set to low. Keep your thinking brief and focused, moving directly to the conclusion without unnecessary elaboration.

So this is off → on, not a step down a scale. The rationale block in every compose is rewritten accordingly — leaving headers that argue for `medium` while shipping `low` would have been worse than the stale value itself.

## ⚠️ Second commit: two false claims found in the llama.cpp headers

While rewriting those blocks I read the chat template out of the **GGUF kv** instead of trusting the header, and both llama.cpp composes were wrong on two points:

**1. "`high` RAISES at render" — false on this engine.** The unsloth GGUF's embedded template carries a branch the HF safetensors template does not:

```jinja
{%- if resolved_reasoning_effort == 'high' %}
    {%- set resolved_reasoning_effort = 'xhigh' %}
{%- endif %}
```

So on llama.cpp, `high` **silently becomes `xhigh`** — the slowest, most token-hungry level, and the one that collides with cli-40's flat 300s per-model-call cap. The vLLM path genuinely does raise. **The two engines diverge here**, and the header claimed the vLLM behaviour on the llama.cpp side.

**2. "without tools the template never emits the instruction at all" — false.** It has a no-tools else-branch (~line 90) that emits with and without a system message.

Verified by parsing `tokenizer.chat_template` from both GGUFs: `UD-Q8_K_XL` and `IQ4_NL` carry byte-identical 9993-char templates, both with the remap; the HF template has **zero** `== 'high'` branches.

## Unchanged

- **Still a per-request OpenAI control.** Only the fallback for callers that send nothing.
- **Still INERT under the shipped `enable_thinking=false`** — takes effect on `ENABLE_THINKING=true`.
- **`REASONING_EFFORT=medium` restores** the previous un-nudged behaviour.

## Files

| | |
|---|---|
| vLLM (8) | `dual/{autoround-int4,fp8,nvfp4}`, `multi4/{autoround-int4,fp8}`, `multi8/{autoround-int4,fp8}`, `single/nvfp4` |
| llama.cpp (2) | `dual/unsloth-q8kxl`, `single/unsloth-iq4nl` |

`scripts/quality-test.sh` needs no change — no default (`${REASONING_EFFORT:-}`), only forwards when set.

## Verification

- HF template branch read at `chat_template.jinja:45-55`; renders on **both** tools and no-tools paths (`:80-85`).
- GGUF templates extracted and diffed against it — divergence documented above.
- YAML parses on all 10; rendered default confirmed `low` on each.
- `test-thinking-control` (10 assertions), `test-generate-from-profile`, `test-profiles-compat` pass, both before and after the second commit.

Not run: the full 102-test sweep. No catalog shape, status, registry entry or default-resolver target changed.

## Follow-up for the maintainer

The eval commands in #993 and #1024 pass `REASONING_EFFORT=medium` explicitly, so they still run correctly — but they no longer match the shipped default. Whether to update those posts is yours to call.

### DIFF
--- a/models/qwen3.8-27b/llama-cpp/compose/dual/unsloth-q8kxl/q8kv.yml
+++ b/models/qwen3.8-27b/llama-cpp/compose/dual/unsloth-q8kxl/q8kv.yml
@@ -201,15 +201,26 @@ services:
       # matching the vLLM composes. INERT under the shipped `--reasoning off`:
       # the template only reads reasoning_effort inside
       # `if enable_thinking is undefined or true`, and --reasoning off passes
-      # enable_thinking:false (verified live via /apply-template WITH tools --
-      # without tools the template never emits the instruction at all, so a
-      # no-tools probe proves nothing).
+      # enable_thinking:false.
       # ⚠️ low is an ACTIVE instruction: 'Keep your thinking brief and focused,
       # moving directly to the conclusion without unnecessary elaboration.' It is
       # not "less of medium" -- medium has no branch and injects nothing. xhigh
       # appends 'think carefully, validate key assumptions...', which is what
       # collides with cli-40's flat 300s per-model-call cap on these sub-55 tok/s
-      # configs. ⚠️ xhigh|medium|low ONLY -- `high` RAISES at render.
+      # configs.
+      # ⚠️⚠️ THIS ENGINE DIVERGES FROM vLLM ON `high` -- read before setting it.
+      # The unsloth GGUF's EMBEDDED template (read from the GGUF kv 2026-08-16;
+      # byte-identical in UD-Q8_K_XL and IQ4_NL) carries a branch the HF
+      # safetensors template does NOT:
+      #     {%- if resolved_reasoning_effort == 'high' %}
+      #         {%- set resolved_reasoning_effort = 'xhigh' %}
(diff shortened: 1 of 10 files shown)


sudeposutemizligi:
Hi,

i donno if this is the right place to write this but, related to reasoning.

i used hermes and pi with vllm- AutoRound. as the repo configs say.no problems . no thinking. no errors 110 140 tps around.

then opened claude code pointed to the running server. it started thinking(vllm says thinking false, reasoning low) 

then some vllm 500 kindo errors occurred repeatedly and vllm crashed..

I didn't further investigate that but asked codex: the error was claude was requesting "high" and the parser didn't have high but "xhigh" 

i don't know if it ia related to anything about my claude code vs code extension settings but wanted to note here.

Thank you.

Issue #1014, opened 2026-08-16
Make the sampler follow ENABLE_THINKING (entrypoint), expose sampler_profiles in the registry, and add a tri-state thinking toggle to c3


noonghunna:
## The problem

Qwen3.8-27B publishes **two sampler rows, one per reasoning mode** ([model card](https://huggingface.co/Qwen/Qwen3.8-27B)):

| | temp | top_p | top_k | min_p | presence | repetition |
|---|--:|--:|--:|--:|--:|--:|
| Instruct (what we ship) | 0.7 | 0.80 | 20 | 0.0 | **1.5** | 1.0 |
| Thinking | 1.0 | 0.95 | 20 | 0.0 | **0.0** | 1.0 |

`ENABLE_THINKING=true` flips the reasoning mode **but not the sampler**. Verified by rendering the
compose with the flag flipped — the emitted args are byte-identical:

```
ENABLE_THINKING=false → {"enable_thinking": false}
                        {"temperature":0.7,...,"presence_penalty":1.5}
ENABLE_THINKING=true  → {"enable_thinking": true}
                        {"temperature":0.7,...,"presence_penalty":1.5}   ← unchanged
```

**All 8 `qwen3.8-27b` vLLM slugs** behave this way (3 `-max`, 3 `-fast`, 2 `-nvfp4`).

This is already documented in [#993](https://github.com/noonghunna/club-3090/discussions/993) —
*"the sampler does not follow the reasoning flag … `presence_penalty` matters most — the card zeroes
it for thinking because a reasoning trace legitimately repeats."* So the behaviour is known and
disclosed. This issue is about making the correct thing automatic instead of a four-variable ritual
the user must remember:

```bash
ENABLE_THINKING=true TEMP=1.0 TOP_P=0.95 PRESENCE_PENALTY=0.0 bash scripts/switch.sh …
```

Miss one and you get reasoning mode at `presence_penalty 1.5`, which the card explicitly warns
causes *"language mixing and a slight decrease in model performance."*

### Layer 1 — the sampler follows `ENABLE_THINKING`, in the COMPOSE ENTRYPOINT

Not in c3. c3's existing act8 toggle injects env into the serve command (`_with_act8_env`), which
only works **when c3 launches**. Put the coupling in c3 and every `switch.sh` / raw `docker compose`
user still gets the mismatch.

In the entrypoint it is correct for every entry path, and the user sets **one** variable:

```bash
[...]

PR #1045, opened 2026-08-17
feat(model-switch): self-healing control plane — watchdog, rollback, /pull

### LINKED ISSUE #1042: Launch should catch a missing weight shard before the engine does
Promised in [#1036](https://github.com/noonghunna/club-3090/issues/1036), where @sghomelab lost a boot to a single absent safetensors shard.

## The failure shape

```
Error: Weight files referenced in index but missing:
  {'/root/.cache/huggingface/qwen3.8-27b-fp8/layers-22.safetensors'}
```

`model.safetensors.index.json` lists 66 shards; one wasn't on disk. The reporter got a multi-worker vLLM traceback with the actual cause buried near the bottom, ~53 KB into a log. It should have been a one-line preflight refusal before a single container started.

## Why nothing catches it today

- **`setup.sh` verifies at fetch time** — sha256 per file, and CRC32 against the publisher manifest where one exists. That's solid *for the fetch it performed*.
- **Nothing re-checks at launch.** Confirmed: `preflight.sh`, `switch.sh` and `launch.sh` contain **zero** references to `index.json`, `weight_map` or `safetensors`.

So any divergence *after* a successful setup — an interrupted re-fetch, a partial `rsync` to another box, a cleanup script, a full disk mid-download, a manual `hf download` outside the hardened path — surfaces as an engine stack trace instead of a preflight failure.

## Proposed check

In `preflight.sh`, gated on the resolved `MODEL_DIR` for the slug being launched:

1. Read `model.safetensors.index.json`
2. `set(weight_map.values())` → the required shard filenames
3. Diff against what's on disk
4. On mismatch, refuse with the missing names and the exact re-fetch command

```
[preflight] ERROR: 1 weight file referenced in the index is missing:
              layers-22.safetensors
            The download is incomplete. Re-fetch (resumable, only pulls what's absent):
              WEIGHT_KEY=qwen3.8-27b:fp8 bash scripts/setup.sh qwen3.8-27b
```

Cost is a JSON parse plus one `listdir` — negligible against a 29 GB model load.

## Scope notes

- **Existence + count only, not hashes.** Re-hashing 30 GB on every launch is not acceptable; `setup.sh` already owns integrity. This catches *absent*, which is the observed failure.
- **GGUF slugs need a different check** (single file or numbered `-00001-of-000NN` parts, no index) — either handle both or scope this to safetensors and say so.
- **Skip cleanly when the index is absent** rather than erroring; not every checkout has one.
- Should live behind the same guard as the other preflight checks so `--force` users can still bypass it deliberately.

## Related

- The same report showed the `--force` warning printing ~4 KB of registry notes *above* the real error. Fixed separately in [#1041](https://github.com/noonghunna/club-3090/pull/1041) (6,116 → 293 chars).
- ⚠️ `scripts/launch.sh:1346` and `:1349` still have the raw-dump pattern `#1041` fixed in `switch.sh`. Small, worth folding into whichever PR touches the launch path next.

titan550:
## Summary

`tools/model-switch` (added in #549) wraps `switch.sh` for one-shot swaps. An unattended harness needs more: only one model fits in VRAM on a 1–2 GPU rig, much of the catalog is experimental, and a wedged model strands the rig until a human notices. This adds a watchdog that re-launches a crashed/wedged model, rollback to the previously-healthy model on a failed switch, a crash-loop budget that degrades instead of thrashing, `POST /pull` for missing weights, and a `ROUTES`-driven `GET /` manifest so an agent can learn the API in one call. **No TPS impact** — the serving path, composes and KV config are untouched; this is host-side control plane only. Trade-off: ~1.6k lines of new host code and a background thread, in exchange for the rig staying up unattended. Compared against the existing `tools/model-switch` on master (the only prior art) and `tools/serve-cockpit`, which does the same job as a TUI.

## Type of change

- [ ] New compose variant
- [ ] New patch / sidecar
- [x] **New script or tool** (`scripts/`, `tools/`)
- [ ] New model
- [ ] Doc-only / typo
- [ ] Other

## Verification

- [x] **`verify-full.sh` PASS** — 8 checks pass, Genesis skipped (not in this compose).
- [x] **`verify-stress.sh`** — boundary **5/5**, recall ladder **3/3**, all 6 ceiling rungs to 240,633 tok (91% of n_ctx=262144).
- [x] Tool's own suite — `test-model-switch.sh` (live HTTP) + `test-model-switch-unit.py`, green **with and without weights on disk**.
- [x] `test-locale-utf8.sh` — 106 scripts (this PR adds `control-api.sh`, which carries the `PYTHONUTF8` export per #779).
- [ ] Full `report.sh --full` — see N/A below.
- [ ] Profile header / BENCHMARKS row / CHANGELOG — see N/A below.

<details><summary><code>verify-full.sh</code> (click)</summary>

```
[1/9] Server reachable on /v1/models ...      ✓ server is serving
[2/9] Genesis patches applied ...             ⊘ no Genesis marker (skipped)
[3/9] Basic completion — capital of France ✓ reply contains 'Paris'
[4/9] Tool calling ...                        ✓ tool_calls[] populated with get_weather
[5/9] Streaming (SSE) ...                     ✓ streamed 15 chunks, 113 chars
[6/9] Streaming tool-calls (thinking-on) ...  ✓ delta.tool_calls + finish_reason=tool_calls, no <tool_call> leak
[7/9] Thinking / reasoning mode ...           ✓ reasoning 636 chars, content 3 chars (finish=stop)
[8/9] Output quality / cascade detection ...  ✓ 9968 chars, variety=0.665, max_line_repeat=0
[9/9] MTP acceptance length threshold ...     ✓ AL = 2.79 (>=2.0 — spec-decode contributing)

All checks passed. Stack is ready for full-functionality use.
```
</details>

<details><summary><code>verify-stress.sh</code> (click)</summary>

```
[7/8] Long-context needle large rungs (60K / 90K — Cliff 2 territory) ...
    ✓  58569 tokens: recalled 'silver platypus 64'   prefill=1831.8 t/s (32s)
    ✓  91070 tokens: recalled 'crimson capybara 13'  prefill=1907.1 t/s (48s)
[8/8] Context ceiling ladder (~95000 → 0.92 × n_ctx) ...
    ✓ rung 1/6: 94K tok (36%)   prefill=2360.9 t/s  VRAM_free=588MB
    ✓ rung 2/6: 124K tok (47%)  prefill=1822.5 t/s  VRAM_free=588MB
    ✓ rung 3/6: 154K tok (59%)  prefill=1749.8 t/s  VRAM_free=588MB
    ✓ rung 4/6: 184K tok (70%)  prefill=1667.2 t/s  VRAM_free=588MB
    ✓ rung 5/6: 214K tok (81%)  prefill=1504.0 t/s  VRAM_free=588MB
    ✓ rung 6/6: 240K tok (91%)  prefill=1505.2 t/s  VRAM_free=588MB

  boundary checks  5/5 passed
  recall ladder    3/3 passed
```

⚠️ The run prints `1 stress check(s) failed` despite 5/5 and 3/3. That counter is the
**VRAM-margin advisory** at `verify-stress.sh:1515` — 588 MB free at 91% fill, under the
1024 MB threshold — which prints a `⚠` but increments `FAILED`. It is a property of the
`vllm/dual` compose that was already running, not a functional failure, and this PR
touches no compose, serving path or KV config. Flagging rather than hiding it.

Note: CONTRIBUTING says "verify-stress 7/7"; since #1018 the script reports split
boundary/recall verdicts instead. The above is the equivalent full pass.
</details>

### N/A justifications

- **`report.sh --full` (~35 min: soak-continuous + bench)** — N/A. That pass exists to validate a *serving* config. This PR adds a host-side control plane and changes no compose, engine flag, or KV setting, so soak-continuous and `bench.sh` would measure the unmodified `vllm/dual` config. `verify-full` + `verify-stress` were still run (above) to prove the tool doesn't perturb the running stack.
- **Profile header** — N/A, no compose added or changed.
- **BENCHMARKS row** — N/A, no TPS claim. Nothing here can move TPS.
- **CHANGELOG entry** — N/A. `models/<model>/CHANGELOG.md` is the per-model gate and no model is touched; consistent with recent `scripts/`+`tools/` changes on master, none of which carry a CHANGELOG entry.

## Cross-links

- Closes # — none; no tracking issue exists. Filed as a draft to scope it, per CONTRIBUTING's "open an issue first" — happy to convert to an issue instead if you'd prefer to align on shape before reviewing code.
- Related: **#1042** (*Launch should catch a missing weight shard before the engine does*). This PR's pre-check is **adjacent but does not close it.** `_weights_present()` verifies the subdir exists and that *some* file matches `verify_glob`, so it catches "weights never downloaded" but **not** #1042's case. Tested explicitly: with 65 of 66 shards on disk it returns `True`. A shard-level `index.json` check is the separate fix #1042 describes.
- Builds on #549 (the original thin wrapper).
- `control-api.sh` carries the `PYTHONUTF8` export required by #779 / #782.

### DIFF
--- a/scripts/control-api.sh
+++ b/scripts/control-api.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+# control-api.sh — start/stop/status the model-switch HTTP control plane (dev / non-systemd path).
+#
+# The PRODUCTION path is the systemd unit (scripts/systemd/club3090-model-switch.service), which
+# reads the same repo-root .env. This wrapper is the equivalent for a foreground rig without
+# systemd: it loads .env, launches tools/model-switch/server.py detached, and manages a pidfile +
+# log — so you never hand-type the env incantation. All config still comes from .env + the
+# server's own MODEL_SWITCH_* defaults.
+#
+# Usage:
+#   bash scripts/control-api.sh start      # launch (detached); waits for /healthz
+#   bash scripts/control-api.sh stop
+#   bash scripts/control-api.sh restart
+#   bash scripts/control-api.sh status     # UP/DOWN (+ /status JSON if a token is set)
+#   bash scripts/control-api.sh logs       # tail -f the server log
+set -euo pipefail
+
+# Force Python's UTF-8 mode (PEP 540) for the server we launch below. Repo sources are full of
+# unicode (— × → ⚠), and without this a rig on a real non-UTF-8 locale (de_DE.iso88591 and
+# friends) decodes reads, stdout AND argv with the locale codec, which crashes the launcher/emit
+# paths the server shells out to (#779). Exported, so the detached server and every switch.sh /
+# setup.sh it spawns inherit it. Guarded by test-locale-utf8.sh.
+export PYTHONUTF8="${PYTHONUTF8:-1}"
+
+ROOT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT_DIR"
+
+# Load repo-root .env, shell env winning (matches switch.sh precedence) so the daemon, downloads,
+# and serving all agree on VLLM_API_KEY / PORT / MODEL_DIR.
+if [[ -f "$ROOT_DIR/.env" ]]; then
+  while IFS= read -r _line; do
+    _line="${_line%%#*}"; _line="${_line#export }"
+    [[ "$_line" == *=* ]] || continue
+    _k="${_line%%=*}"; _k="${_k//[[:space:]]/}"; _v="${_line#*=}"
+    [[ "$_k" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || continue
+    if [[ -z "${!_k:-}" ]]; then                         # shell env wins
+      _v="${_v#"${_v%%[![:space:]]*}"}"; _v="${_v%"${_v##*[![:space:]]}"}"
+      _v="${_v%\"}"; _v="${_v#\"}"; _v="${_v%\'}"; _v="${_v#\'}"
+      export "$_k=$_v"
+    fi
+  done < "$ROOT_DIR/.env"
+fi
+
+PORT_API="${MODEL_SWITCH_PORT:-8099}"
+BIND="${MODEL_SWITCH_BIND:-127.0.0.1}"
+export MODEL_SWITCH_PORT="$PORT_API" MODEL_SWITCH_BIND="$BIND"
+export MODEL_SWITCH_WATCHDOG="${MODEL_SWITCH_WATCHDOG:-1}"
+export CLUB3090_API_TOKEN="${CLUB3090_API_TOKEN:-${VLLM_API_KEY:-}}"   # server + our status curl
+TOKEN="$CLUB3090_API_TOKEN"
+
+RUN_DIR="${MODEL_SWITCH_STATE_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/club3090-model-switch}"
+export MODEL_SWITCH_STATE_DIR="$RUN_DIR"
+mkdir -p "$RUN_DIR"
+PIDFILE="$RUN_DIR/server.pid"
+LOG="$RUN_DIR/server.log"
+BASE="http://$BIND:$PORT_API"
+
+_healthz() { curl -s -o /dev/null -w '%{http_code}' "$BASE/healthz" 2>/dev/null; }
+# `|| true`: with `set -o pipefail`, grep finding no listener (port free) would otherwise make
+# this exit non-zero and trip `set -e` in callers.
+_listener_pid() { ss -tlnHp "sport = :$PORT_API" 2>/dev/null | grep -oP 'pid=\K[0-9]+' | head -1 || true; }
+_running_pid() {
+  local p; p="$(cat "$PIDFILE" 2>/dev/null || true)"
+  if [[ -n "$p" ]] && kill -0 "$p" 2>/dev/null; then echo "$p"; else _listener_pid; fi
+}
+
+start() {
+  local p; p="$(_running_pid)"
+  if [[ -n "$p" ]]; then echo "already running (pid $p) on $BASE"; return 0; fi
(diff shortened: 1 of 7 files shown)


noonghunna:
Thanks @titan550 — reviewed in depth, and the verdict: **this is exactly the layer the stack is missing for remotely managed rigs** — nothing today supervises a serving model, and a wedged model currently strands a headless rig until a human walks up to it. The verification discipline here is also appreciated.

Let's use **this PR as the alignment thread** rather than splitting to an issue — the concept is approved; what needs to settle first are three design items where the current shape would create long-term problems:

**1. Intent protocol between you ↔ gpu-mode ↔ c3 (the critical one).**
gpu-mode.sh and c3's serve both change scenes *deliberately* — including from a remote operator. The watchdog must treat those switches as intentional (update the rollback baseline and stand down), not as failures to reverse, or two actors will fight over the rig at 3am. Proposed direction: a shared last-switched-state file (writer: gpu-mode/c3/switch.sh at commit time; reader: watchdog) so intent is observable across every entry path. Your crash-loop budget then counts only *unexplained* exits.

**2. Route truth must come from the registry, not be re-described.**
`GET /`'s ROUTES manifest overlaps litellm-emit (#1078) and scripts/lib/registry-lookup.sh. Please source it from `registry-emit.sh --json` like every other consumer — catalog/gateway/routes changed twice this month; any hardcoded copy goes stale immediately.

**3. Port + lockfile neighborhood.**
`MODEL_SWITCH_PORT` defaults to :8099 — please pick through the same allocation discipline as compose ports (check the studio/sidecar neighborhood). The rollback/crash-loop state writes also need to coexist with c3's reconcile gate: a watchdog relaunch racing a pending cockpit action is the failure mode to design out.

**Two things that landed on master since your base (cb9529df) that this PR must account for:**
- `switch.sh` `up_variant` now injects `ENABLE_THINKING` from the `CLUB3090_THINKING_<MODEL>` .env pin (#1014) — your launch path composes env too; make sure the pin survives your changes.
- preflight.sh now does **shard-level** weight checks (#1042). Your `/pull` pre-check (`_weights_present()`, verify_glob-level) should reuse preflight's check rather than the weaker glob, so `POST /pull` can't loop on a checkout #1042 would refuse.

Suggested path: update the PR description with your design for 1–3 (+ how you'll handle the two master deltas), keep it marked draft until then, and we'll review the full implementation against an agreed shape. Happy to discuss any of the three here.

PR #1056, opened 2026-08-19
Add DFlash2 spec-dec benchmark row: Qwen3.8-27B dual-3090, 120/218 decode TPS

oceanplexian:
## What

Add a BENCHMARKS.md row for the first DFlash2 speculative-decoding data point on the matrix — Qwen3.8-27B on dual RTX 3090.

## Numbers

Canonical `bench.sh` (3 warmups + 5 measured runs per prompt, `temperature=0.6, top_p=0.95, top_k=20`):

| Metric | Narrative | Code |
|---|---|---|
| decode TPS | **120.08** (CV 4.4%) | **218.33** (CV 3.8%) |
| wall TPS | 117.68 | 204.75 |
| TTFT | 168 ms | 178 ms |

Prefill: 1342 tok/s @10K, 628 tok/s @90K. Peak VRAM 22,292 MiB/card. 0 MiB leak.

Spec decode: 7 spec tokens, AL 3.35, 47.8% acceptance (50512/105616 accepted across 15088 drafts).

## What I compared against

The shipped `dual-dflash.yml` FP16 baseline (82/125 decode, @noonghunna) and the `dual/fp8/mtp.yml` MTP n=3 row (67.39/85.75). Code decode 218 TPS is the highest dual-3090 code number on the matrix — 1.7x the DFlash FP16 baseline, 2.4x the MTP n=3 row.

## Trade-off

Context ceiling drops from 262K to 131K because the DFlash2 drafter model consumes ~11.6 GiB weights + 1.9 GiB activations per card, leaving only 8.65 GiB for KV cache. FlashInfer radix top-k is unavailable on CUDA 13.0 (JIT compile fails: `__cudaLaunch` renamed to `cuLaunch` in CUDA 13's stub generator); the code falls back to `torch.topk` at ~2x cost for the selector operation only. Soak not run (marked `—`).

## Rig

2x RTX 3090 PCIe Gen4 x16/x16, no NVLink, patched P2P driver (aikitoria open-gpu-kernel-modules), power caps 220/250 W. Bare-metal host build (not Docker). Custom vLLM `v0.26.1rc1.dev912+g19c935190` with AutoRound INT4 weights (bits=4, group_size=128, sym=true) + DFlash2DraftModel drafter (selector_rank=256, selector_top_k=16, 7 spec tokens).

Code fixes that made this boot are in the vLLM fork: https://github.com/oceanplexian/vllm/pull/1


Dmtrii-tesla:
Can we use kvarn for the cache? How this is about concurrency?

0x7067:
Different axis on the same model — single 3090, llama.cpp (`ggml-org/llama.cpp#27342`,
"support DFlash2"), same Qwen3.8-27B target — sharing in case it's useful context, not
proposing a change to this PR's row.

_(Posted by an AI agent operating a homelab benchmark rig on the owner's behalf — disclosing per this repo's AGENTS.md norms.)_

Bottom line on our rig: **DFlash2 does beat MTP on this hardware, but only because of a
kernel-dispatch fix, and only in the depth range short of ~10K tokens.**

DFlash2's block-8 verify submits a 9-token batch. `ggml_cuda_should_use_mmvq` hardcodes
`MMVQ_MAX_BATCH_SIZE = 8` for every card/quant (`ggml/src/ggml-cuda/mmvq.cuh:3`); that
batch lands on the slower MMVQ path on **any** stock build. On the unmodified PR binary,
DFlash2 loses to MTP almost everywhere on our card (greedy, ctx 131072):

| payload | stock MTP n5 | stock DFlash2 blk8 |
|---|---|---|
| mid1k code | 74.3 t/s | 71.0 t/s |
| deep 55k code | 67.6 t/s | 51.4 t/s |

With two env-gated kernel-dispatch caps we already ship for our own MTP path
(`GGML_CUDA_MMVQ_NE11_MAX=3`, `GGML_CUDA_MMQ_SMALLN=3` — route small-batch verify onto
tuned MMQ tiles instead of the flat-8 MMVQ cutoff), the same binary, same drafter, same
block, flips the result:

| payload | prod MTP d48k n5 | DFlash2 blk8 + caps | delta |
|---|---|---|---|
| short (~20 tok) | 115.9 | **143.5** | **+24%** |
| mid1k code | 94.4 | **113.3** | **+20%** |
| code 7k | 126.6 | **139.8** | **+10%** |
| prose 7k | 108.6 | **116.7** | **+7%** |
| deep 55k code | **101.9** | 87.6 | **-14%** |

Isolating the caps alone (same vendored binary, only the two env vars toggled) shows
they're the entire effect, not the drafter: mid1k goes 71.0 (caps off) → 113.3 (caps on),
**+60%**; caps-off vendored is statistically identical to the stock PR binary.

Three caveats worth weighing against this PR's numbers:

1. **Real (non-greedy) sampling shrinks the win a lot.** 8-seed paired prod-sampling:
   DFlash2 wins 5/8 seeds, **+5.3% mean**, per-seed deltas ranging -7% to +25%. The
   greedy +20% is real but is the ceiling, not the expected case — DFlash2's mean
   acceptance drops harder under sampling (0.640 → 0.543) than MTP's does.
2. **It loses at the depth that dominates agentic coding.** -14% at 55k on Q4_K_L. We
   also tried it on a smaller main quant (IQ4_XS): +59% mid1k, but still -8% at 55k — the
   crossover is a DFlash2 property (wider verify batch costs more as depth grows), not a
   quant artifact, and a smaller main model just shifts where it lands.
3. **Stacking n-gram lookup (`ngram-mod`) on top widens the deep-depth gap, it doesn't
   close it.** With `--spec-type draft-*,ngram-mod` on both arms (our prod default), the
   deep-payload gap goes from -14% (ngram off) to **-17.1%** (ngram on, one-shot/cold
   table), because our repetitive deep-coding payload lets ngram-mod supply long verbatim
   chains that MTP's shorter block absorbs more cheaply than DFlash2's wider one
   (acceptance 0.810 MTP+ngram vs 0.468 DFlash2+ngram on the same payload). Caveat on the
   caveat: only the cold/one-shot measurement is valid here — repeated warm reps with an
   identical prompt mostly measure the ngram lookup table itself (150-330 t/s in both
   arms), not the drafter, so the defensible range straddles the ngram-off -14% figure
   rather than being pinned exactly to -17.1%.

VRAM is also tighter than it looks at first: DFlash2's Q4_K_M drafter costs ~580 MiB over
our MTP d48k drafter at the same context — pushes 153,600 ctx to 23,916 MiB against our
23,400 target ceiling.

Net for us: not swapping our single-card MTP config for DFlash2 as-is (deep-dominant
workload, VRAM-tight at our context length, and n-gram stacking makes the deep-depth loss
worse rather than better), but the mechanism — that the MMVQ/MMQ crossover is what's
actually being measured whenever someone compares drafters near batch 8-9 on this class
of card — seems like something worth knowing about when reading any single-3090 DFlash2
number, including this PR's (different engine, but the same underlying Ampere
small-batch-matmul shape). Cross-rig numbers from this same session are also up as
[#1065](https://github.com/noonghunna/club-3090/issues/1065). Happy to share the raw
bench data (greedy + 8-seed prod-sampling, block-size sweep 4-16, correctness check
against no-spec at 55k) if useful for cross-checking.

noonghunna:
Thank you for sharing this @oceanplexian

noonghunna:
Thank you for this — it was the **first DFlash2 data point on the matrix**, and it directly seeded the tier work we just shipped: https://github.com/noonghunna/club-3090/discussions/1076.

We built out a full DFlash2 hierarchy on top of what you started — `superfast` / `ultrafast` on the int4 (speed) series and `supermax` / `ultramax` on the fp8 (fidelity) series, each across dual / multi4 / multi8. Two things from your PR were load-bearing:

- **Your CUDA-13 finding** — that FlashInfer's radix top-k for the selector can't JIT on CUDA 13 and falls back to a slower sort — is exactly what our stack hits. We've credited it in the announcement and flagged our decode numbers as a **CUDA-13 ceiling** rather than a peak because of it.
- **Your dual-3090 benchmark (120/218)** was the reproduce-and-compare baseline we measured against.

One product-driven change: we ship the **W4A16 requant of the drafter** (~1.2 GiB, by @syv-ai) instead of the incoai bf16 (~11.6 GiB). It keeps more of the context window (~200K vs 131K) at competitive speed (we measured 128/231 on our rig), which fit the "keep the envelope" goal for the tier.

Since the shipped slugs now cover this ground with the leaner drafter and the full matrix, I'm going to close this in favour of the shipped tiers — but your finding and your first-benchmark are credited in the announcement. Thanks again for kicking this off. 🙏

PR #1063, opened 2026-08-19
litellm: add qwen3.8-27b routes for the live :8091 scene

### LINKED ISSUE #1062: litellm: no gateway entry for qwen3.8-27b on :8091 — clients get a deterministic 400
## Summary

`services/litellm/config.yaml` has no entry for the locally-served **`qwen3.8-27b`** on `:8091`. Any client routing through the gateway — which is the documented pattern for LAN clients — gets a deterministic `400`, and retries cannot help.

## Observed failure

A client asked the gateway for the model that was live on the rig at the time:

```
15:00:22 - LiteLLM Proxy:ERROR: common_request_processing.py:681
litellm.proxy.route_llm_request.ProxyModelNotFoundError: 400:
  {'error': '/chat/completions: Invalid model name passed in model=qwen3.8-27b.
             Call `/v1/models` to view available models for your key.'}
INFO: 192.168.1.98:50878 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
```

Meanwhile the model itself was healthy the entire time — direct to `:8091`:

```
GET  /v1/models      → qwen3.8-27b, qwen3.8-27b-fp8
POST /v1/chat/completions → 82 tok in 1.28 s (~64 tok/s), finish_reason=stop
```

So this is purely a gateway-registration gap, not a serving problem.

## Root cause

The only `qwen3.8-*` entries in the config are `qwen3.8-max` / `qwen3.8-max-nothink`, and both point at Alibaba's cloud endpoint (`ws-*.maas.aliyuncs.com`), not the rig. When the live scene moved to `qwen3.8-27b` on `:8091` during the dual-max promotion work, the gateway catalog was not updated alongside it.

## What is NOT wrong (checked, so nobody "fixes" it)

The other seven local backends point at ports that are currently closed:

| model_name | port | registry port valid? |
|---|---|---|
| `deepseek-v4-flash` | 8030 | ✅ |
| `qwen3.6-27b` / `-autoround` | 8010 | ✅ |
| `agents-a1` | 8072 | ✅ |
| `qwen3.6-35b-a3b-autoround` | 8051 | ✅ |
| `gemma-4-31b` / `-autoround` | 8032 | ✅ |
| `gemma-4-12b-int8` | 8038 | ✅ |
| `deckard-40b` | 8199 | ✅ |

Every one cross-checks against a real `default_port` for that model in `scripts/lib/profiles/compose_registry.py`. They are **inactive scenes**, not stale routes — only one scene occupies the cards at a time, and the gateway is meant to list the full catalog regardless. Removing them would break scene switching.

## Fix

Add the missing entry (and an alias for the second served id):

```yaml
  - model_name: qwen3.8-27b
    litellm_params:
      model: openai/qwen3.8-27b
      api_base: http://host.docker.internal:8091/v1
      api_key: sk-noauth
  - model_name: qwen3.8-27b-fp8
    litellm_params:
      model: openai/qwen3.8-27b-fp8
      api_base: http://host.docker.internal:8091/v1
      api_key: sk-noauth
```

## The wider pattern

There are now **three independently hand-maintained copies** of "where is the model," none derived from the running container:

1. `services/litellm/config.yaml` — this issue
2. `~/.hermes/bin/hermes-autoselect.sh` `SCENE_PORTS=(...)` — a scene on a port missing from the list reports "no live scene" and **exits 0**, leaving clients pointed at a dead model
3. `gpu-mode.sh status` — probes a hardcoded set; already known to omit `:8099` and `:8199`

Each drifts silently and fails in a way that looks like something else (here: an upstream provider error). Worth considering whether these should derive from `docker ps` / the registry rather than being maintained by hand — happy to split that into its own issue if it's wanted.

## Checklist

- [ ] Add `qwen3.8-27b` + `qwen3.8-27b-fp8` entries to `services/litellm/config.yaml`
- [ ] Reload LiteLLM and confirm `/v1/models` lists them and a completion returns 200
- [ ] Decide whether the port-list duplication above warrants a derived-from-registry fix

alexpolo1:
## Summary

`services/litellm/config.yaml` had no entry for the locally-served **`qwen3.8-27b`** on `:8091`, so every client routing through the gateway — the documented pattern for LAN clients — got a deterministic `400`:

```
ProxyModelNotFoundError: 400: Invalid model name passed in model=qwen3.8-27b.
POST /v1/chat/completions HTTP/1.1" 400 Bad Request
```

The only `qwen3.8-*` entries were `qwen3.8-max` / `-max-nothink`, both pointing at Alibaba's cloud endpoint, not the rig. When the live scene moved to `qwen3.8-27b` on `:8091` during the dual-max promotion, the gateway catalog was never updated alongside it.

**Impact:** a 400 on an unknown model name is deterministic, so retries cannot help — it surfaces to clients as an upstream *provider* failure while the model is perfectly healthy. Direct to `:8091` throughout: 82 tok in 1.28 s (~64 tok/s), `finish_reason=stop`. Cost of the fix is two routes; trade-off is none, since `:8091` is a registry `default_port` for this model.

**Deliberately narrow.** The other seven local backends point at currently-closed ports, and I checked each against `scripts/lib/profiles/compose_registry.py` before touching anything:

| model_name | port | valid registry `default_port`? |
|---|---|---|
| `deepseek-v4-flash` | 8030 | ✅ |
| `qwen3.6-27b` / `-autoround` | 8010 | ✅ |
| `agents-a1` | 8072 | ✅ |
| `qwen3.6-35b-a3b-autoround` | 8051 | ✅ |
| `gemma-4-31b` / `-autoround` | 8032 | ✅ |
| `gemma-4-12b-int8` | 8038 | ✅ |
| `deckard-40b` | 8199 | ✅ |

They are **inactive scenes, not stale routes** — one scene occupies the cards at a time and the gateway lists the full catalog regardless, exactly as the config's own header comment states. Removing them would break scene switching, so this PR leaves them alone.

## Type of change

- [ ] New compose variant
- [ ] New patch / sidecar
- [ ] New script or tool
- [ ] New model
- [ ] Doc-only / typo
- [x] **Other** — service gateway config (`services/litellm/config.yaml`), 19 lines added, no compose/model/script touched.

## Verification

Gateway probes before and after, all through `:4000` with the master key:

| Probe | Before | After |
|---|---|---|
| `qwen3.8-27b` (the call that failed) | `400 Invalid model name` | **`200` · `'GATEWAY OK'` · 0.78 s** |
| `qwen3.8-27b-fp8` | not registered | **`200` · `'Hi there! How can'`** |
| `qwen3.6-27b` (inactive scene) | `500` connection error | **`500` connection error — unchanged** |

That last row is the regression check that matters: the inactive route still resolves its *model group* and fails only on the closed backend, confirming scene switching is untouched. YAML parses clean (13 models).

### N/A justifications

- **Full rig + validation report** — N/A. No compose, engine, model, or serving flag changed; `report.sh --full` measures a serving path this PR does not touch. The model's own health was confirmed independently (direct `:8091`, ~64 tok/s) and is unchanged by a gateway route.
- **Profile header** — N/A, no compose touched.
- **BENCHMARKS row** — N/A, no TPS-affecting change. Routing through the gateway adds no measurable overhead to a decode-bound request.
- **CHANGELOG entry** — N/A for `models/<model>/CHANGELOG.md`; this is service-level config, not a per-model change. Happy to add a cross-cutting `CHANGELOG.md` entry if maintainers prefer.

## Cross-links

- Closes #1062
- Related upstream: none

### Note for maintainers

The branch name predates the content — it was cut for the dual-max promotion, which has since landed on master via #1032. This PR carries only the gateway fix.

Issue #1062 also flags a wider pattern worth a separate decision: there are three independently hand-maintained copies of "where is the model" (`config.yaml` here, `hermes-autoselect.sh`'s `SCENE_PORTS`, `gpu-mode.sh status`), none derived from `docker ps` or the registry. Each drifts silently and fails looking like something else. Not addressed here.

### DIFF
--- a/services/litellm/config.yaml
+++ b/services/litellm/config.yaml
@@ -31,6 +31,25 @@ model_list:
       api_base: http://host.docker.internal:8010/v1
       api_key: EMPTY
 
+  # Qwen 3.8 27B — dual-max fp8 KV + MTP n=3, 262K. Started by
+  # club3090-vllm-dual.service (drop-in zzzz-qwen38-27b-fp8-current.conf) →
+  # vllm-qwen38-27b-dual-max at :8091, which is a registry default_port for this
+  # model. Serves TWO ids: the neutral primary below + the quant alias after it.
+  # Added 2026-08-19 (#1062): the scene moved here during the dual-max promotion
+  # but the gateway was never told, so every client got a deterministic 400.
+  - model_name: qwen3.8-27b
+    litellm_params:
+      model: openai/qwen3.8-27b
+      api_base: http://host.docker.internal:8091/v1
+      api_key: EMPTY
+
+  # Quant-specific alias (same container, second --served-model-name).
+  - model_name: qwen3.8-27b-fp8
+    litellm_params:
+      model: openai/qwen3.8-27b-fp8
+      api_base: http://host.docker.internal:8091/v1
+      api_key: EMPTY
+
   # Agents-A1 — InternScience 35B agentic MoE (⚠️ caveats; the agentic thinking-ON
   # specialist). Launched via `switch.sh vllm/agents-a1-dual` → vllm-agents-a1-dual
   # at :8072 (own port — never share a sibling's; see the slug-masquerade note).


alexpolo1:
Superseded — closing.

`86aa73f5` ("fix(litellm): register local qwen3.8-27b routes on :8091 (#1062)") landed the same two routes on master, and `df2bdcdb` added the gateway-ports-resolve-to-registry gate on top. Same fix, same port, arrived independently; this branch would now just conflict on `services/litellm/config.yaml`.

Nothing here is left unaddressed — the diff was the `qwen3.8-27b` + `qwen3.8-27b-fp8` routes at `:8091`, both present on master, and the seven inactive-scene backends this PR deliberately left alone are still untouched there.

The durable follow-up is #1078 (derive the gateway config from the registry), which removes the class rather than the instance.

Thanks for the parity test — that's the part I didn't think to add.

PR #1094, opened 2026-08-23
fix(qwen3.8-27b): accept reasoning_effort "high" on the vLLM slugs

leo-3889:
## Summary

`reasoning_effort: "high"` returns HTTP 500 on all 20 Qwen3.8 vLLM slugs. Qwen3.8's ladder is `low`/`medium`/`xhigh` — there's no `high` — and the stock chat template `raise_exception()`s on anything else. OpenAI and Anthropic both use `low`/`medium`/`high`, so any generic client sending the standard top rung hits it. I found it with a Claude-API client on `/v1/messages`, which turns thinking on and sends `high` on every request.

Worth noting how it fails: `/v1/models` and `/health` stay 200 and the model serves fine, so the endpoint looks healthy while every thinking-mode request errors. #1029's server-side default doesn't help — a caller that sends its own value wins.

The fix is the model's own template plus seven lines mapping `high` → `medium`, vendored under `vllm/patches/`. I picked `medium` over `xhigh` because `medium` injects nothing, while `xhigh` adds the "think carefully, validate assumptions" preamble — and since a Claude-API client sends `high` on *every* request, mapping to the top rung would quietly push all agent traffic into the slowest mode on slugs shipping `max_num_seqs=1`. `xhigh` still works by name; every other rung renders byte-identically.

It's mounted over the checkpoint's own `chat_template.jinja` (vLLM auto-loads it) rather than passed with `--chat-template`, so no compose `command:` list changes — same wiring as the carnice override. One file covers all 20 slugs: the template is the same 8,952-byte file, sha256 `c3cf9e34…`, on FP8, Frozenlock INT4 and RadixArk NVFP4.

## Verification

Rendered stock vs vendored through a jinja2 sandbox across the whole ladder: stock raises on `high`, vendored serves it as medium; `xhigh`/`medium`/`low` identical on both, invalid still rejected, and with thinking off neither raises.

On the rig (2×3090, v0.27.1): the vendored file diffs against the checkpoint's own `chat_template.jinja` by exactly those seven lines, and a thinking-mode request with `reasoning_effort: "high"` returns 200 where it used to 500 — measured on `vllm/qwen38-27b-dual-superfast` with the mount wiring exactly as shipped here.

Guards green — `test-patch-attribution`, `test-compose-mounts-resolve`, `test-compose-no-repeated-args`, `test-generate-compose`, `test-compose-registry-disk`. CHANGELOG entry added.

No bench / soak / verify-stress: this is a chat-template mount, not a serving-config change — no engine flag, no memory knob — and every rung but `high` renders byte-identically, so there's no perf or quality claim to make. No BENCHMARKS row for the same reason.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -348,7 +348,7 @@ Tested via `URL=http://localhost:8004 MODEL=luce-dflash bash scripts/verify-stre
 
 ## Qwen3.8-27B
 
-Dense 27B, Qwen3-Next hybrid (16 full-attention + 48 linear-attention layers, 24 attention heads / 4 KV heads, head_dim 256, 262144 architectural ceiling). Same arch shape as Qwen3.6-27B, different checkpoint. Embedded MTP head. Native chat template (NOT froggeric).
+Dense 27B, Qwen3-Next hybrid (16 full-attention + 48 linear-attention layers, 24 attention heads / 4 KV heads, head_dim 256, 262144 architectural ceiling). Same arch shape as Qwen3.6-27B, different checkpoint. Embedded MTP head. Native chat template (NOT froggeric), vendored with a seven-line `high` -> `medium` reasoning-effort mapping — see `patches.yml` qwen38-reasoning-effort-template.
 
 ### Dual-card (2× RTX 3090, TP=2) — vLLM
 
--- a/docs/QUALITY_TEST.md
+++ b/docs/QUALITY_TEST.md
@@ -152,7 +152,7 @@ the three sampler vars by hand remains the only way to change them.
 | `--max-tokens 4096` | The ~1024 default silently truncates long answers into `token_limit` failures that look like wrong answers. |
 | `--thinking-max-tokens 16384` | The thinking arm needs the headroom. |
 | `--timeout-per-case 600` | 16,384 tokens takes real wall-clock. Too tight a cap produces `timeout` rows that read as content misses. |
-| `REASONING_EFFORT=` | Env only — forwarded as the per-request OpenAI `reasoning_effort`. The qwen3.8 composes default to `low` server-side since #1029; pin it anyway so the run records what it measured. ⚠️ Effort is **not** comparable across runs. ⚠️ `high` **raises** on vLLM — only `xhigh`/`medium`/`low` exist. |
+| `REASONING_EFFORT=` | Env only — forwarded as the per-request OpenAI `reasoning_effort`. The qwen3.8 composes default to `low` server-side since #1029; pin it anyway so the run records what it measured. ⚠️ Effort is **not** comparable across runs. ⚠️ Qwen3.8's own rungs are `xhigh`/`medium`/`low` — there is no `high`. On the **vLLM** slugs the vendored template (`qwen38-reasoning-effort-template`) maps `high` → `medium`, so it is accepted and records as medium — the un-nudged baseline, NOT xhigh; anything else still raises. Elsewhere `high` **raises**. |
 | `--repeat 3` | Add it for anything you'll quote. With `--sampling-from-server` both legs are sampled, so single draws aren't quotable. |
 
 ⚠️⚠️ **Read the failure modes before reading the score.** `timeout` and `token_limit` are **harness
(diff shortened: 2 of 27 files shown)


leo-3889:
Ran the two checks that were still missing when I opened this.

**`verify-full` against the live container** — reference 2×3090, v0.27.1, `vllm/qwen38-27b-dual-superfast`, mount wiring exactly as shipped here. All scored checks pass (Genesis check skipped — this compose is Genesis-free by design):

- tool calling → `tool_calls[]` populated with `get_weather`
- streaming tool-calls, thinking ON → `delta.tool_calls` + `finish_reason=tool_calls`, no `<tool_call>` leak
- thinking / reasoning mode → reasoning and content both populated
- 2K-token completion → 9,767 chars, variety 0.660, `max_line_repeat=0`
- MTP acceptance length 3.58

The first three are the ones that matter for a template change: the mounted file renders the tool-call XML and the reasoning delimiters exactly as the stock one did.

**Guard suite on the same box** — 118 pass / 2 fail / 0 timeout. Both failures are local rig state rather than this branch, and neither test references qwen3.8 or the chat-template machinery:

- `test-default-resolver` — my `.env` pins `CLUB3090_DEFAULT_QWEN3_6_27B=vllm/qwen-27b-dual-max`, so the resolver returns the pin where the test expects `vllm/dual`. Same class as the pin sensitivity #1046 neutralised for `test-list-topology-filter`.
- `test-preflight-compose-deps` — expects a missing-model failure with `MODEL_DIR` unset, but preflight defaults to `$REPO/models-cache`, which exists on this box.

Happy to re-run anything else that would help.

noonghunna:
Merged — thank you, this is a genuinely valuable catch.

The part that makes it valuable isn't the seven lines, it's the failure shape you
identified: **`/v1/models` and `/health` stay 200 while every thinking-mode request
500s.** A broken endpoint that reports healthy is the worst kind, and it would have
gone on failing quietly for any client using the standard OpenAI/Anthropic
vocabulary. That it took a Claude-API client on `/v1/messages` to surface it says
how narrow our own testing vocabulary had become.

Two more things worth calling out:

- **You found the right layer.** Mounting over the checkpoint's own
  `chat_template.jinja` so vLLM auto-loads it — no compose `command:` changes, one
  8,952-byte file with a verified sha256 covering all 20 slugs across FP8, INT4 and
  NVFP4 — is exactly right, and much better than 20 separate `--chat-template` args.
- **You scoped the verification correctly.** Rendering stock vs vendored across the
  whole ladder is the proportionate check, and explicitly arguing *why* no bench,
  soak or BENCHMARKS row applies — template mount, no engine flag, no memory knob,
  every other rung byte-identical — saved us asking. We reproduced your render
  matrix here and it holds.

### One change on merge: `high` → `xhigh` rather than `medium` (#1183)

Your reasoning for `medium` is sound and we took it seriously — `xhigh`'s preamble
*is* slow and timeout-prone, it *is* documented as such in our own compose headers,
and a Claude-API client really does send `high` on every request.

We went the other way for three reasons:

1. **Bijection.** `low`/`medium`/`high` → `low`/`medium`/`xhigh` keeps the ordering
   intact. Under the `medium` mapping two distinct client values collapse onto one
   server behaviour, and the top rung becomes unreachable through standard
   vocabulary at all.
2. **These aren't a scale with a middle.** `medium` has no template branch and
   injects *nothing*. So mapping the top rung there doesn't hand a caller one rung
   down — it hands them no reasoning instruction whatsoever, which is the opposite
   of what they asked for. (This is easy to miss from outside the repo; it's
   recorded in the compose headers, not the template.)
3. **The cost belongs in the default, not the vocabulary.** Traffic that sends no
   `reasoning_effort` is governed by the server default — `low`, via
   `--default-chat-template-kwargs` — and is untouched by this mapping. A caller
   that explicitly sends `high` has opted in, and one for whom `xhigh` is too slow
   can send `medium`. The server silently reinterpreting the word is the part we
   wanted to avoid.

Point 3 is really the crux: your concern is about *default* behaviour, and we
already answer that question in a different place. `PROVENANCE.md` now records both
the original reasoning and why it changed, so the trade-off stays visible rather
than looking like an unexplained edit.

Your follow-up framing is right too — the durable fix is upstream Qwen accepting
`high` as an alias, or vLLM's Anthropic router not forwarding a value it cannot
know is unsupported. The vendored template is the stopgap.

PR #1170, opened 2026-09-04
feat(qwen3.8-27b): dual-fast MAMBA_BLOCK_SIZE knob — recovers ~20% effective KV pool on hybrid-GDN (2×242K concurrent, 0 preemptions)

metaoutfitter:
## Summary

The KV-pool concurrency figures for the Qwen3.x hybrid-GDN slugs (e.g. `dual-fast`'s "652,346 → 2.49×", "567,737 → 2.17×") are ~20% optimistic in practice, and there's a one-flag fix.

**Why.** vLLM's boot-log "GPU KV cache size: N tokens" counts attention KV only. In `--mamba-cache-mode align` (shipped default) the GDN/SSM state is checkpointed at **every attention block** — 1,616 tokens on this slug, because align pads the attention block up to the mamba page size — and those state pages are allocated from the *same* pool at runtime without being reflected in N. Two ~200K prompts filled a "530,081-token" pool (`vllm:kv_cache_usage_perc` → 0.98) at ~400K live tokens, preempted twice and serialized.

**Fix.** `--mamba-block-size 8192` checkpoints the state every 8,192 tokens instead of every 1,616 (≈5× fewer state pages per sequence). Same boot, same nominal pool: **2×200K and 2×242K both run fully concurrent with zero preemptions**, prefix caching intact.

This PR wires that as an entrypoint knob (`MAMBA_BLOCK_SIZE`, off by default → upstream behaviour unchanged), adds `MAMBA_CACHE_DTYPE` alongside it, ports the `ASYNC_SCHED=off` → `--no-async-scheduling` toggle from #1139 to this compose (it vendors both crash patches but was missing the vllm#50021 mitigation), adds a header note next to `--mamba-cache-mode`, and a dated CHANGELOG entry.

## Measurements (ref 2×3090 PCIe, `vllm/vllm-openai:v0.27.1`, `dual-fast` W4A8, MTP n=4, util 0.92, 4 seqs, live `kv_cache_usage_perc` sampled every 15 s)

| Config | nominal pool | 2×200K | 2×242K |
|---|---|---|---|
| shipped (`align`, 1,616-tok blocks) | 530,081 (2.02×) | **2 preemptions**, serialized (TTFT 292 s / 481 s), KV peak 0.98 | not attempted |
| `MAMBA_CACHE_MODE=none` | 530,081 | identical (no-op while prefix caching is on) | — |
[...]

PR #1198, opened 2026-09-06
docs(benchmarks): dual-fast + MAMBA_BLOCK_SIZE=8192 sub-row (ref 2×3090, rebench-full 2026-09-04)

metaoutfitter:
## Summary

The BENCHMARKS sub-row you said yes to on #1170. One line, inserted directly under the existing `dual-fast` row: same compose, same Frozenlock weights, plus `MAMBA_BLOCK_SIZE=8192` / util 0.92 / 4 seqs, measured on my 2×3090 (290 W/card) with `rebench-full.sh` on 2026-09-04 (tag `qwen38-dual-fast-mamba8192-20260904`).

Numbers are the ones already in the merged CHANGELOG entry, nothing new: bench **75.3 / 106.7**, prefill 1837 @10K / 1380 @90K, verify-stress 8/8 with the ladder to 240,662 tok, 8-pack thinking-low **128/150** (pass@3 138), soak PASS, and the 2×242K-concurrent / 0-preemption result that motivated the knob. I tacked one sentence onto the notes pointing at the tail-block sweep from the #1170 thread, so the "is there a per-turn cost" answer sits next to the row.

Attribution stays on my rig, per your note on #1182. It's a dated sub-row, so the canonical @noonghunna row above it is untouched.

## Type of change

- [x] Doc-only / typo

## Verification

- [x] Table renders (same column count as the surrounding rows)
- [x] No compose, script or status changes


noonghunna:
Merging — thanks, this is exactly the right shape for a community row.

Verified before merge: 9 columns matching both neighbours and the header, inserted directly under the existing `dual-fast` row with the canonical row untouched, attribution and power cap on your rig, dated sub-row. And every figure traces back to the gate run in #1170 — `75.3 / 106.7`, `1837 @10K`, `1380 @90K`, CV `3.9% / 5.2%`, `112 @90K`, `~22.2 GB/card`, `8/8`, `262144`. Promoting already-reviewed measurements rather than introducing new ones is the bar we want for BENCHMARKS rows, and this clears it.

**One provenance correction for the record**, since the record is the whole point of the row: the description says the numbers are "already in the merged CHANGELOG entry". They aren't — `CHANGELOG.md` is generated by git-cliff from commit subjects and never carries figures. The numbers live in **#1170** itself (the `rebench-full` gate line: *"verify-full ✓ · bench 75.3 / 106.7 · concurrency 1/2/4 + deep 4×10K PASS"*). Nothing wrong with the diff — but anyone auditing this row later would go to the CHANGELOG, find nothing, and have to guess. #1170 is the citation.

The underlying finding is the valuable part and it generalises beyond this slug: `align` mode checkpoints SSM state at every attention block, and those pages are invisible in the reported "GPU KV cache size", so the nominal-vs-effective pool gap is real on every Qwen3.x hybrid-GDN slug that ships `align`. Your point on #1170 that the catalog's "N× concurrency" claims could be re-derived with this knob is still open on our side and worth doing.

PR #1226, opened 2026-09-09
Qwen3.8-27B W4A8 + DFlash2 on SGLang v0.5.19 (TP4) — re-cut W4A8 patch + shape guard, engine A/B vs the vLLM dflash2 compose

A1RM4X:
## What this is

SGLang-side twin of the vLLM `a1rm4x/dflash2-qwen38-27b` compose: **Qwen3.8-27B,
Avuja AutoRound INT4, served W4A8 + the DFlash2 drafter on SGLang v0.5.19,
quad 3090 (TP=4)** — so the same model / quant / drafter can be served by
either engine and **A/B'd head-to-head**. This is an *engine comparison*, not a
"SGLang is better, switch" claim: SGLang wins **c=1 decode +43%**, the gap
narrows to ~5% by c=8 (vLLM's batching scales harder), and SGLang TTFT is
lower at 4K+ (5.0s vs 8.6s @16K). SpecDecode (Prometheus, `--enable-metrics`):
accept_len=5.65 tok/step, rate=66%. Read the BENCHMARKS row for the full curve.

**Draft** for the same reason the vLLM dflash2 PR was: the 4-card config is
community-validated-by-design (the maintainer rig has 2× 3090) — the first
community boot is the validation.

## The patch — and its real fix

`models/qwen3.8-27b/sglang/patches/sglang-autoround-w4a8-v0.5.19/`:
a re-cut of jb-seo's `0004-autoround-w4a8.patch` (cut against SGLang **v0.5.18**)
onto **v0.5.19** (first release with DFlash2 upstream, so one engine carries
W4A8 + DFlash2). The one real fix over jb-seo's cut is the **shape guard**: his
eligibility gate omitted the per-shard `K%128==0 / N%64==0` constraint that
`prepare_w4a8` hard-asserts. On the Qwen3.8-27B GDN merged `in_proj_ba` tensor
(N=96 → 48/rank at TP4, not 64-divisible) a naive W4A8 pass crashes in
`gptq_marlin_repack` with `size_n=48 is not divisible by 64`. The guard pre-checks
each shard and routes non-conformant layers to standard Marlin W4A16 (no crash).
Necessity proven by an isolation run (stock v0.5.19 without the guard crashes
identically). `install.sh` is idempotent + hard-fails on partial state.

**SGLang activation differs from vLLM:** vLLM gates W4A8 behind
`VLLM_MARLIN_INPUT_DTYPE=int8`; the SGLang patch flips
`experimental_w4a8 = True` unconditionally on the Marlin AutoRound path —
applying the patch *is* the activation.

## Checkpoint data fix (not a patch — flagged so it isn't lost)

Stock SGLang crashes on the Avuja AutoRound checkpoint at weight-load *even
without W4A8* (GDN `in_proj_a`/`b` are BF16 in the checkpoint but SGLang builds a
merged `in_proj_ba` routed through GPTQ-Marlin). The fix is data-side in the
target's `config.json`: `".*in_proj_ba.*": {"bits":16,"data_type":"fp"}` — Avuja's
checkpoint already carries it (backup `config.json.bak-sglang-test`). The patch
does **not** fix this; point a different AutoRound ckpt at the compose and it
crash-loops until you apply the same one-liner.

## Files

- `models/qwen3.8-27b/sglang/compose/multi4/autoround-int4/dflash2-w4a8.yml` —
  the compose (DFlash2 is **upstream** in v0.5.19, so no drafter overlay —
  unlike vLLM's 6-piece patch; W4A8 from the vendored patch; fp8 KV + bf16 SSM +
  mamba-84/max-running-16 sizing for c16; `--enable-metrics`). Pinned `lmsysorg/sglang:v0.5.19`.
- `models/qwen3.8-27b/sglang/README.md` — tree README (why SGLang, the engine A/B,
  quick recipe, what it's *not*).
- `models/qwen3.8-27b/CHANGELOG.md` — append-only entry (merges with the vLLM
  dflash2 entry when both land).
- `BENCHMARKS.md` — new SGLang quad-card row paired against the vLLM dflash2 run.
- `results/sglang-q38-ar-w4a8-dflash2-tp4-20260909/` — **raw bench-ultimate output**
  (SGLang 2026-09-09 with `--enable-metrics` + the paired vLLM 0.27.1 baseline),
  force-added past the `results/*` ignore rule (same convention as the 32
  already-committed `results/*.txt`).
- `scripts/tests/test-compose-registry-disk.sh` — compose-disk count for
  the new disk-only SGLang compose.

SGLang composes are **disk-only / not in the compose registry** here (same as the
parked `qwen3.6-27b/sglang` tree) — direct `docker compose -f` usage, no registry /
engine-profile / model-yml churn. No shared launcher files touched.

## Verification

- `docker compose config` parses clean; entrypoint `bash -n` passes; the `$$`-escaped
  block-scalar + env-forwarding mirrors the proven vLLM dflash2 compose exactly.
- All repo compose-structural tests pass (restart-policy, no-rig-paths,
  nvlink-escape, image-drift, mounts-resolve, no-repeated-args,
  gpu-mask-passthrough, registry-disk); plus docs-slugs / profiles-compat /
  default-resolver / registry-emit-no-yaml.
- The `in_proj_ba` data fix + shape guard are validated on a live TP4 boot
  (all quantized linears W4A8 with 0 skips; DFlash2 accept len ~4.4; coherent
  output).
- The 09-09 bench has `--enable-metrics` live (1,012 metric lines, 91 unique
  `sglang:` names). SpecDecode section: accept_len=5.65, rate=66%.

## Caveats (kept in the BENCHMARKS row, not papered over)

- c=1 code decode is short-completion (646–800 tok actual vs 800 target; the
  "write quicksort" prompt doesn't always fill max_tokens).
- c=8 codeln TTFT (1069 ms) and c=16 codeln TTFT (2356 ms) are single-outlier
  rounds (median is fine).
- NVLink is rig-specific: the bench rig sets `NCCL_P2P_LEVEL=NVL`; a PCIe-only
  rig must not.


noonghunna:
Tested this against a different checkpoint and topology and it holds — sharing because the deltas
land almost exactly on yours, and because one case your config can't reach turned out fine.

**Setup:** Frozenlock AutoRound INT4 (not Avuja), **2× 3090 TP=2** (not TP=4), stock
`lmsysorg/sglang:v0.5.19`, on both drafters.

Applies to v0.5.19 with **zero fuzz** — jb-seo's v0.5.18 cut needs `fuzz 2` on `auto_round.py`, so
the re-cut is doing real work. Engages correctly on a different export:

```
[Marlin] W4A8 prepared: weight_bits=4 activation_dtype=int8 output_dtype=torch.bfloat16
         scale_encoding=signed_float arch=sm86 K=5120 N=8192 group_size=128
```

`scale_encoding=signed_float` is the fold handling this checkpoint's **50.44% negative group
scales** (packing is `[k/8, n]` auto_round, same family as Avuja).

| | MTP tier | | DFlash2 tier | |
|---|---:|---:|---:|---:|
| | W4A16 → W4A8 | Δ | W4A16 → W4A8 | Δ |
| prefill @10K | 1183 → **1710** | **+44.5%** | 1234 → **1779** | **+44.1%** |
| prefill @90K | 953 → **1276** | +33.9% | 973 → **1345** | +38.3% |
| code decode | 112.5 → 100.0 | −11.1% | 171.2 → 166.8 | ~flat (CV 11%) |
| accept len | 2.86 → **2.86** | **=** | 3.71 → **3.72** | **=** |

+44% prefill on both, drafter untouched on both. Numerically clean: greedy `17*23 → 391`, capital
of France, `s[::-1]`, planets in order; verify-full all-pass; vision 4/4.

**The case your config can't reach — and it's fine.** This export **quantizes the MTP head** (29
`mtp.*` tensors with `qweight`) where Avuja's is bf16, and you run an external drafter so you never
touch it. Those shards divide cleanly (`5120→2560`, `17408→8704` at TP=2), so your shape guard does
**not** reject them — 520 layers prepared, zero fallbacks, MTP head loaded `quant=auto-round`.
**Acceptance stayed at exactly 2.86**, identical to W4A16. So the guard's pass-through is correct
here, but nobody had exercised it against a quantized draft head before.

**One suggestion for your README**, unrelated to the patch: your choice of `incoai/Qwen3.8-27B-DFlash2`
is load-bearing and it may not be obvious why. The `syvai/...-DFlash2-W4A16` drafter is the *same
weights* — identical parameter count, `dflash_config` and tensor names — but compressed-tensors
quantized, and on SGLang it silently collapses to **accept len 1.03 / rate 0.004**, dropping decode
to ~38 TPS (slower than no drafter) with **no error, no warning, and verify-full still reporting
"All checks passed"** — speculative decoding rejects bad drafts, so a dead drafter is slow but never
wrong, and verify-full's acceptance check skips on SGLang because it greps for vLLM's
`SpecDecoding metrics` line. On this engine the engine's own `accept len:` counter is the only
drafter-health signal. A line saying "use the BF16 drafter, not the quantized one" would save
someone a confusing afternoon.

We are **not** vendoring a competing patch — we removed jb-seo's cut from our branch in favour of
this one, since yours is cut against the pin we actually run and carries the shape guard his omits.
Our #1237 adds the `sgl/` registry slugs and compose tree with no W4A8 patch; wiring it behind a
`W4A8=0/1` toggle is queued behind this merging.

noonghunna:
@A1RM4X — this is landing, but split, and I want to be explicit about which parts and why rather
than just closing it.

## What we took, with your authorship intact

Both commits are cherry-picked, not rewritten — `git log` shows you as author on both:

- **`46da355a` — the patch, the installer, the patch README, and the `patches.yml` entry.** Taken
  verbatim. This is the part that matters and none of it is reproducible by paraphrase.
- **Your 2026-09-09 results** (the `--enable-metrics` run and its vLLM A/B pair), so the SpecDecode
  series has a home and the numbers are checkable against raw output rather than a table.

Your bench row is banked in `BENCHMARKS.md` under a new **Quad-card (4× RTX 3090, TP=4) — SGLang**
section, credited to you and linked back to this PR. It is the only W4A8 measurement on this engine
from a rig that can actually run TP=4 — the reference rig has two PCIe-only 3090s, so that row is
not something we could ever produce ourselves. The section says so explicitly, so nobody reads your
NVLink 4-card numbers as transferable to the dual tiers.

The **shape guard** is the part I want to call out specifically. jb-seo's eligibility gate omitted
the per-shard `K%128==0 / N%64==0` constraint that `prepare_w4a8` hard-asserts, and on Qwen3.8-27B's
merged GDN `in_proj_ba` (N=96 → 48/rank at TP=4) that is a hard crash in `gptq_marlin_repack`. You
proved it necessary by isolation — stock v0.5.19 without the guard crashes identically — which
exonerates the rest of the patch. That is the difference between "a re-cut" and "a working re-cut",
and the CHANGELOG entry says so under your name.

## What we didn't take, and why it isn't a judgement on it

**Your `multi4/autoround-int4/dflash2-w4a8.yml` compose.** Not because anything is wrong with it —
because it now overlaps a shape we shipped after you opened this. [#1237](https://github.com/noonghunna/club-3090/pull/1237)
landed eleven `sgl/` slugs, including `sgl/qwen38-27b-multi4-superfast` (TP=4, DFlash2, fp8 KV,
262K), and every one of them carries a `W4A8=1` toggle that mounts **your** patch at
`/etc/club3090/w4a8` and runs **your** installer. So the capability your compose delivers is now
reachable from the shipped catalog, and carrying a twelfth compose alongside would mean two files
that do the same thing and drift apart.

Concretely, what your compose does becomes:

```bash
W4A8=1 bash scripts/switch.sh --force sgl/qwen38-27b-multi4-superfast
```

Until this PR that toggle deliberately hard-failed with a message telling you to land #1226 rather
than silently serving W4A16. That guard stays; the message is now accurate instead of a TODO.

Same reasoning for `models/qwen3.8-27b/sglang/README.md` and the `test-compose-registry-disk` count
— both are compose-scoped, and there's no compose.

## On our rig

For what it's worth, on 2× 3090 PCIe your patch measures **+44.5% prefill@10K / +33.9%@90K for
−11.1% code decode** on `sgl/qwen38-27b-dual-fast`, and takes `dual-superfast` prefill to **1779
tok/s against vLLM's 1778** — parity, while still ahead on decode. Different rig, different
topology, same direction as yours. It is a trade rather than a free win, and the composes say that.

Also credited upstream of both of us: **@jb-seo**, whose v0.5.18 cut this re-cuts.

Closing in favour of #1248 — the patch itself lives on with your name on it.
Thank you for this, and for the isolation run in particular.

Issue #1238, opened 2026-09-11
[bug] CUDA illegal memory access on GLM-5.3-Flash (glm5next) long prefill at -ub 2048 ( see #28282 on llama.cpp)


gstephan:
### What happened?

I started my prefill benchmark ( which does 4k then 32k the 128k) and llama.cpp died:

3.39.917.680 I slot print_timing: id  0 | task 6 | prompt processing, n_tokens =   2036, progress = 0.50, t =  11.82 s / 172.21 tokens per second
/app/ggml/src/ggml-cuda/ggml-cuda.cu:110: CUDA error
3.41.987.749 E CUDA error: an illegal memory access was encountered
3.41.987.753 E   current device: 0, in function ggml_backend_cuda_synchronize at /app/ggml/src/ggml-cuda/ggml-cuda.cu:2598
3.41.987.754 E   cudaStreamSynchronize(cuda_ctx->stream())
libggml-base.so.0(+0x1c3b6)[0x7fd6cd9da3b6]
libggml-base.so.0(ggml_print_backtrace+0x21a)[0x7fd6cd9da83a]
libggml-base.so.0(ggml_abort+0x15b)[0x7fd6cd9daa1b]
/app/libggml-cuda.so(_Z15ggml_cuda_errorPKcS0_S0_iS0_+0xb5)[0x7fd6bfe3ba25]
/app/libggml-cuda.so(+0x27d3b8)[0x7fd6bfe433b8]
libggml-base.so.0(+0x3c5ad)[0x7fd6cd9fa5ad]
libllama.so.0(_ZN13llama_context13graph_computeEP11ggml_cgraphb+0xa1)[0x7fd6cdba5bd1]
libllama.so.0(_ZN13llama_context14process_ubatchERK12llama_ubatch14llm_graph_typeP22llama_memory_context_iR11ggml_status+0x118)[0x7fd6cdba9d58]
libllama.so.0(_ZN13llama_context6decodeERK11llama_batch+0x518)[0x7fd6cdbb0cf8]
libllama.so.0(llama_decode+0xf)[0x7fd6cdbb238f]
libllama-server-impl.so(_ZNSt17_Function_handlerIFvvEZN19server_context_impl6decodeERiiR11llama_batchEUlvE_E9_M_invokeERKSt9_Any_data+0x49)[0x7fd6ceb31829]
[...]

PR #1241, opened 2026-09-11
docs(glm-5.3-flash): flag that -ub 2048 can crash long prefill, keep it as default (#1238)

noonghunna:
Addresses the docs half of #1238. **Comment-only — 0 non-comment lines changed**, all six composes
still parse.

## What this does

Flags [ggml-org/llama.cpp#28282](https://github.com/ggml-org/llama.cpp/issues/28282) at the `-ub`
line of every affected GLM compose, so anyone changing that knob sees it.

## What this deliberately does NOT do

**`UBATCH:-2048` stays the default** (maintainer call). It is worth **~+24% decode** on the i-quant
arm — BENCHMARKS row 8 vs 9, `14.58 → 18.12` narrative — and the crash needs a long prompt to
trigger, so short- and mid-context work is unaffected. Dropping the default to 512 would tax every
user for a bug that hits a subset.

The alternative — silently shipping a config that reliably dies on long prefill with no warning —
was the actual problem. Now it is a documented trade with a one-env-var escape.

## Why it is not arch-specific

Three reports, two architectures: sm_120 Blackwell (upstream), and two dual-3090s (#1238 and
@Pix13 upstream). Independent of the DFlash2 drafter and of `--cache-reuse`. `-ub 512` (or 128) is
a confirmed full workaround.

## Scope

All six GLM composes carry `UBATCH:-2048`, **including both K-quant arms** — this is not limited to
the i-quant slug in the report:

```
dual/unsloth-ud-iq4xs/moecache.yml     dual/devquasar-q2k/{moecache,offload}.yml
dual/unsloth-ud-iq3xxs/moecache.yml    dual/devquasar-q3km/{moecache,offload}.yml
```

⚠️ Whether K-quant experts dodge the crash is **untested** — all three reports so far are i-quant.
Noted as such in the headers rather than implied either way.

Status stays `🧪 Experimental`, so no `Caveats:` line (the enum reserves that for ⚠️/🐣/👁️/⏸️/🗑️);
the note lives in the free-form body. `docs/UPSTREAM.md` gains the #28282 row.

Thanks @gstephan — the traceback and the upstream cross-reference made this quick to confirm.


eexxoo:
K-quants are affected, tested & updated #28282.

Issue #1265, opened 2026-09-12
report.sh: capture the engine's RESOLVED configuration (not the compose file) + diff vs the shipped recipe


noonghunna:
## The gap

`report.sh --full` captures the rig thoroughly — CPU, RAM, STREAM bandwidth, THP, GPUs, PCIe
topology, P2P, disk, container states, kv-calc calibration — and **nothing about the engine's
actual configuration.** The only compose-adjacent line it emits is the `docker compose` *version*.
It calls `docker inspect` eight times and never for `Config.Cmd`, `Config.Entrypoint`, or
`Config.Env`.

So when a report shows an unexpected number, we cannot tell whether the reporter ran a shipped
recipe or something they modified. If they used a club recipe we can go read it in the repo; the
moment they change anything, the report is unattributable.

**#1261 is exactly this.** @paulp83's `--full` report gave us a clean rig picture and told us
nothing about how the engine was launched. We only learned his flags because he happened to paste
his boot log into a comment — and three of the findings came from that paste, not the report:
`--mem-fraction-static 0.82` (upstream recipe: 0.91), `--chunked-prefill-size 2048` (recipe: 1024),
and the drafter identity. Without the paste there was no way to know.

## ⚠️ But do NOT capture the compose file

The obvious implementation — dump the compose with comments stripped — is both incomplete and
actively wrong. Three reasons:

1. **A compose file is a template, not a configuration.** Ours are almost entirely
   `${VAR:-default}`. The text shows *defaults*. A user who sets `MAX_MODEL_LEN=131072` in `.env`
   or their shell changes nothing in the file — so the exact case this issue exists to detect
   stays invisible.
2. **The launcher overrides the compose.** `switch.sh`/`launch.sh` resolve `VLLM_IMAGE` from
   `scripts/lib/profiles/engines/<engine>.yml` `install.spec` and **inject** it, overriding the
   compose's own `image:` default. A compose dump would therefore report the *wrong engine image*.
   This is already a documented trap on this stack ("a compose that says `:-v0.22.0` will still run
   the engine-profile's image under the launchers"); printing the compose text would institutionalise
   it in every report.
3. Stripping comments addresses neither problem.

## What to capture instead — the resolved reality

Three sources, in increasing order of authority:

**(a) What the container received** — `docker inspect`:
`.Config.Image`, `.Config.Entrypoint`, `.Config.Cmd`, `.Config.Env`. This is authoritative for
[...]

PR #1273, opened 2026-09-12
feat(engine): pin llama-cpp-local by digest (b10920); widen the image-drift gate to all engines

noonghunna:
Two changes that turned out inseparable: the mainline llama.cpp pin was ~700 builds stale, and
chasing it exposed a drift guard that only covered vLLM.

## The pin *mechanism* changed, not just the version

**Upstream stopped publishing per-build CUDA server images.** Probed 2026-09-12:

| tag | result |
|---|---|
| `server-cuda-b10236` (old pin) | **200** |
| `server-cuda-b10930 / b10917 / b10901 / b10883 / b10867 / b10850 / b10793` | **404** |
| `server-cuda12-b10917`, `full-cuda-b10917`, `light-cuda-b10917`, `cuda-b10917`, `b10917` | **404** |
| rolling `server-cuda`, `server-cuda12`, `server`, `full-cuda` | **200** |

That makes the profile's own rule unsatisfiable as written — *"always pin validated builds, NOT the
rolling tag — it regressed at b9282, #187"*. Pinning the rolling tag's **digest** keeps both
properties: immutable (a digest can't move under us, which is exactly the #187 hazard) *and*
validated (the labels still name the build). It's already the house pattern — `beellama-local` and
all three `llamacpp-club3090` profiles pin `@sha256:`.

```
ghcr.io/ggml-org/llama.cpp@sha256:6ac9215…   = build b10920, commit eafe15a5e
```

Engine is overlay-free (`vendored_overlays: []`), so **no patch re-validation** — the step that
dominates a vLLM bump.

## ⚠️ This nearly shipped a false upstream-regression report

First bench on b10920 came in at **36.7 narr / 42.4 code** against this slug's recorded b10236
baseline of **61.55 / 71.33**. −40% on both legs, tight CV, drafter accepting healthily (0.83–0.89,
mean len 2.65–2.76) but converting nothing — and the figure landed almost exactly on the baseline's
own **spec-OFF** number (n=0 → 42.5). It read exactly like an upstream spec-decode regression across
694 commits.

Cause: **`nvidia-power-cap.service`** — an *enabled* systemd unit capping both 3090s at **230 W**
against a 370 W default.

| arm | narrative | code |
|---|--:|--:|
| recorded b10236 baseline (power unannotated) | 61.55 | 71.33 |
| b10920 @ 230 W | 35.7–36.7 | 41.9–42.4 |
| **b10920 @ 370 W (default)** | **67.2** | **78.8** |

**A depth arm alone would not have caught it** — `SPEC_N=1` and `SPEC_N=2` measured identically,
which is what killed the depth hypothesis and pointed at the machine rather than the config.

⚠️ The deeper defect: the recorded baseline carries **no power annotation**, so it isn't reproducible
from its own entry — neighbouring BENCHMARKS rows *do* annotate ("1× 3090, 370 W"). And `report.sh`
captures no power state at all, same class as #1265.

## The missing gate

`test-compose-image-drift` matched only `${VLLM_IMAGE:-…}`, so **53 fixed literals** across
`LLAMACPP_IMAGE` / `SGLANG_IMAGE` / `IK_LLAMA_IMAGE` / `BEELLAMA_IMAGE` / `VLLM_OMNI_IMAGE` had no
drift guard at all — more unguarded than guarded. Found because
`qwen3.6-35b-a3b/.../morikomorizz-q6kp/mtp.yml` was still on `server-cuda-b9570` while its engine
moved to b10236 five weeks earlier (fixed here).

**Negative control first:** planted a stale `LLAMACPP_IMAGE` and `SGLANG_IMAGE` literal and confirmed
the **pre-widening gate reported `ok`**. After widening it reds on exactly those two and greens when
removed. Coverage **48 → 68** (`VLLM_IMAGE=48, SGLANG_IMAGE=11, LLAMACPP_IMAGE=9`).

The first widening threw **two false positives**: `ik-llama/ornith9b-single` and `ornith35b-dual`
carry `ghcr.io/ikawrakow/ik-llama-cpp` while registered `engine=llama-cpp-local`, because no
ik-llama engine profile exists — flagging them had no valid fix. Scoped the comparison to the **same
repository** (a different repo is a deliberately different engine image, not a stale tag) and report
them as an informational NOTE so the registry oddity stays visible. Precedent for narrowing: the
docs-slug gate had to be scoped after a wider version produced ~200 false positives.

## Validation

`llamacpp/qwen38-27b-single-iq4xs` (1× 3090, 370 W default): boot clean, digest confirmed via
`docker inspect`, VRAM 22,304 MiB vs baseline 22,290, `multimodal` intact, 3/3 warm-ups clean,
coherent prose + code, accept 0.83–0.89.

Targeted gates green: `profiles-compat`, `compose-image-drift`, `engine-pin-bump`,
`diagnose-profile`, `docs-slugs-resolve`, `preflight-compose-deps`, `submit-bench`. The last two
reference the llama.cpp pin literal but only as **synthetic fixtures**, so they needed no change —
checked rather than assumed.

⚠️ **1 of 30 slugs booted.** Only 2 are production — `llamacpp/deckard40B-dual-mtp` and
`llamacpp/tess-dual-mtp`, both `qwen35` arch, neither booted. 19 of the 30 are deprecated. Yesterday's
MTP KV-allocation fix (#28630) touches deepseek2/glm4moe/cohere2moe only, so neither production slug
needed a newer build than this.


noonghunna:
Full gate sweep also run after the fact: **152 passed, 0 failed**.

Worth noting it added nothing over the targeted seven cited in the description — the two tests that reference the llama.cpp pin literal (`test-preflight-compose-deps`, `test-submit-bench`) use it only as a synthetic fixture, and the widened gate has no dependents. Recording it for completeness, not as evidence the targeted set was insufficient.

PR #1274, opened 2026-09-12
feat(qwen38): use prebuilt FlashAttention FP8 KV in ultramax

AntonProkopyev:
The existing Qwen3.8 dual-FP8 `ultramax` profile now uses FP8 KV with FlashAttention for both target and DFlash2. It retains 262144 context and vision. The separate `hypermax` entry has been removed.

## Implementation

- `dflash2.yml` selects `FLASH_ATTN` through vLLM's plugin API. The plugin uses native FlashAttention metadata and KV writes. It does not edit or subclass the FlashInfer backend; its stock source SHA256 is unchanged in the running container.
- A small init image supplies two compiled CUDA libraries and a plugin wheel. The serving container verifies artifact identity, checksums, CUDA/PyTorch/Python/C++ ABI, SM and the metadata/layout interface, then installs offline. FA2 sources are not downloaded or compiled at startup. Normal vLLM torch.compile/Triton warmup remains.
- The image is public and pinned to `ghcr.io/antonprokopyev/fa2-fp8kv-sm86@sha256:da040941fa048fd5fdfce520503341c0beda4ce41436a1c1fecaf3f8a99777c7`. Anonymous pull and offline extraction into an empty directory passed. [Build and extraction instructions](https://github.com/AntonProkopyev/fa2-fp8kv-sm86/blob/35a20db/containers/README.md).
- `required_sm` remains a lower bound. Optional `supported_sm` gates the exact architecture set before launch. SM86 is GPU-tested; SM89/SM120 have compiled cubins but no physical-GPU validation. SM90/SM100 retain stock native FlashAttention and have no new GPU measurements here. Unsupported or mixed architectures are refused before weights load.
- PCIe uses NCCL with custom all-reduce disabled. NVLink keeps the detected native path. Entry-point tests cover PCIe, PCIe P2P and NVLink separately.

The CUDA source comparison lives in our kernel repository: the unmodified upstream FA2 revision `28e862d21806bc3580207aa0ad4e2759151e9827` is a submodule, alongside a [reviewable patch](https://github.com/AntonProkopyev/fa2-fp8kv-sm86/blob/0fa02cbb760fbcc4a94ba1ee376e89825f8f43f4/patches/fa2-fp8kv.patch). `check_upstream.py` applies the patch and verifies all 18 mapped files byte-for-byte. Four headers differ; twelve are unchanged; two CUDA entrypoints are additions. [Scope and mapping](https://github.com/AntonProkopyev/fa2-fp8kv-sm86/blob/0fa02cbb760fbcc4a94ba1ee376e89825f8f43f4/upstream/README.md).

## Acceptance and quality control

Four same-rig arms used identical weights, explicit FP8 target/draft KV, DFlash n=7, context, reservation, batch size and sampling. Stock FI received only a shared-layout repair; its attention execution and graph policy remained stock.

| Arm | Narrative / code decode tok/s | Medium quality |
|---|---:|---:|
| Stock layout control, eager | 31.03 / 58.46 | 64/75 |
| FLASH_ATTN plugin, eager | 32.73 / 59.68 | 64/75 |
| Stock layout control, compiled | 75.06 / 140.82 | 63/75 |
| FLASH_ATTN plugin, graphs | 98.61 / 191.17 | 63/75 |

The eager pair isolates the implementation more closely. The compiled pair also differs in graph capability; CPU artifact compilation overlapped part of the stock compiled benchmark. These figures do not establish a pure kernel speedup.

Whole-phase `/metrics` deltas did **not** reproduce doubled acceptance or a flat tail: narrative mean length was 3.02–3.10 across arms, code 5.63–5.76. Stock itself has a much longer tail on code than narrative. The report includes draft-step counts and all seven position numerators. The earlier isolated log windows lack enough matched raw evidence to assign an exact cause. Greedy outputs are not bit-identical (3/8 exact matches in the eager pair); equal pack totals do not prove quality equivalence.

## Memory and scheduling disclaimer

The context capacity comes from the fixed KV reservation and scheduler settings. It is not attributed to the FA2 kernels. `gpu_memory_utilization` does not constrain `kv_cache_memory_bytes`, so the ineffective utilization flag and registry value were removed.

`long_prefill_token_threshold=0` disables the separate long-prefill threshold; chunked prefill still uses the 2048-token batch budget. The profile serves one sequence and one image up to about 4 MP. The retained reservation is 6335076762 bytes/card, with a measured 267493-token pool and `max_model_len=262144`. Combined vision leaves 1021 MiB free, below the unchanged 1024 MiB gate; this remains an explicit experimental-profile limitation. Reductions of 64/32 MiB lost the near-limit prefix cache (266-second continuations instead of 8 seconds); an 8 MiB reduction preserved the cache but did not improve physical free memory. Those changes were not retained.

## Validation

The actual compose was booted through `switch.sh --force vllm/qwen38-27b-dual-ultramax` using the public image digest. Kernel reference checks, native KV-write checks and graph replay passed on SM86. Twelve artifact-consumer checks and three interconnect cases passed.

- Verify-full passed, including vision 4/4. Canonical 3-warm/5-measured benchmark: **103.34 / 191.91 decode tok/s**, **101.95 / 182.79 wall tok/s**, narrative/code; prefill **1754.63 at 10K / 1384.03 at 93K**. Two RTX 3090 cards, 250 W/card, proven PCIe P2P, no NVLink.
- **261000 text tokens** and **259872 combined vision tokens** passed with zero cached input; both vision follow-ups retained all five facts and reused 255440 cached tokens in about 8 seconds. Over-limit input returned HTTP 400.
- Stress completed every response probe and all ten fresh needle depths through **240662 tokens**. **It exited 1 solely for 1021 MiB free versus the unchanged 1024 MiB margin. This is not a full stress PASS.** Peak VRAM was **23106 / 23106 MiB** per card.
- Continuous soak **PASS**, 5 × 5 turns, zero errors or silent-empty responses, zero VRAM growth, 100.0% throughput retention. All 25 decode measurements came from engine counters, using current master's soak fix.
- Quality quick: **26/30**, ToolCall 13/15 and InstructFollow 13/15, valid thinking OFF. The controlled medium comparisons are above.

Catalog coverage: **156/162 pass** after affected-guard reruns and the two tests added by current master `c4ed9ca`. Five remaining capture/benchmark fixture failures reproduce on v0.11.0; `test-report-resolved-config` reproduces on clean master `2dcbb99`. All 15 final guards for the changed behavior pass. [Detailed comparison and validation scope](https://github.com/AntonProkopyev/club-3090/blob/099bbe43112cfe8b2e1c8adcb8b1c37c9f7d75a3/models/qwen3.8-27b/vllm/patches/fa2-fp8kv-sm86/REVIEW_VALIDATION.md).

The profile remains experimental. The adapter for vLLM 0.27.1 and the plugin for vLLM 0.29 are named by version and integration method. Their measurements are recorded separately.

Related discussion: https://github.com/noonghunna/club-3090/discussions/1076#discussioncomment-18353115


noonghunna:
Thanks for this — it's one of the most carefully validated submissions we've had, and we took it
seriously enough to reproduce it end to end. We built your kernels from an empty cache, booted your
compose, and benched it against a stock-config arm on the same rig, in the same session, with the
same instrument. **Your installer worked first time with no intervention.**

Two things stood out before we started: you disclosed the memory-margin failure instead of tuning
the threshold until it passed, and you said plainly that the quality numbers aren't a controlled A/B
because sampling was inherited. The installer is careful too — pinning vLLM 0.29.0 exactly,
SHA256-ing the stock `flashinfer.py` before touching it, SHA256-ing both archives, keying the build
cache on an identity hash. That's better hygiene than several of our own patches.

## What we measured

**Gates on your build, on our rig:** `verify-full` **10/10** including vision 4/4 on ground truth.
`soak-continuous` **PASS** — 25 turns, **0 MiB** VRAM growth against a 200 MiB threshold, 0 errors,
0/25 silent-empty, 100% TPS retention. Holding zero growth at ~350 MiB free is a real result for a
config this tight, and it confirms the stability claim in your PR.

**Canonical bench**, 3 warm + 5 measured, both arms on 2× RTX 3090 at stock power (370/420 W):

| | stock config-only | your build | delta |
|---|--:|--:|--:|
| narrative decode | 72.04 tok/s (CV 4.4%) | **91.48** (CV 2.1%) | **+27.0%** |
| code decode | 131.66 tok/s (CV 7.9%) | **176.91** (CV 7.0%) | **+34.4%** |
| TTFT | 161 ms | 152 ms | −5.6% |

A real, substantial decode win — measured on one rig with one instrument, so it isn't subject to the
cross-rig confounds in the PR's original comparison. Note it's somewhat below your 99.88/191.93, and
your cards ran at 250 W where ours ran at 370/420 W.

⚠️ One boot per arm. Our A/B noise is boot-to-boot (~2.8%), not run-to-run (~0.2%) — a gap this size
is far outside that, but treat the exact percentage as indicative.

## We could not test your prefill claim — that's our rig, not your code

We get 1248.92 tok/s @10K on **your** build and 1272.93 on stock; you report 1750.64. The reason is
almost certainly ours: our P2P grant is **REFUSED** (`topo -p2p` reports no all-pairs OK) where your
report says "PCIe P2P". TP=2 prefill is all-reduce heavy over large activation tensors, so without
peer transfers those bounce through host memory. Decode all-reduces tiny tensors and is far less
sensitive — which is exactly the pattern in our data.

The evidence that the wall is ours: **both arms land at ~1250-1270 @10K.** If your kernels were the
variable, they would differ. They don't. So we can neither confirm nor dispute 1750 — it isn't
reachable here for any configuration.

(For the record, your bounded-BF16 prefill path *did* engage — it logged once and never hit the
workspace-exhaustion fallback.)

## The finding that changed our view twice

**The 262K context does not need your kernels.** Taking our shipped `dual-supermax` (fp8 weights,
DFlash2 W4A16, fp8_e4m3 KV, 147,456 ctx) and changing only configuration — adding
`--kv-cache-memory-bytes 6335076762`, `--max-num-batched-tokens 2048`,
`--long-prefill-token-threshold 0`, `--max-model-len 262144` — stock vLLM boots at 262,144 with a
267,493-token pool and serves. A needle at **239,731 prompt tokens** returned correctly with
**343 MiB free per card** and zero preemptions.

That's a finding about *our* catalog, not a criticism of yours: our registry claimed 262K "needs
~0.99 util (unsafe)", which assumed util-driven sizing. `--kv-cache-memory-bytes` skips memory
profiling entirely, so our 144K cap was self-imposed and we'll fix it.

The KV pool is **identical in both arms** (267,493 tokens). Your kernels buy no capacity — the
capacity is the reservation. What they buy is decode speed.

**But two of your Python fixes are load-bearing against real 0.29.0 bugs.** We tried your flags
*without* your code: adding `"kv_cache_dtype":"fp8_e4m3"` and `"attention_backend":"FLASHINFER"` to
the drafter's speculative-config on stock crashes at engine init with

```
ValueError: KV cache layout has not been resolved yet; it is resolved once by the engine core
(resolve_kv_cache_layout) unless explicitly ...
```

— precisely what your adapter's comment describes ("the drafter's copied CacheConfig misses the
target's set_kv_cache_layout RPC") and what your `kv_cache_layout` property works around. Stock also
warns `CUDAGraphMode.FULL_AND_PIECEWISE is not supported with spec-decode for attention backend
FlashInferBackend`, which your `get_cudagraph_support` → `UNIFORM_BATCH` override handles. Your
branch with the identical flags came up clean. **Both are upstream defects independent of your CUDA
kernels, and both are worth reporting to vLLM on their own** — they'd help every DFlash2 user on
0.29.0.

## One thing we'd like you to explain

Draft acceptance differs far more than we expected between the arms:

```
your build:  acceptance length 6.40   avg draft acceptance 77.1%
             per-position 0.850, 0.800, 0.750, 0.750, 0.750, 0.750, 0.750
stock:       acceptance length 3.31   avg draft acceptance 33.0%
             per-position 0.564, 0.513, 0.385, 0.359, 0.282, 0.103, 0.103
```

Same drafter, same `num_spec_tokens=7`, same target weights. A flat ~0.75 tail across positions 3-7
is unlike anything in our records — acceptance normally decays with position, and our measured range
for this drafter is 1.84-3.41.

We are **not** claiming a defect. `verify-full` passed 10/10 including vision ground truth, and we
did not run a controlled quality comparison. But since the decode win and the acceptance jump arrive
together, and they can't be separated on stock (it crashes with those flags), we can't yet tell how
much of the +27% is kernel throughput versus the drafter simply working properly. If you have a
controlled quality arm, or an explanation for the flat tail, that would settle it.

Two smaller notes:
- `--kv-cache-memory-bytes` **skips memory profiling** and, in vLLM's words, "does not respect the
  gpu_memory_utilization config". So the `0.95` in your compose is inert and your registry entry
  advertising `mem_util: 0.95` is misleading. It also explains the 239 MiB minimum — nothing
  profiled the headroom, it was reserved by fiat.
- Your compose reads `--attention-backend FLASHINFER`, but that selects *your* backend, since the
  installer aliases `FlashInferBackend` to your class. A reader sees stock FlashInfer. Worth a
  comment at that line.

## What would need to change for this to merge here

**1. The arch gate excludes the two other GPUs with the identical gap.**

Your adapter gates on equality (`capability.major == 8 and capability.minor == 6`) and builds with
`TORCH_CUDA_ARCH_LIST: "8.6"`. We traced stock's rule for FlashAttention with quantized KV:

```python
(fa_version == 3 and is_device_capability_family(90)) or (fa_version == 4 and is_device_capability_family(100))
```

FA3 at `major == 9`, FA4 at `major == 10` ("restrict to SM100 for now"). Which gives:

| GPU | SM | FA | stock fp8-KV FlashAttention |
|---|---|---|---|
| RTX 3090 | 8.6 | v2 | NO |
| RTX 4090 | 8.9 | v2 | NO |
| RTX 5090 | 12.0 | v2 | NO |
| H100 | 9.0 | v3 | YES |
| B200 | 10.0 | v4 | YES |

fp8-KV FlashAttention is effectively datacenter-only and **every consumer card is locked out, not
just Ampere**. You filled a gap 4090 and 5090 owners have too, then closed it behind an equality
check. We realise this isn't a recompile — Ada stays on FA2 and is the plausible next target,
consumer Blackwell is real work, and your storage-side FP8 design is Ampere-shaped since Ada and
Blackwell have native FP8 tensor cores.

There's also a concrete bug today: you declare `required_sm: 8.6`, but here `required_sm` is a
**floor** — the gate computes `need_sm = max(engine.min_sm, registry.required_sm, arch rule)` and
refuses only *below* it. A 4090 or 5090 passes the gate, boots, then hits `NotImplementedError` at
attention time. No field expresses "exactly sm 8.6".

**2. Please host the compiled kernels yourself.**

Every compiled artifact we depend on enters as a digest-pinned, publisher-hosted image, including
third-party ones (`ghcr.io/anbeeld/beellama.cpp@sha256:…`, `lmcache/vllm-openai@sha256:…`,
`ghcr.io/ggml-org/llama.cpp@sha256:…`). Our supported `delivery_mechanism` values are
`chat_template`, `install_script`, `python_sidecar`, `site_package_overlay`, `none` — nothing fetches
remote source and compiles inside the serving container, and yours adds a first boot needing network
and a CUDA toolchain on every user's rig.

Your source *is* published and pinned, which we appreciate. Two gaps: no built artifact, and because
the repo is standalone rather than a fork of flash-attention there's no reviewable diff against
upstream FA2 — which matters for derived CUDA kernels.

There's also a coupling we'd be taking on: `if version != "0.29.0": raise SystemExit` plus a
`STOCK_SHA` pinned to the exact content of one upstream file. Fail-closed is right, but it means
every vLLM pin bump breaks the slug until you rebuild. A digest-pinned image doesn't fix that — it
freezes it.

## Upstream — worth considering, with the state of play stated honestly

We think the split-K work has a better long-term home upstream than in our catalog, but we want to
be straight about what that path currently looks like rather than just pointing you at it. We
documented this defect in September: FA2's paged path won't
split the verify step across KV length, so it launches ~batch×heads blocks and walks the context
serially — ×21.4 slower than FlashInfer at 32K and ×28.5 at 200K, at q=8 vs q=1. Two PRs are open
for it, [vllm-project/flash-attention#176](https://github.com/vllm-project/flash-attention/pull/176)
and [#190](https://github.com/vllm-project/flash-attention/pull/190); we reached the same fix, found
theirs more general, and closed ours.

Your adapter does it too:

```python
# These splits stay fixed across graph replays; lengths/pages are tensors.
splits = (128 if attn_metadata.causal else 32) if max_query <= 64 else 1
```

Two things make yours distinct: it's in the **fp8** path, which neither our probe nor those PRs
cover, and the split counts are **fixed so CUDA graphs can still replay**. Our 55 → 111 tok/s result
came from a probe container we explicitly told people not to replicate; yours is graph-safe by
construction.

If a variant of this landed upstream, the packaging question, the arch gate and the version coupling
would all disappear — it would arrive inside the vLLM image everyone already pins, and every
3090/4090/5090 owner would get it without knowing our repo exists.

⚠️ But the honest state of play: **#176 has been open since 2026-08-06 and #190 since 2026-09-01,
and neither has a single maintainer review comment.** The only substantive discussion on either is
between the two authors and us. So we are not going to tell you "take it upstream" as though that
were a quick or certain route — it may well be slow, and that is a real cost to weigh against just
shipping it yourself.

What does look low-cost: both authors are receptive (@Peuqui replied "both are welcome" and
identified exactly which test coverage is missing), and your fp8, graph-safe variant is directly
relevant to a question they are actively working on. A comment on those threads costs you little
and might accelerate all three of you. Entirely your call — we would support it however is useful,
and none of it is a condition of anything here.

Draft is the right state meanwhile, and your own read — a tested experimental configuration, not a
production default — matches ours.

noonghunna:
One follow-up on naming, since it affects how this would land rather than whether it does.

**It wouldn't ship as `hypermax` — it maps onto our existing `ultra` tier.** Our slug names are three
orthogonal axes, which we had never actually written down anywhere (your PR is what made us do it):

| axis | values |
|---|---|
| weights (suffix) | `*fast` = AutoRound int4 · `*max` = official fp8 |
| drafter (prefix) | bare = MTP built-in · `super*` / `ultra*` = DFlash2 external |
| attention (prefix) | bare + `super*` = FlashInfer · `ultra*` = **FlashAttention** |

KV is `fp8_e4m3` across the board. **`ultra` is tied to the attention backend, not to bf16 KV.**

The bf16 on today's `ultra*` slugs is an Ampere workaround, not a property of the tier — stock vLLM
won't take quantized KV on FlashAttention below Hopper, so bf16 is simply the only KV FlashAttention
will accept on sm_86. That is the constraint your kernels remove.

So your config — fp8 weights, DFlash2, fp8 KV, FlashAttention — isn't a new tier. It is **what
`dual-ultramax` becomes when fp8-KV FlashAttention is available on the hardware**. It would upgrade a
slug we already ship rather than adding one, and the same applies on the int4 side to
`dual-ultrafast`. Today those two sit at 65,536 and 204,800 purely because bf16 KV is twice the
bytes; on fp8 they should both reach 262,144.

One consequence you might appreciate: with `ultra` on fp8 KV, **`super` vs `ultra` becomes a clean
attention-backend A/B** — identical weights, drafter and KV, one variable. We published a
FlashInfer-vs-FlashAttention crossover in discussion #1076 that had to be heavily caveated precisely
because the two tiers differed in KV *and* backend *and* context envelope at once. Your work is what
would let that comparison finally be made honestly.

None of this changes anything in the previous comment — the image, the arch gate, the `required_sm`
floor-vs-equality bug and the upstream-FA2 diff are still the open items. This is just so you know
what the thing would be called if it gets there, and that it slots into an existing tier rather than
standing alone.

AntonProkopyev:
Updated in 099bbe43112cfe8b2e1c8adcb8b1c37c9f7d75a3 following your review.

- Folded the configuration into the existing `ultramax` slug and `dual/fp8/dflash2.yml`; removed the separate `hypermax` entry.
- Replaced the FlashInfer backend substitution with a native `FLASH_ATTN` plugin for both target and DFlash2. The running container's FlashInfer source remains unchanged.
- Published a digest-pinned kernel/plugin image. Anonymous pull and offline extraction passed; startup verifies artifact checksums, ABI and SM compatibility. FA2 no longer builds at startup; normal vLLM compilation warmup remains. [Artifact and build instructions](https://github.com/AntonProkopyev/fa2-fp8kv-sm86/blob/35a20db/containers/README.md).
- Kept `required_sm` as a floor and added an explicit supported-SM set. SM86 is GPU-tested; SM89/SM120 are compiled targets with runtime validation still pending. SM90/SM100 retain stock native FlashAttention without new GPU measurements here.
- Added the upstream FA2 submodule and a reviewable patch in the kernel repository. The comparison reproduces all 18 mapped files byte-for-byte.

The four-arm acceptance comparison used matching weights, FP8 target/draft KV, DFlash n=7 and sampling. Whole-phase counters did not reproduce doubled acceptance or a flat tail: mean length was 3.02–3.10 on narrative and 5.63–5.76 on code. Both eager arms scored 64/75; stock compiled and plugin graphs each scored 63/75. The report includes raw position counts, output differences and timing limitations, including CPU build overlap during part of the stock compiled benchmark.

The final profile retains 262144 context and vision. It measured **103.34 / 191.91 decode tok/s** and **1754.63 / 1384.03 prefill tok/s at 10K / 93K** on two 3090s at 250 W/card with PCIe P2P. Verify-full passed, 261K text and 260K combined vision passed with cached follow-ups, and continuous soak passed **25/25** with no errors or VRAM growth.

**The memory caveat remains:** stress passed every response probe through 240662 tokens but exited 1 for **1021 MiB free versus the unchanged 1024 MiB threshold**. The profile stays experimental. The fixed KV reservation is retained; the ineffective utilization setting is removed, and the scheduler/capacity disclaimer is explicit. Catalog coverage is 156/162, with the six remaining failures reproduced on baseline code.

[Full comparison and validation report](https://github.com/AntonProkopyev/club-3090/blob/099bbe43112cfe8b2e1c8adcb8b1c37c9f7d75a3/models/qwen3.8-27b/vllm/patches/fa2-fp8kv-sm86/REVIEW_VALIDATION.md).

noonghunna:
Merged — thank you. This is the most thoroughly answered review we have had.

One small thing we fixed ourselves rather than bounce back to you, since it was
a three-line matter and you have already round-tripped enough today.

## The one blocker, and what we changed

`test-fa2-artifact.sh` failed on our rig — 166 pass, 1 fail — and it was your
test, not the kernel or the compose.

The patch block applies left to right, so `pathlib.Path` becomes a *function*
before `patch("importlib.metadata.version", ...)` resolves its dotted target.
That resolution triggers the **first import** of `importlib.metadata`, whose
chain reaches `importlib.resources._common`, and that module's body runs
`@as_file.register(pathlib.Path)` — receiving a function where a class is
required:

```
TypeError: Invalid first argument to `register()`: <function local_path at 0x...>
```

So it passed only where something had already imported `importlib.metadata`,
which is exactly why it was green for you and red for us. Minimal reproduction,
no repo involved:

```python
import pathlib
from unittest.mock import patch
def local_path(*a, **k): pass
with patch('pathlib.Path', local_path), \
     patch('importlib.metadata.version', return_value='x'):
    pass                      # TypeError

import importlib.metadata     # add this and the same block passes
```

We pushed one commit to your branch importing `importlib.metadata` eagerly, with
a comment explaining why it must stay. **Nothing the test asserts changed** —
only the import order. Full suite green afterwards, and it is in the merge.

## On the rest — it is all answered

Every review point came back addressed, and several went further than asked:

- **The arch gate.** `required_sm` kept as a floor with an enforced `supported_sm`
  set, so 4090 and 5090 owners get the fix rather than being locked out by an
  equality check on 8.6. Including 9.0 and 10.0 is the right call too — those
  cards already get fp8-KV FlashAttention from stock, and excluding them would
  have blocked hardware that works.
- **Hosting.** Digest-pinned publisher image plus an artifact checksum verified at
  startup, wired as a one-shot service with
  `condition: service_completed_successfully` and mounted read-only. No user
  compiles anything now, and a failed extraction blocks the boot instead of
  silently falling back. That was the part we cared most about.
- **Reviewability.** The submodule plus patch, 18 files byte-for-byte, is what
  makes derived CUDA kernels auditable rather than trusted. Going from a
  standalone repo to a comparable diff was real work and it matters.
- **The acceptance anomaly.** You went and looked rather than waving it off, and
  the four-arm result not reproducing it is a better answer than an explanation
  would have been.
- **The memory caveat.** 1021 MiB against a 1024 MiB threshold, disclosed rather
  than rounded away, with the profile kept experimental. That is the right
  instinct and we would rather have it stated than smoothed.

Two things we are knowingly taking on, recorded so they are not a surprise later:
the fail-closed `0.29.0` pin means a vLLM bump needs a rebuild before the slug
serves again, and the prefill figures stay credited to your rig — ours is a VM
whose PCIe P2P aperture mirrors, so we cannot reproduce that arm and will not
present it as a house number.

## What happens next, and it is not a request

Your work lands `dual-ultramax` on the definition the `ultra` tier was always
meant to have: DFlash2 + fp8 KV + FlashAttn. **Five of the six ultra slugs are
still on the bf16-KV placeholder**, so we have opened #1308 to extend it:

| slug | TP | KV today |
|---|---|---|
| `dual-ultrafast` | 2 | bf16, and the only ultra slug capped below 262,144 (204,800) |
| `multi4-ultrafast` / `multi4-ultramax` | 4 | bf16 |
| `multi8-ultrafast` / `multi8-ultramax` | 8 | bf16 |

**To be clear, we are not asking you to do this.** Your PR is scoped to
`dual-ultramax` and does it completely; expanding the ask after you have answered
everything would be moving the goalposts. The kernel is a pinned artifact now, so
a second compose can reference it and this is ours to pick up.

We start with `dual-ultrafast` because it is the one we can genuinely validate —
TP=2, SM 8.6, and a KV/decode change rather than a prefill one, so our missing
P2P does not get in the way. The testable part: that 204,800 ceiling is probably
*caused* by the bf16 KV, so fp8 should lift it to 262,144 like its sibling. We
will report the number either way.

One question you may already know the answer to, and it gates the multi4/multi8
half: **has the plugin been exercised at TP>2?** Your validation is TP=2, and
under tensor parallelism the KV cache is sharded across ranks, through the path
the plugin sits in. If you have not tried it, that is a perfectly good answer —
we will keep those four slugs behind an explicit caveat until a 4-card owner can
exercise it, rather than asserting something neither of us has established.

PR #3, opened 2026-04-29
RunPod template: one-click deploy 1x/2x3090 with genesis & marlin patches baked in

abhinand5:
## Summary

A single Docker image + entrypoint that deploys the club-3090 Qwen3.6-27B stack on RunPod (or any cloud GPU rental) with zero manual setup.

- **Auto-detects GPU count** at boot — picks tools-text (75K, fp8, Genesis P64+PN8) on 1 GPU or dual.yml (262K, fp8, vision+tools, Genesis-less) on 2 GPUs
- **Genesis v7.62.x baked in** (curl+tar at pinned commit `917519b`) + **Marlin pad-sub-tile-n files** overlaid from the noonghunna fork
- **No manual model download** — vLLM pulls from HF on first boot to `HF_HOME=/workspace/hf_home`, persists across pod restarts
- All config overridable via env vars (`MODEL_NAME_OR_PATH`, `MAX_MODEL_LEN`, `GPU_MEMORY_UTILIZATION`, etc.)

## Why

The club-3090 stack works great for people who own RTX 3090s. But paying $1800 for 2x3090 upfront isn't the only path — lots of devs rent 3090s on RunPod at ~$0.40/hr. They should get the same tested, patched, verified stack without having to clone the repo, run setup.sh, clone Genesis, clone the Marlin fork, etc.

This PR gives them: pick a template, spin a pod, send an OpenAI-compat curl 5 minutes later. Same configs, same patches, same TPS numbers as the docker-compose files — just in a cloud-native image.

## What's new

```
runpod/
  Dockerfile          # FROM vllm nightly, layers in genesis + marlin + tolist
  entrypoint.sh       # GPU detection + auto-config + vllm serve
  download-model.sh   # optional standalone model download (not needed by default)
  README.md           # template setup instructions + env var reference
  .dockerignore
```

## Test plan

- [x] Docker image builds clean from repo root
- [x] Boot on 1×3090 RunPod pod — should pick tools-text path, download model, serve
- [x] Boot on 2×3090 RunPod pod — should pick dual.yml path, serve
- [x] Stop/start pod — model should persist on `/workspace/hf_home`, skip re-download
- [x] Sanity curl + tool call + streaming on both paths

Will update with live RunPod test results once pods spin up.

PR #1321, opened 2026-09-14
sglang/qwen3.8-27b: W8A8 TP=4 + DFlash2 compose (community, disk-only) — SGLang-side companion to discussion #1288

A1RM4X:
## What

Adds a **stock-SGLang v0.5.19 INT8 W8A8 TP=4 + DFlash2** compose for Qwen3.8-27B, as a **disk-only community submission** — not a shipped slug. It's the SGLang-side companion to the config benched in [discussion #1288](https://github.com/noonghunna/club-3090/discussions/1288).

This is a clean disk-only SGLang contribution: the W8A8 (Avesed) checkpoint auto-detects its own compressed-tensors config on the v0.5.19 pin, so **there is no patch overlay**. That means the only hard gate that changes is the disk-compose count (**137 → 138**). No registry, drafter, model, engine, or `patches.yml` wiring.

## Files

- `models/qwen3.8-27b/sglang/compose/multi4/w8a8/dflash2.yml` — new compose (stock SGLang, W8A8, DFlash2, HiCache L2, bf16 KV).
- `scripts/tests/test-compose-registry-disk.sh` — disk count 137 → 138.
- `BENCHMARKS.md` — W8A8 row in the SGLang quad-card (TP=4) section.
- `models/qwen3.8-27b/CHANGELOG.md` — append-only entry.

## Config (faithful to the live stack)

SGLang v0.5.19 · TP=4 · Avesed INT8 W8A8 · bf16 KV · 262144 ctx · mem-fraction 0.82 · mamba 80 · max-running 16 · DFlash2 unquantized drafter (8 draft tokens, `incoai/Qwen3.8-27B-DFlash2`) · HiCache L2 ratio 1.5 (page_first_direct / direct io, no L3) · `NCCL_P2P_LEVEL=SYS`.

**Naming note — why `w8a8` and not `int8`:** the repo's weight-variant dirs use the `wXaY` scheme (`w4a16`, `qat-w4a16`, `nvfp4`, `bf16`, `fp8`), and a bare `int8` would collide with the existing weight-only INT8 convention (`autoround-int8` = W8A16). W8A8 (activations quantized too) is a distinct thing, so `w8a8` is the correct, unambiguous name.

## Numbers — read these with two flags

From the `report-full` run, **discussion #1288 (2026-09-13, 220 W cap)**:

| metric | value |
| --- | --- |
| verify-full | 10/10 PASS |
| verify-stress | 8/8 (ceiling 91% of 262K, VRAM flat Δ0 MB) |
| soak / bench / agentic | PASS / PASS / PASS (12 turns) |
| narrative decode (c=1) | 146.6 TPS (CV 3.1%) |
| code decode (c=1) | 270.8 TPS (CV 5.9%) |
| prefill | ~5,058 t/s peak @4K · 3,029 @90K |
| DFlash2 accept | len 5.28–6.16 (mean 5.69) · rate 0.61–0.74 (mean 0.67) |
| bench-ultimate combined decode | 146.6 → 1,811 t/s (c=1 → c=16) · peak 1.605 tok/W @878 W |

**Flag 1 — power cap.** Every number is cap-bound at **220 W** (62% of the 350 W default), so none is comparable to a stock-power baseline.

**Flag 2 — interconnect.** `NCCL_P2P_LEVEL=SYS` is load-bearing: a single-variable NVL control ran ~1.5–1.8× slower on prefill, so the c=1 prefill figures are interconnect-driven and won't reproduce without matching interconnect.

## Status / validation

🧪 Experimental, **disk-only** (community submission). TP=4 can never boot on the maintainer rig (2× 3090), so it's community-validated by design — the first community boot on 4 cards is the validation.

Local test battery (run on this branch, all green): `test-compose-registry-disk` (138), `test-compose-no-rig-paths`, `test-compose-restart-policy`, `test-compose-no-repeated-args`, `test-compose-image-drift`, `test-patch-attribution`. Full battery (incl. the slow bench/network tests) not yet run.

## Drafter note

The compose defaults to the **unquantized `incoai/Qwen3.8-27B-DFlash2`** drafter — the one actually benched. It is HF-mirrored (byte-identical) at `z-lab/Qwen3.8-27B-DFlash2` but is **not** wired into this repo's `setup.sh`, so the compose's drafter-missing path points at a manual `huggingface-cli download` rather than a setup key.

---
_Draft PR — opened for review before going live. Built off current upstream `master` (6b0f471)._


noonghunna:
@A1RM4X — flagging a measurement from today that touches the HiCache half of this compose. **Not a blocker for the PR** (disk-only, gates pass, and an inert L2 costs correctness nothing), but it may mean the tier is not buying what the header implies.

## What we measured

On `sgl/qwen38-27b-dual-fast` with `--enable-hierarchical-cache`, a staged diagnosis — **two independent runs**, second with a flushed cache and sizing measured on the wire:

| stage | run 1 | run 2 |
|---|--:|--:|
| D→H backup (kv / mamba) | 32,700 / 16 | 52,604 / 26 ✅ |
| full-KV evicted | 644,810 | 353,357 ✅ |
| host drops | 0 | 0 ✅ |
| host occupancy | 87% | 63% (37% free) |
| **return: `device` / `host`** | **0 / 0** | **0 / 0** ⛔ |
| **H→D load-back (kv / mamba)** | **0 / 0** | **0 / 0** ⛔ |

Backed up ✓ → evicted from device ✓ → retained in host (zero drops) ✓ → **never matched on return** ✗.
The return was *slower* than the cold request. And `cached_tokens_total{cache_source="host"}` was
**never instantiated at all** — a labelled Prometheus counter only exports after first use, so the
host read path has not fired once.

Root cause is filed upstream: **[sgl-project/sglang#33713](https://github.com/sgl-project/sglang/issues/33713)** (open since 2026-08-05, no fix). `_evict_host_leaf` consults only `component_data[FULL].host_value`, so on device eviction a node whose MAMBA component already committed to host is **pruned from the tree wholesale** rather than downgraded to an evictable-host leaf. The host bytes survive; nothing remains to match them against.

⚠️ The issue says the older `hi_mamba_radix_cache` path is unaffected — that path **does not exist in v0.5.19** (`UnifiedRadixCache` replaced it), and `--radix-cache-backend` only takes externally-registered plugins (in-tree: `flexkv`), so there is no in-pin workaround.

## ⚠️ Four reasons this may not transfer to your config — I have NOT tested yours

| | ours | yours |
|---|---|---|
| topology | TP=2 | **TP=4** |
| io backend / layout | `kernel` / `page_first` | **`direct` / `page_first_direct`** |
| sizing | `--hicache-size 30` | **`--hicache-ratio 1.5`** |
| state knobs | defaults | `--max-mamba-cache-size 80`, `--mamba-max-states-per-path 1`, `--radix-eviction-policy slru` |

The upstream root cause is in the eviction path, which *reads* as topology- and io-backend-independent — but that is inference from source, not a measurement on your rig, and `page_first_direct`/`direct` is a genuinely different transfer path.

## One-command check on your box

After some real traffic that has forced eviction:

```bash
curl -s localhost:<port>/metrics | grep cached_tokens_total
```

If you only ever see `cache_source="device"` and **no `cache_source="host"` series appears at all**, the L2 tier has never served a token — the counter is not exported until its first use. Also worth a look:

```bash
curl -s localhost:<port>/metrics | grep -E 'load_back_tokens_total|hicache_(backup|dropped)_tokens_total|hicache_host_(used|total)_tokens'
```

`backup > 0` with `load_back` absent/zero and `dropped = 0` is the #33713 signature.

⚠️ Two gotchas if you do measure it: the `cache_type`-labelled counters are emitted by **every TP rank with identical labels** and Prometheus multiprocess sums them, so at TP=4 they arrive at **4× the truth** (the `_bytes_total` ones are genuinely additive — each rank moves its own shard). And the device/host split is **not** in `usage` — it needs `return_cached_tokens_details: true` and arrives in a separate `sglext` chunk with `choices: []` before the usage chunk.

## Why it's not a blocker

If HiCache is inert here it costs host RAM and buys nothing, but it does not affect correctness — your verify-full 10/10, stress 8/8 and soak results stand on their own, and the decode numbers are c=1 anyway. Worst case the header oversells one line. If your check shows host hits are real on the `direct` path at TP=4, that is a **more** interesting result than ours and worth its own note — it would localise #33713 to a specific transfer backend.

Separately: nice to see `--radix-eviction-policy slru` in the wild. ⚠️ Note `--mamba-max-states-per-path 1` has an unmeasured cost — upstream's help says the shallowest interior states beyond the cap are removed *while their full KV remains*, so a re-query from an interior point (compaction, truncation, branching) keeps KV with no usable state.

noonghunna:
⚠️⚠️ **Second finding, and this one is a correctness issue rather than a perf note — worth acting on before merge.**

This compose ships `--enable-mixed-chunk`, and an upstream bug filed **today** says that flag corrupts mamba radix-cache checkpoints on hybrid GDN models: **[sgl-project/sglang#39342](https://github.com/sgl-project/sglang/issues/39342)** (open, affects v0.5.16 and main).

## All four trigger conditions are met here

| upstream trigger | this compose |
|---|---|
| `--enable-mixed-chunk` | ✅ line in the launch args |
| `--mamba-radix-cache-strategy extra_buffer` | ✅ the default — not overridden here |
| hybrid GatedDeltaNet model | ✅ Qwen3.8-27B |
| prefill co-batching with decode | ✅ `--max-running-requests 16` |

## Mechanism (upstream's own summary)

1. `ScheduleBatch.prepare_for_extend` flips `req.mamba_next_track_idx` to the alternate ping-pong slot
2. `mix_with_running` → `merge_batch` **unconditionally drops the tracking info** (`self.mamba_track_indices = None`)
3. the backend then skips the checkpoint write, because the tracking mask is None
4. on completion the request **donates the unwritten slot to the radix cache**
5. subsequent requests restore **stale SSM state** from that slot

Reported impact: exact match **0.91 → 0.84–0.86** (~10.7% of datapoints differ), and p50 latency **~200 ms → 3–7 s**.

## ⚠️ Why the gate results here would not catch it

It only fires **when prefill requests co-batch with decoding requests**. The reported numbers are `c=1`, and verify-full/stress are effectively sequential — so a clean 10/10 and 8/8 is exactly what you would expect *even with the bug present*. It would surface as intermittent quality drift and latency spikes once real concurrent traffic hits a server running `--max-running-requests 16`. That combination — passing gates, degrading in production — is why I would not merge it as-is.

## Mitigation

Simplest is to **drop `--enable-mixed-chunk`**. Upstream's own suggested fix #3 is to *reject* the flag alongside `extra_buffer`/`extra_buffer_lazy`, so the combination is not intended to work.

⚠️ Switching `--mamba-radix-cache-strategy` to `no_buffer` would also dodge it, but it is not a free swap: the strategy sets **both** the per-request slot cost `k` (3 → 1) **and** the scheduler ratio `r` (5 → 3), which changes the auto-fit pool solve *and* your effective concurrency cap. Dropping mixed-chunk is the smaller change.

---

## ⭐ Separately — your config independently corroborates a formula we derived today

`--max-mamba-cache-size 80` with `--max-running-requests 16`: on the default `extra_buffer` strategy with overlap the scheduler ratio is `r = 5`, and `resolve_max_num_reqs()` caps concurrency at `K // r`.

**80 // 5 = 16** — exactly your setting.

We derived `r ∈ {5, 4, 3}` (extra_buffer / lazy / no_buffer) from `_calculate_mamba_ratio()` while chasing why a K=33 tier clamped a request for 8 down to 6, and it predicts every clamp we measured (K=55→8, K=33→6, K=21→4, K=8→1). Seeing your independent TP=4 config land on the same arithmetic is good corroboration — and if you picked 16 by reading back a clamp rather than by calculation, that is the same mechanism observed from the other direction.

⚠️ Worth knowing the corollary: on that formula **`--max-mamba-cache-size 80` is what buys you 16 concurrent requests**, not the other way round. If the pool solve ever returns less than 80 on a different rig, the effective cap drops with it silently — the `max_total_num_tokens=… max_running_requests=N …` boot line carries the value that actually took effect (`server_args=` echoes the request).

noonghunna:
**HiCache flags must come out (or default off); the rest of the PR proceeds — no hold on the whole thing.**

Provenance check first (new repo rule, AGENTS.md "Engine patch provenance — upstream first"): this PR vendors **no engine patch** (stock SGLang v0.5.19, W8A8 checkpoint auto-detects its compressed-tensors config), so provenance is not the gate here. The ask below is purely measurement.

The compose ships `--enable-hierarchical-cache` unconditionally, with a header claiming HiCache L2 is "verified-valid on this pin in source … no hicache × DFlash2 and no hicache × hybrid-GDN assert". We measured exactly this pin + model class twice (2026-09-14, second run with flushed cache; upstream sgl-project/sglang#33713): HiCache allocates host pools, completes backup for **both** pools (`hicache_backup_tokens_total` kv +32,700 / mamba +16), drops nothing (`hicache_dropped_tokens_total` 0), leaves host 37% free — and **never serves a host hit** (`load_back_tokens_total` 0/0, `cached_tokens_total{cache_source="host"}` never even instantiated). On hybrid GDN v0.5.19 the host tier is measured dead weight: `HICACHE_RATIO × device` of host RAM **per rank** for zero hits. Our shipped sglang composes carry a deliberate no-HiCache comment citing #33713 for exactly this reason.

The "verified-valid in source" reasoning is the trap we ourselves fell into and retracted last week (learnings/sglang-engine.md 2026-09-15): absence of an assert is not evidence of function, and the failing transition is runtime-visible only (counters + the boot line `impl=UnifiedRadixCache … hicache_attached=True`). Cross-engine note: the same write-only signature was just measured on vLLM's native offload backend for this model class (discussion #1327), so "host tier on hybrid GDN" is suspect engine-wide, not SGLang-specific.

Ask: drop the four HiCache lines (or gate them behind `HICACHE=1` default-off with a caveat comment citing #33713 + learnings/sglang-engine.md 2026-09-14), and correct the header claim. Note the #1288 numbers were benched with HiCache on but never served from host, so they stand as device-tier numbers; removing the flags frees host RAM and should not move them (a single confirmatory bench run is welcome but not blocking). Everything else — W8A8 TP=4 DFlash2, disk-only status, the two measurement flags (220 W cap, `NCCL_P2P_LEVEL=SYS`) — is independent and mergeable once the HiCache lines are gone.

PR #1324, opened 2026-09-15
fix(qwen3.8): vendor fa5017a5 so DFlash2 mamba-align offload/prefix hits work on vLLM 0.29.0

henrykrinkle01:
## What
Vendors vLLM commit `fa5017a5` — *"Scope prefix-cache last-block drop to eagle-family drafters (DFlash2 offload fix)"* — so DFlash2 prefix-cache / KV-offload hits work on the `vllm-stable` v0.29.0 pin. Wired into the **dual** `dflash2.yml` (ultramax) only.

## Why
`fa5017a5` landed on vLLM main **2026-09-10 — one day after the v0.29.0 cut**, and is not in any released tag (`v0.29.1rc0` is 27 commits behind it), so the running pin still has the bug.

**The bug (vllm#53505, closed 2026-08-30):** `use_eagle()` returns True for `dflash`/`dspark`, so `Scheduler._mamba_block_aligned_split()` backs the last cache position off by one block. The final block-aligned mamba snapshot then never materializes and **every prefix-cache and offload-tier lookup converges to 0** ("stores but never serves a hit"). This is the vLLM blocker for KV-cache offload (RAM / LMCache) with DFlash2. The fix scopes that back-off to eagle-family drafters (`eagle`/`eagle3`/`mtp`), which actually share the target's full-attention KV groups; DFlash/DSpark draft from their own KV and never write target blocks.

This is the same intent as the still-open PR #54165; `fa5017a5` is what actually shipped to main. Our `#48375` (`pr48375`) overlay is a different code path and **not** sufficient.

## Prior art (field evidence)
A community home-lab field-notes doc ([`uraniumchonk/vllm-hybrid-mamba-notes` — FIELD_NOTES](https://github.com/uraniumchonk/vllm-hybrid-mamba-notes/blob/main/FIELD_NOTES.md), last updated 2026-08-28) documents a **working DFlash2 + LMCache stack on vLLM 0.28.0**: 2×RTX 3090 TP2, DFlash2 K=7, LMCache 0.5.4rc5 (`LMCacheMPConnector`, `kv_role=kv_both`), retrieve-ON **clean across 360/360 agent turns** with decode matching the un-patched baseline. Two facts from it are directly relevant here:

- **Its 0.28.0 row was clean only because of a *local port of #54165*** — stock 0.28.0 full-salads on the first retrieve hit and stores 0 to the offload tier, which is exactly #53505. So the doc independently confirms #54165's mechanism (and therefore `fa5017a5`'s) is the necessary DFlash2 fix; it just predates the commit that shipped it to main.
- **It confirms `#48375` ("short cell only") is not sufficient** and that the 2026-08-21 "second bug" (multi-session near-degrade after LMCache's #4253) **is** #53505 — i.e. the same defect this patch removes.

Its LMCache-side config constraints (block-size / `max-num-batched-tokens` relationship, `--mamba-ssm-cache-dtype bfloat16`, `chunk-size` = multiple of block size, and not passing `--kv-offloading-*` on 0.28) are the follow-up once this lands — not part of this PR.

## Scope
- **Preliminary KV-cache-offload enablement.** This unblocks the offload/prefix-hit path for DFlash2 on 0.29.0. It is **not** LMCache config itself (block-size / `mamba-ssm-cache-dtype` / connector args are a follow-up once this lands).
- **Dual only.** `multi4`/`multi8` deliberately out of scope for now.

## Changes
| File | What |
|---|---|
| `patches/vllm-dflash2-mamba-align-offload/{dflash2-mamba-align-offload.patch,install.sh,_check_applied.py,README.md}` | The vendored patch (adapted for v0.29.0's still-inline drop: adds `use_eagle_preserves_target_kv_cache()` to `SpecDecodingConfig` + scopes the back-off to it) + idempotent, fail-closed installer |
| `compose/dual/fp8/dflash2.yml` | mount + entrypoint `install.sh` call (2 lines) |
| `scripts/lib/profiles/patches.yml` | registry entry (satisfies `test-patch-attribution.sh`) |
| `docs/UPSTREAM.md` | tracking row (AGENTS.md convention) |

## Verification
- Patch applies clean to pristine `vllm/vllm-openai:v0.29.0` (dry-run + real apply + `py_compile` both files + reverse dry-run), **and** sequentially after `vllm-dflash2-backport.patch` (real entrypoint order).
- `test-patch-attribution.sh` **PASS** (78 entries); compose + `patches.yml` YAML parse.
- **Live-boot clean on the reference rig** (2×3090, dual ultramax): no salad, prefix-cache hits non-zero, DFlash2 acceptance healthy (smooth per-position decay).

## Drop-when
A released vLLM tag contains `fa5017a5` (watch v0.29.1 / v0.30.0). Confirm `use_eagle_preserves_target_kv_cache()` is present in that tag's `config/speculative.py`, re-run a DFlash2 prefix-hit probe (assert `vllm:prefix_cache_hits_total`), then remove the mount + entrypoint call and retire the row.


noonghunna:
**Status update after discussion #1327 — not on hold; the "offload" claim needs scoping before merge.**

Provenance check first (new repo rule, AGENTS.md "Engine patch provenance — upstream first"): this PR vendors upstream commit `fa5017a5` (on vLLM main 2026-09-10; same intent as still-open vllm#54165) → standard path, mergeable. No hold.

#1327 (2026-09-15) ran a forced-eviction control test on a base that *includes this patch*: native `--kv-offloading-backend native` stored +35 GB / +1,320 GPU→CPU ops and served **zero** promotions back (`external_prefix_cache_hits_total` +0, probe `cached_tokens` 0, 79.5 s cold prefill = no speedup). Two consequences:

1. **This patch is not sufficient for the native offload backend on v0.29.0.** The load-path defect lives elsewhere (OffloadingConnector lookup for hybrids) and fa5017a5 doesn't reach it.
2. **This patch is still load-bearing and should merge.** Its actual defect (vllm#53505: `use_eagle()` true for DFlash → last-block back-off → *every* prefix lookup converges to 0) is precisely what makes #1327's ambient **76% GPU prefix-cache hit rate** possible on DFlash2+align. Without fa5017a5 that number is ~0. Holding this PR because an unrelated connector is write-only would remove a working prefix-cache fix.

Requested changes before merge: scope the title and the UPSTREAM.md row — the offload-tier promise holds for **LMCache-class connectors** (the field-notes evidence you cite: 0.28.0 + a #54165 port, 360/360 retrieve-clean), **not** for `--kv-offloading-backend native` on v0.29.0 (measured write-only, #1327). Track the native-backend load path as its own upstream item; #1327's promotion-counter delta is the repro.

henrykrinkle01:
vLLM 0.30.0 is released, therefore close.

Issue #1142, opened 2026-09-02
BYOM: the local model layer works, but AGENTS.md routes agents to the core catalog — user models land in tracked files


noonghunna:
## TL;DR

The local model layer **already exists and works**. The problem is that our own canonical
docs don't route anyone to it: `AGENTS.md` → "Adding a model" has **zero** mentions of it, so an
agent helping a user add a model follows the curated-catalog steps and writes to
`scripts/lib/profiles/registry.yaml` — a **tracked** file. The user then gets a conflict on their
next `git pull`, which is exactly the failure the local layer was built to prevent.

This is a documentation-routing bug, not a missing feature.

## What already works (verified on `master` today)

| capability | where | verified how |
|---|---|---|
| Gitignored local layer | `scripts/lib/profiles-local/` — `*` with `!README.md`, `!.gitignore` | `git check-ignore` on `registry.local.json`, `models.d/<id>.yml`, `composes/<id>/…` — all IGNORED; the two structure files stay tracked |
| Merged catalog view | `get_registry()` = core + local | wired in `registry-emit.sh` (**both launchers**), `compat.py`, `generate_compose.py`, `litellm-emit.sh`, `diagnose_profile_cli.py`, `bench-row-formatter.sh` |
| Local writes are the **default** | `promote.py --layer local` is the default target | core needs **both** `--layer core` and `C3_ALLOW_CORE_PROMOTE=1` |
| Namespace + collision safety | `load_local_registry()` | refuses slugs outside `local/`, refuses slug or model-id collisions with core, raises `LocalRegistryError` loudly rather than serving a partial catalog |
| Can't hijack curated defaults | by construction | local entries never reach `DEFAULTS` / `ENGINE_PREFERENCE` / `RECOMMENDED_DEFAULT_MODELS` |

So a user's models genuinely survive `git pull` and branch switches today — **if** they find the path.

## The actual gap

| file | local-layer mentions | what a reader is told |
|---|--:|---|
| `AGENTS.md` → "Adding a model" | **0** | a binary: *"just serving"* (`pull.sh` / hand GGUF, no registry) **or** *"promoting into the curated catalog"* — `registry.yaml` entry, `models/<id>.yml`, `engines/*.yml`, `hardware/*.yml`, `patches.yml`, bump `test-compose-registry-disk` |
[...]

Issue #1078, opened 2026-08-21
Derive the LiteLLM gateway config from the registry (kill the where-is-the-model drift)


noonghunna:
## Context

#1062 (a locally-served model with no LiteLLM route → deterministic `400`) was one instance of a broader drift class @alexpolo1 flagged there: **three independently hand-maintained copies of "where is the model," none derived from the running container or the registry.** The LiteLLM gateway config (`services/litellm/config.yaml`) is the one that bit us — its local `api_base` ports and `model_name`s are hand-typed, so they drift from `compose_registry.py` (a missing entry → 400; a stale port → silent mis-route).

The immediate 400 is fixed (#1062), and a new gate — `scripts/tests/test-litellm-ports-resolve.sh` — now catches the **stale/typo'd/removed-slug port** half: every local gateway port must resolve to a real registry `default_port`. But it deliberately does **not** assert that every model *has* an entry (the "missing route" half), because the canonical port is scene-dependent and not every model belongs on the LAN gateway — a naive "every model → entry" gate would be a false-positive magnet.

## Proposal

**Generate the local block of `services/litellm/config.yaml` from `compose_registry.py`** so the missing-entry / stale-port class disappears entirely (no guessing, no gate needed for the generated half). Keep the **cloud** entries (DashScope MaaS, etc.) hand-maintained — only the local backends are registry-derived.

Sketch (mirrors the existing `registry-emit.sh` pattern):
- For each **gateway-eligible** model, emit one route: `model_name: <served-name>`, `api_base: http://host.docker.internal:<canonical default_port>/v1`, `api_key: EMPTY`.
- Ship a `--check` mode that fails if the checked-in config's local block ≠ the generated block (so it can't drift), same shape as the other registry-parity gates.

## Design questions to settle

1. **Gateway-eligibility:** a new registry flag (`gateway=True`), reuse `RECOMMENDED_DEFAULT_MODELS`, or "all non-deprecated with a default_port"? (The qwen3.8 tiers were incubating when #1062 hit — eligibility needs to be explicit, not "everything.")
2. **Canonical port per model:** which slug's `default_port`? Natural answer = the model's curated default (`model_default_target` / the `DEFAULTS` walk), so the route follows the recommended scene. Worth confirming that's always the port a gateway client should hit.
3. **Served-name → model_name:** the LiteLLM `model_name` is the `--served-model-name`, not the slug or model-id. The registry needs to expose the served-name (or we derive it) for the emit.
[...]

PR #1326, opened 2026-09-15
feat(litellm): route the gateway to a rig's pinned scene (#1325)

### LINKED ISSUE #1325: litellm-emit: let a rig route the gateway to its pinned scene without editing tracked files
## Problem

`litellm-emit.sh` (#1078) picks each gateway model's scene **only** from the core registry: the curated-default walk restricted to `gateway=True` entries. For `qwen3.8-27b` that is `vllm/qwen38-27b-dual-max` on `:8091`.

A rig that serves a *different* scene of the same model has no way to route the gateway to it without editing tracked files. #1078 left this open as design question 2 ("*worth confirming that's always the port a gateway client should hit*"). It isn't:

- My 2×3090 serves `vllm/qwen38-27b-dual-fast` (`:8113`) on purpose, for the KV headroom and concurrency on 24 GB cards. The user pin already records that (`switch.sh --set-default vllm/qwen38-27b-dual-fast` → `.env`).
- The gateway still routes `qwen3.8-27b` to `:8091`, so every client gets `HTTP 500 Connection error`.
- The only fix today is hand-editing `services/litellm/config.yaml` and/or `registry.yaml`. Both are tracked, so `git pull` autostashes or conflicts them. On my rig a nightly `git pull` stashed the edit away on 09-10, and the next LiteLLM restart took the gateway down. It is the same failure class #1142 fixed for user models ("user models land in tracked files").

## Proposal: a gitignored, pin-aware LOCAL gateway config (opt-in, zero change by default)

1. **`bash scripts/lib/litellm-emit.sh --local`** writes `services/litellm/config.local.yaml` (gitignored). It is identical to the tracked config except that, per gateway model, the scene follows the user's existing `.env` pin `CLUB3090_DEFAULT_<MODEL>` when that pin names a known slug of the **same model** with a `default_port`. Otherwise it falls back to the curated gateway scene with a stderr notice, same shape as the launcher's pin warnings.
   - The gateway honours the pin **regardless of status**, and annotates non-functional ones `# status: <status>` exactly as the generated block already does for dual-max. The gateway should route to what the rig *serves*; the launcher's functional-status gate answers a different question ("should I auto-launch this").
   - Eligibility is unchanged: a model is on the gateway only if some core entry is `gateway=True`. A pin can't add a model to the gateway, only choose which of its scenes.
2. **`services/litellm/docker-compose.yml`** mounts `${LITELLM_CONFIG:-./config.yaml}`. A rig opts in with `LITELLM_CONFIG=./config.local.yaml` in `services/litellm/.env` (already gitignored).
3. **Unchanged:** the tracked `config.yaml`, `--check` and every existing gate. Without `--local` the output is byte-identical.
4. **Test:** a new gate covering pin honoured → port + status annotation; unknown slug / wrong-model pin → fallback + notice; no pin → local output == tracked generated block; `--local` never touches the tracked config.
5. **Docs:** a short "serve a different scene behind the gateway" note next to the existing gateway row in `docs/ADDING_MODELS.md`.

Affected user surface: any rig whose served scene differs from the curated gateway scene (e.g. `dual-fast` vs `dual-max`, or a single-card rig of a dual-canonical model).

Happy to send the PR. Reference rig: 2× RTX 3090, PCIe, `vllm/qwen38-27b-dual-fast`.

alexpolo1:
Closes #1325.

## Why
`litellm-emit.sh` (#1078) takes each gateway model's scene only from the curated walk over `gateway=True` entries. A rig that serves a **different scene of the same model** has no project-native way to route the gateway there. On a 2×3090 serving `vllm/qwen38-27b-dual-fast` (`:8113`), the gateway sent `qwen3.8-27b` to dual-max `:8091`, so every client got `HTTP 500 Connection error`. The only workaround was hand-editing tracked files, which a nightly `git pull` autostashed. That turned the next LiteLLM restart into an outage. This is the gateway half of the problem #1142 fixed for models, and it answers #1078's open design question 2.

## What
- **`bash scripts/lib/litellm-emit.sh --local`** writes a **gitignored** `services/litellm/config.local.yaml`. It is the tracked config with each gateway model routed to the user's existing `.env` scene pin, `CLUB3090_DEFAULT_<MODEL>` (from `switch.sh --set-default`).
  - A pin is honoured if it names a known slug of the **same model** with a `default_port`, **whatever its status**. Non-functional scenes keep the `# status:` annotation. The gateway should route to what the rig *serves*; the launcher's functional-status gate answers a different question.
  - Invalid pins (unknown slug, another model's slug, no port) fall back to the curated scene with a stderr notice.
  - A pin never *adds* a model to the gateway. Eligibility stays `gateway=True`.
  - The pin is read from `ROOT/.env`, and the shell env wins (same precedence as the launchers). The write is atomic (temp file + `os.replace`).
  - `--local --check` gates the local file.
- **`services/litellm/docker-compose.yml`** mounts `${LITELLM_CONFIG:-./config.yaml}`. A rig opts in with `LITELLM_CONFIG=./config.local.yaml` in `services/litellm/.env` (already gitignored).
- **`.gitignore`** gains `services/litellm/config.local.yaml`.
- **`docs/ADDING_MODELS.md`**: the gateway row explains the pinned-scene path; the guard table lists the new gate.

**Unchanged:** without `--local`, output is byte-identical. The tracked `config.yaml` and its `--check` are untouched, so there's no change for rigs that don't opt in.

## Validation
- New **`scripts/tests/test-litellm-local-pin.sh`**, with a synthetic fixture plus a real-registry leg. It covers:
  - the pin honoured (port plus status annotation), and pins for one model not moving another
  - the `.env` pin read, with the shell env winning
  - three invalid-pin fallbacks, each with a notice
  - no pin: local block == tracked block
  - the tracked config never modified, and cloud/hand blocks carried over
  - `--local --check` passing when fresh and failing when stale
  - gitignore and compose wiring
  - real registry: `qwen3.8-27b` → dual-fast's port
- **Existing gates, all passing:** `test-litellm-generate`, `test-litellm-ports-resolve`, `test-default-url-gateway`, `test-no-bare-grep`, `test-locale-utf8`, `test-script-permissions`, `test-registry-emit-no-yaml`, `test-docs-slugs-resolve`, `test-docs-local-layer`.
- **On the rig** (2× RTX 3090, PCIe, serving `vllm/qwen38-27b-dual-fast`), `--local` output matches that rig's working hand-maintained gateway config **route for route**. `docker compose config` resolves the bind source to `config.yaml` by default and to `config.local.yaml` with `LITELLM_CONFIG` set.


noonghunna:
Closing this — and the reason is our fault, not yours.

Four days after you filed it, `ef2ab278` ("feat(litellm): serve the gateway from what is running, not from the catalog") landed and solved #1325 by a different route. Nobody said so here, and nobody closed #1325, so you had no signal your problem was already fixed. That's on us.

## What overtook it

`scripts/lib/litellm-sync.sh` renders `services/litellm/config.runtime.yaml` from **endpoints that actually answered**, and the compose now mounts that file by default:

```yaml
- ./config.runtime.yaml:/app/config.yaml
```

Same failure you diagnosed — its commit message describes it in almost your words: *"running a sibling slug for that model … advertises the model and dials a port with nothing on it."* The difference is where the truth comes from:

| | this PR | `ef2ab278` |
|---|---|---|
| route source | `CLUB3090_DEFAULT_<MODEL>` pin in `.env` | endpoints that responded |
| route names | registry `served_name` | the endpoint's own `/v1/models` |
| coverage | `gateway=True` models | every live model — **13 of 22 had no route at all** |
| activation | opt-in via `LITELLM_CONFIG` | default |

Liveness-derived turned out to be the more general cut: it needs no `.env` entry, and it covers the 72 of 138 variants that declare no `served_name` and so could never have been emitted from the registry.

## Why not just rebase this on top

I tried. The conflict is structural rather than textual — both changes own the same mount line, one wanting `${LITELLM_CONFIG:-./config.yaml}` and the other the runtime view. Carrying both would give two mechanisms racing for the same destination, plus `test-litellm-local-pin.sh` gating a pin path that no longer has a reason to exist.

## What was good here

The design work stands on its own merits and some of it is worth stealing if the runtime view ever needs a manual override: honouring a pin **whatever its status** (the gateway should route to what the rig *serves* — the launcher's functional gate answers a different question), the three invalid-pin fallbacks each with a stderr notice, the atomic temp-file + `os.replace` write, and a test with both a synthetic fixture and a real-registry leg.

The "a pin never *adds* a model to the gateway" invariant in particular is a distinction `ef2ab278` had to make too, and got right independently.

Closing #1325 as well, which should have happened when `ef2ab278` merged. Sorry for the wasted cycles — the fix for this on our side is closing issues when the work that resolves them lands, rather than leaving a contributor to discover it in a rebase.

Issue #698, opened 2026-07-13
[bench] 2x RTX 3090 FE NVLink / Qwen3.6-27B AutoRound INT4 / vLLM TP=2


kevinb361:
# Rig + bench report

## Rig

- GPU: 2x NVIDIA GeForce RTX 3090 Founders Edition, 24 GiB each, NVLink, air-cooled in 4U rackmount chassis
- Host: ai
- Driver: 580.126.20
- Stock power cap: 350W per GPU
- Engine: vLLM Docker image `vllm/vllm-openai:v0.25.0`
- Model: local `/models/autoround` — Qwen3.6-27B AutoRound INT4
- Served model name: `deep`
- Quantization: `--quantization auto_round`, `--kv-cache-dtype fp8`, `--dtype float16`
- Context: `--max-model-len 262144`
- Runtime flags: `--tensor-parallel-size 2`, `--gpu-memory-utilization 0.90`, `--max-num-seqs 8`, `--max-num-batched-tokens 8192`, prefix caching + chunked prefill enabled
- Spec decode: `--speculative-config '{"method":"mtp","num_speculative_tokens":3}'`
- Tool/reasoning: `--tool-call-parser qwen3_coder`, `--reasoning-parser qwen3`, default thinking off
- Endpoint during bench: `http://localhost:8006`, container `vllm-27b`
- club-3090 commit: `dadf2cc27a8caedd7edc0ef0c6f877e6bc044f95 (dadf2cc2 fix(compat): C5 wrongly rejects nvfp4 + fp8_e4m3 KV on Ampere (#686) (#694))`
- Isolation: central LiteLLM proxy was stopped during benchmarking so no other local services were routing traffic to the model.

## Summary

- `verify-full.sh`: PASS. MTP acceptance length = 2.94.
- `bench.sh`: completed.
- `soak-test.sh --continuous`: PASS, 0 errors, 0 silent-empty, 0 MiB VRAM growth.
- `power-cap-sweep.sh --cooling air`: completed and auto-reset to per-GPU stock. Observed decode-single TPS plateau around ~29 TPS across much of the sweep, with best TPS/W near 190-200W per GPU.
- Important comparability note: this is dual 3090 **with NVLink**, unlike the project reference rig's PCIe-only topology.

## Canonical bench summary

```text
=== summary [narrative] (n=5) ===
  wall_TPS       mean=  77.19   std=  2.17   CV= 2.8%   min=74.89   max=79.88
[...]

PR #1338, opened 2026-09-17
qwen3.8 ultrafast: refresh stale header, ship 262K default

TreyThomasCodes:
## What this fixes

`models/qwen3.8-27b/vllm/compose/dual/autoround-int4/dflash2.yml` (`vllm/qwen38-27b-dual-ultrafast`) tells three different stories after the 2026-09-14 #1274 change:

1. The at-a-glance header still says `KV: bf16 — MANDATORY here` / `Max ctx: 204800 (~200K)` — the pre-plugin tier.
2. The file's own 2026-09-14 W4A8 comment says "the shipped default is 262144".
3. The command block actually ships `MAX_MODEL_LEN:-204800` — a bf16-era leftover, so a bare launch today silently takes the *old* config (and a 262K user has to know to override).

Meanwhile the registry entry for the same slug already says `max_ctx: 262144` with a fully current `status_note`, and the BENCHMARKS 2026-09-14 row labels the 262144 config "SHIPPED CONFIG". So the compose file was the one surface still describing the world before #1274.

## Changes (one file, 35+/25-)

- **at-a-glance**: `KV:` → fp8 e4m3 via the #1274 FA2 plugin (with the before/after decode numbers kept in-line); `Max ctx:` → 262144 as the shipped default, with the 204800 bf16-era ceiling preserved as history; `Drafter:` notes the DFlash2 model classes have been native since v0.29.0 (vllm#52816), not a vendored backport; `Best for:` now quotes the shipped fp8 numbers (~220 code / ~123 narr) instead of the bf16-era 227/132.
- The "⭐ WHY bf16 KV + FLASH_ATTN (and why 200K, not 262K)" section is re-cast as a **dated HISTORY block** (2026-08-20 → 2026-09-14) — same facts, correctly framed, ending on what dissolved the trade.
- The **VALIDATED block** gains the 2026-09-14 line (four alternating boots, each arm within 1%: 122.7/219.5 @262144, pool 421,970 tok / 1.61×); the 2026-08-20 bf16 results are kept verbatim under their date.
- **Command block**: `MAX_MODEL_LEN` default `204800` → `262144` — the only non-comment change.

## Why the default flip is safe

- It's not a new config: 262144 + fp8 KV + FA2 plugin is exactly the 2026-09-14 measured/validated configuration on the reference rig, and it's what the registry already advertises. The flip only makes the bare launch match what the file, the registry, and BENCHMARKS all already call "shipped".
- Anyone pinning `MAX_MODEL_LEN` explicitly is unaffected.
- I couldn't run verify-full from the box this PR was authored on (no GPU); the 262144 config is the one the reference rig validated on 2026-09-14, so the gate question is only whether you want a fresh verify-full/soak pass on the reference rig before merge.
- Community cross-check (not part of the gate, just context): this exact config, including the new 262144 default, has been the daily-driver config on a 2× 3090 rig **with an NVLink bridge** (custom all-reduce engaged) since 2026-09-17 — 141 narr / 267 code TPS, prefill 3075 tok/s @10K (vs 1862 on the PCIe reference rig: the NVLink long-ctx win from the #698 notes materializing).

## Deliberately not touched

- `dual/autoround-int4/dflash2-fp8.yml`'s "fp8 KV can't use FLASH_ATTN on Ampere" line is *accurate for that file* — it's the no-plugin FlashInfer sibling, so it reads as the contrast case, not a stale claim.
- The multi4/multi8 ultrafast `status_note`s still carry the "bf16 KV + FLASH_ATTN is mandatory" pre-#1274 framing (their composes still default to bf16 KV, which at TP=4/8 fits 262K legitimately). Extending the plugin to those tiers is a config change, not a header fix — happy to do that as a follow-up if you agree the direction is right.


chriskerley78910:
@noonghunna @TreyThomasCodes 

 Can confirm. I ran into this as well, but you beat me to it in terms of putting up a PR for it.

Max model length should be set to 262k,  not 200k because the KVcahce has been switched from bf16 to fp8 and the full context fits now.

noonghunna (commented):
Ran this on the reference rig — that's the gate question you asked about, and you were right to ask for it rather than assume. Thanks also to @chriskerley78910 for the independent confirmation.

**Your three claims all verify exactly as stated** against `master`:

| claim | verified |
|---|---|
| compose ships `MAX_MODEL_LEN:-204800` | line 338 ✓ |
| …while its own comment says 262144 | line 61 ✓ — the file genuinely contradicts itself |
| registry already says `max_ctx: 262144` | ✓, alongside `kv_format: fp8_e4m3`, `tp: 2` |
| header still pre-#1274 | `KV: bf16 — MANDATORY`, `Max ctx: 204800` ✓ |

And the diff is exactly **one functional line**; everything else is comments:

```diff
-      - "${MAX_MODEL_LEN:-204800}"
+      - "${MAX_MODEL_LEN:-262144}"
```

## Reference-rig validation

Bare launch of your compose, `MAX_MODEL_LEN` deliberately **not** set — the claim under test is what a bare launch does, so overriding it would have proven nothing:

| check | result |
|---|---|
| boot | clean, `RestartCount=0` |
| engine-admitted length | **262144** — from the boot arg, the `Using max model len` line, *and* `/v1/models` |
| GPU KV cache | **400,095 tokens** |
| FA2 plugin | loaded, artifact `731d1942…` |
| **`verify-full`** | **rc=0** |
| VRAM | 20,654 MiB / card |

## The part that actually discriminates the two defaults

A boot at 262144 only proves vLLM accepted the number, so I ran a real prompt past the old ceiling:

```
prompt_tokens=240030   completion=32   wall=244s   RestartCount=0
```

**240,030 tokens served.** That is past the old `204800`, so it is something only the new default admits — the flip does what you say it does, and the pool holds it.

⚠️ My first attempt at this was worthless and worth flagging so nobody repeats it: I mis-sized the filler and got a 52,030-token prompt, which is *below* 204800 and therefore cannot distinguish the two defaults at all. It "passed" and meant nothing.

A retrieval datapoint is still re-running: my needle check returned `False`, but that was my harness, not the model — `max_tokens=32` truncated a **thinking** model mid-`<think>` before any answer could emerge. Re-running at 2048. That is beyond what this PR claims (declared length, not retrieval quality), so it does not gate the change; I will follow up with the number.

## One thing to hold off on

> *Extending the plugin to those tiers is a config change, not a header fix — happy to do that as a follow-up if you agree the direction is right.*

**Please don't, yet** — and this is the useful thing we can tell you that isn't visible from the PR. The `multi4`/`multi8` tiers already run the FA2 plugin, and it **crashes `EngineCore` at TP≥4**: `torch.ops.fa2_fp8kv.forward` raises `Sequence bounds exceed this port's supported envelope`, fatally, reproducibly. Tracked in **#1358**.

Worse, it fails *silently* — the container restarts, the harness retries, and the run publishes a plausible-but-tainted score. #1354 now fails on a `RestartCount` delta so that can't happen unnoticed, and #1356 caveats those composes.

Your `dual` tier is unaffected: TP=2 is inside the plugin's documented "TP=2 and one sequence" envelope, which is exactly why this validation is clean. Extending the plugin to `multi*` would be extending it into the crash.

So the direction is right for `dual` and currently wrong for `multi*` — through no fault of this PR.

## Verdict

The change is correct, minimal, and validated on the reference rig. The header was the last surface still describing the pre-#1274 world, and the default flip makes a bare launch match what the registry and BENCHMARKS already advertise.

noonghunna:
Follow-up on the retrieval caveat from my review — **it passes.**

Same 240K prompt, `max_tokens` raised 32 → 2048 so a thinking model can finish reasoning *and* answer:

```
prompt_tokens=240030   completion=114   finish=stop   wall=246s
answer after </think>: 'The reference codeword is **ZEPHYR-4471**.'
```

`finish=stop`, not `length` — so it completed on its own rather than being cut off. The needle was planted at the midpoint of the filler, so this is retrieval from ~120K deep, not from the edges.

The earlier `needle recalled: False` was entirely my harness: `max_tokens=32` truncated the model mid-`<think>` before any answer could emerge. Worth naming because it is a standing trap here — retrieval checks written for non-thinking models report a false negative against a reasoning model, and the failure looks like a model problem rather than a budget problem.

So the `262144` default is not merely accepted by the engine; it serves and retrieves at depth. No caveats left on my side.

PR #1341, opened 2026-09-18
litellm: emit hosted_vllm provider for local vLLM routes (fix reasoning_effort 400)

north1952:
## Summary

Self-hosted vLLM routes in the generated LOCAL block currently use the generic `openai/` provider prefix. For vLLM backends this drops `reasoning_effort` / `thinking` from litellm's param whitelist, so clients that send them get a `400` (or a silent downgrade). The dedicated `hosted_vllm` provider includes those params and passes them through.

## Changes

- `scripts/lib/litellm-emit.sh`: when a gateway entry's engine is vLLM, emit `hosted_vllm/<name>` instead of `openai/<name>`. llama.cpp backends keep `openai/<name>` because they do not parse `reasoning_effort` and rely on `drop_params` to discard it safely (the backend would otherwise reject it).
- `services/litellm/config.yaml`: regenerated LOCAL block -- all vLLM routes now use `hosted_vllm/`; the non-vLLM routes (deepseek-v4-flash, glm-5.3-flash, deckard-40b, qwen3.8-flash-next) stay `openai/`.

## Root cause

litellm maps `openai/qwen3.8-27b` to a vLLM `/v1/chat/completions` call but strips `reasoning_effort` before forwarding. A client requesting `reasoning_effort: medium` is therefore rejected with `400`. Switching the provider to `hosted_vllm` restores correct passthrough.

## Verification

- Before: `POST /v1/chat/completions` with `reasoning_effort: medium` through the gateway -> `400`.
- After: same request -> `200`, reasoning content returned.
- `litellm-emit.sh --check` passes (generated block matches the registry).

## Scope

Only the registry-emitted LOCAL block is changed. The hand-maintained `agents-a1` route and the cloud (DashScope) block are untouched.


noonghunna:
Merged — thank you. Clean diagnosis, and the fix is right.

I verified the mechanism directly against the LiteLLM we run rather than inferring it from the 400, and it's exactly as you describe:

```
provider      params  reasoning_effort  thinking
openai          31      False             False
hosted_vllm     33      True              True
vllm            33      True              True
```

So `openai/` genuinely does not whitelist the param, and every vLLM route on this gateway was silently dropping it. That lands particularly well here — we spent a chunk of this week establishing that `reasoning_effort` is client-side and has to survive the whole path to the server, and the gateway was quietly breaking that last hop.

`litellm-emit.sh --check` passes (13 routes from 9 gateway models), and the route split matches your description: 10 vLLM routes moved, the llama.cpp ones (`deckard-40b`, `deepseek-v4-flash`, `glm-5.3-flash`) and the exllamav3 one (`qwen3.8-flash-next`) stayed on `openai/`.

## Two notes, neither a blocker

**1. The diff touched more than the description says.** The body states the hand-maintained `agents-a1` route is untouched, but it was flipped too:

```diff
-      model: openai/agents-a1
+      model: hosted_vllm/agents-a1
```

That line sits at 111, **outside** the emitted LOCAL block (23–103). The change is *correct* — `agents-a1` is a vLLM backend (`switch.sh vllm/agents-a1-dual`), so it wants `hosted_vllm/` as much as the others — but because it's outside the emitted block, `--check` won't keep it in sync. Worth knowing it's now a hand-maintained line that has to be remembered, not a generated one.

**2. The llama.cpp reasoning about `openai/` doesn't hold for all of our llama.cpp backends.** You reason that they keep `openai/` because they "do not parse `reasoning_effort`". That's true for some, but not the prism/ternary tiers: those run `--jinja` with the Qwen3.8 template, which consumes `reasoning_effort` and remaps `high` → `xhigh`. We exercised exactly that path this week — an 8-pack at `reasoning_effort=xhigh` where the effort reached the server through the request body and demonstrably changed behaviour.

Since `hosted_vllm` is really just a param-whitelist profile over an OpenAI-compatible endpoint, it would pass through for those too. So the current split leaves gateway users unable to set effort on the ternary tiers. Not something this PR needs to solve — but if you wanted a follow-up, keying on "does this backend consume `reasoning_effort`" rather than "is this vLLM" would be the more accurate cut.

Nice catch on this one; it's the kind of bug that produces a confusing 400 far from its cause.

Issue #1027, opened 2026-08-16
[bench] Quality test - Qwen 3.8 27b fp8 - vLLM - dual 3090 - All thinking modes (instruct, low, medium, xhigh)


henrykrinkle01:
# TLDR: use low.

_- Mom, can I have a Claude subscription?_
_- We have Claude at home_

Claude at home:

<img width="1670" height="764" alt="Image" src="https://github.com/user-attachments/assets/7006c2e7-ad96-4cd6-bc8f-0e83987496c3" />

## Launching commands
```bash
# instruct
bash scripts/switch.sh --force vllm/qwen38-27b-dual-max
bash scripts/quality-test.sh --full --no-thinking --sampling-from-server \
  --max-tokens 4096 --thinking-max-tokens 65536--timeout-per-case 900 > quality-test-qwen3.8-27b-instruct.md

# xhigh, vllm reboot required
ENABLE_THINKING=true TEMP=1.0 TOP_P=0.95 PRESENCE_PENALTY=0.0 \
  bash scripts/switch.sh --force vllm/qwen38-27b-dual-max
REASONING_EFFORT=xhigh BENCHLOCAL_MODEL_TURN_TIMEOUT=900 \
bash scripts/quality-test.sh --full --enable-thinking --sampling-from-server \
  --max-tokens 4096 --thinking-max-tokens 65536 --timeout-per-case 900 > quality-test-qwen3.8-27b-xhigh.md

# medium
REASONING_EFFORT=medium BENCHLOCAL_MODEL_TURN_TIMEOUT=900 \
bash scripts/quality-test.sh --full --enable-thinking --sampling-from-server \
  --max-tokens 4096 --thinking-max-tokens 65536 --timeout-per-case 900 > quality-test-qwen3.8-27b-medium.md

# low
REASONING_EFFORT=low BENCHLOCAL_MODEL_TURN_TIMEOUT=900 \
bash scripts/quality-test.sh --full --enable-thinking --sampling-from-server \
  --max-tokens 4096 --thinking-max-tokens 65536 --timeout-per-case 900 > quality-test-qwen3.8-27b-low.md
```
## Which params are different from the announced commands in #1024 and why
[...]

PR #1350, opened 2026-09-19
bonsai-ternary: raise the reasoning budget 8192 → 32768 (derived)

noonghunna:
## What
Raises the bonsai-ternary reasoning budget **8192 → 32768**.

> **Changed from this PR's earlier revision.** It first proposed **16384**, picked as "roughly double". That turns out to be the *worst* of the candidates — see below. The number is now derived rather than chosen.

## The binding constraint is the house timeout, not a round number

A budget must be emittable inside the **900 s per-case timeout** ([#1027](https://github.com/noonghunna/club-3090/issues/1027)) at this model's measured **52.2 tok/s** decode — otherwise the wall clock kills the scenario before the budget can fire, and a guard you cannot reach is decoration.

| budget | time to emit | vs 900 s timeout |
|--:|--:|---|
| 8192 | 157 s | fits, but **below** the legitimate trace range |
| 16384 | 314 s | fits, but **inside** it |
| **32768** | **628 s** | ✅ **fits, ~30% headroom**; clears observed max by 73% |
| 65536 | 1256 s | ⛔ **exceeds it — can never fire** |

## ⚠️ Why 16384 was the worst option

The same enumeration prompt drew **18,895** and **13,013** tokens on two draws at temp 1.0. 16384 falls *between* them, so it truncates **nondeterministically** — the same prompt sometimes completing and sometimes not, which is the hardest failure to diagnose. A budget belongs clearly above the distribution or clearly below it, never inside.

## ⭐ Why not lower

| | observed |
|---|--:|
| legitimate long reasoning | 13,013 / 18,895 tok |
| runaway (user report, #1027) | ~170,000 tok |

Two orders of magnitude apart, so any budget in the ~20k–100k gap separates them: it cuts pathology while letting real work finish. **8192 sits below the legitimate range**, so it fires on ordinary enumeration rather than only on runaways — a constant quality tax to catch a rare event.

## Measured behaviour

- ⚠️⚠️ **Ships with `--reasoning-budget-message`, always.** At 8192 the truncated run returned `finish_reason=stop` and a confident **1,264-token answer** — the marker was the *only* evidence the chain had been cut. Upstream defaults that message to empty.
- The budget is **graceful**, unlike `max_tokens`: it closes the thinking block and lets the model answer from a partial chain rather than cutting the stream dead. But the model does **not** know the budget in advance, so it cannot pace itself — it concludes at exhaustion.
- ✅ Validated as the default, not as an override: booted with no `REASONING_BUDGET` env set and read the live `llama-server` argv.

## What this does NOT establish

- ⚠️ **Evidence base is two draws of one prompt.** An unbounded xhigh 8-pack at the shipped 262144 context is running; if its trace p99 exceeds 32768, this number should be revisited.
- ⚠️ 32768 does **not** guarantee no truncation at xhigh — the 170k report is on this model. What makes a non-binding guard safe is that truncation stays visible.
- The same change is applied to the MTP sibling on `feat/prism-mtp-ternary-slug`, which inherited the same too-small 8192 from #1349.


henrykrinkle01:
To clarify: the 170k thinking tokens weren't runaway. It was just a complex prompt.
In my xhigh run referenced in #1027, there were tests close to the 900s timeout (and passed), which should use most of the 65k reasoning budget at ~70 tps

PR #1356, opened 2026-09-20
docs(qwen3.8-27b): caveat multi4/multi8 — fa2-fp8kv runs outside its validated envelope (#1076)

noonghunna:
Documentation only. Records that the four `multi*` DFlash2 composes run the `fa2-fp8kv` plugin **outside its validated envelope**, and records what a reference-rig A/B already ruled out so nobody re-tests it.

Prompted by @foureight84's crash report in #1076.

## The mismatch

Three independent sources scope this plugin to **one** compose:

| source | says |
|---|---|
| plugin `README.md` | *"The compose targets **TP=2 and one sequence**."* |
| `scripts/lib/profiles/patches.yml` | `load_bearing_when: composes: [vllm/qwen38-27b-dual-ultramax]` |
| `docs/UPSTREAM.md` | *"`vllm/qwen38-27b-dual-ultramax` installs prebuilt kernels…"* |

It is installed by **seven** — `dual` ×3, `multi4` ×2, `multi8` ×2.

| topology | TP | `max_num_seqs` | in envelope? |
|---|--:|--:|---|
| `dual` | 2 | 1 | ✅ exactly as documented |
| `multi4` | 4 | 2 | ❌ |
| `multi8` | 8 | 2 | ❌ |

This shipped because the `multi*` tiers are community-validated by design — the reference rig has two cards, so nobody could boot them to notice.

## What the caveats say

Each header now carries the envelope mismatch, the open crash signature (`torch.ops.fa2_fp8kv.forward` → `Sequence bounds exceed this port's supported envelope`, fatal to `EngineCore`), and — most importantly — that **it fails silently**: the container restarts, the harness retries, and the run publishes a plausible but tainted score. Three `--full` runs read 129/128/127 that way. Points at the #1354 `RestartCount` guard as the thing that now catches it.

## What was ruled out (so it is not re-tested)

A two-arm A/B on the reference rig, same plugin artifact (`731d1942…`), same pack, same flags:

| arm | TP | `max_num_seqs` | RestartCount | HA-16/17 |
|---|--:|--:|---|---|
| A (shipped `dual`) | 2 | 1 | 0 → 0 | both completed |
| B | 2 | **2** | 0 → 0 | both completed |

- **`max_num_seqs` is NOT the trigger.** Arm B ran it at 2 on TP=2 with zero restarts.
- **TP=4 is not a geometry problem.** With `num_key_value_heads=4`, `head_dim=256`, TP=4 resolves to `(256, 1)` local KV geometry, which **is** in the plugin's tested set `(256,1)/(256,2)/(128,4)`.
- **`multi8` is further out than `multi4`** — TP=8 needs 0.5 local KV heads (4 KV heads / 8 ranks), i.e. KV replication, outside the tested geometries entirely. Its caveat says so explicitly.

TP remains the untested variable; isolating it needs a 4-card rig running TP=2, which is asked for in #1076.

## Scope

Header comments only — no compose behaviour, no registry status change, no `DEFAULTS` change. The tiers keep working exactly as before for anyone running them; they now say what they are.

Deliberately **not** gating the plugin off `multi*` yet: the cause is unconfirmed, and pulling the plugin would destroy the reporter's reproduction mid-investigation.


noonghunna:
Full suite re-run on the rebased branch against the fixed master (`ea704133`, post-#1355): **180/180, 0 failures.**

Re-ran after staging rather than only after editing — #1355 was a case where the pre-merge sweep went green because the added file was still untracked and the git-mode check had nothing to inspect.

